PR #12, opened 2023-11-13
MGMT-16108: introduce the Resource server

danielerez:
Added an initial implementation for the resource server. The server is serving the 'resourcePools' endpoint which fetches data from the search query API and converts the results into O2 objects.

In a nutshell, the mapping is done as follows:
* Node object from search API --> O2 Resource object
* Cluster object from search API --> O2 ResourcePool object

As some objects/properties don't have direct mapping, for now this PR does *not* include the following:
* Handling for ResourceType objects
* Mapping for these properties:
  * ResourcePool.location
  * ResourcePool.globalLocationID
  * ResourcePool.description
  * Resource.resourceTypeID


internal/cmd/server/start_resource_server.go
```diff
@@ -0,0 +1,326 @@
...
+	"strings"
+
+	"github.com/gorilla/mux"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift-kni/oran-o2ims/internal"
+	"github.com/openshift-kni/oran-o2ims/internal/exit"
+	"github.com/openshift-kni/oran-o2ims/internal/logging"
+	"github.com/openshift-kni/oran-o2ims/internal/service"
+	searchmodel "github.com/stolostron/search-v2-api/graph/model"
```
jhernand, line 31: Is there any chance that we can do without importing this? I ask because this means that means adding all the dependencies of that project, not just the the model. I had bad experiences with that, due to dependency conflicts. Can we maybe copy the required files?

internal/cmd/server/start_resource_server.go
```diff
@@ -0,0 +1,326 @@
...
+// GraphqlVars contains the GraphQL variables passed to the search API server
+type GraphqlVars struct {
+	Input []struct {
+		Filters []struct {
+			Property string   `json:"property"`
+			Values   []string `json:"values"`
+		} `json:"filters"`
+		Relatedkinds []string `json:"relatedKinds"`
+	} `json:"input"`
+}
```
jhernand, line 47: This is very similar to the `SearchInput` struct in the `searchmodel` package imported above. What is the reason for the additional struct?

internal/service/resource_pool_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	graphqlQuery string
+	graphqlVars  *searchmodel.SearchInput
+	resources    data.Stream
+}
+
+func NewResourcePoolFetcher(
+	logger *slog.Logger,
+	cloudID, backendURL, backendToken string,
+	graphqlQuery string, graphqlVars *searchmodel.SearchInput) ResourcePoolFetcher {
+	return &resourcePoolFetcher{
```
jhernand, line 51: Can we use the "builder" approach to create this object? Otherwise when this already long list of parameters grow it will be hard to understand.

go.mod
```diff
@@ -4,20 +4,25 @@ go 1.21
 
 require (
 	github.com/PaesslerAG/jsonpath v0.1.1
+	github.com/golang/mock v1.6.0
```
jhernand, line 7: We are already using `go.uber.org/mock` which is a maintaned fork of `github.com/golang/mock` please make sure to use it.

internal/cmd/server/start_resource_server.go
```diff
@@ -0,0 +1,326 @@
...
+		"",
+		"Token for authenticating to the backend server.",
+	)
+	return result
+}
+
+// ResourceServerCommand contains the data and logic needed to run the `start
+// resource-server` command.
+type ResourceServerCommand struct {
+}
```
jhernand, line 80: I see that you are going to use the `logger` in some methods of this type. In that case please add a `logger *slog.Logger` field here, initialize it in the `run` method and avoid passing it as parameter to methods. That is the reason for the existence of this type: to be able to do such things.

internal/service/resource_pool_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Transform Clusters to ResourcePools
+	resourcePools = data.Map(clusters, r.mapClusterItem)
+
+	return
+}
+
+func (r *resourcePoolFetcher) getSearchResults(ctx context.Context) (result []any, err error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
jhernand, line 114: Using `InsecureSkipVerify: true` is not acceptable for production environments. It is OK for now, but we need to address it. Please open a ticket explaining that we need to have command line options to configure the trusted certificate authorities.

internal/service/resource_pool_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	httpClient := &http.Client{Transport: tr}
+	client := graphql.NewClient(r.backendURL, graphql.WithHTTPClient(httpClient))
+
+	// Set GraphQL query and vars
+	request := graphql.NewRequest(r.graphqlQuery)
+	request.Var("input", *r.graphqlVars)
```
jhernand, line 121: What is the advantage of using the `graphql` package here? Can't we do this with a plain HTTP client? My concern is that it looks like this `graphql` package will store all the result set in memory before returning control. That means that we can't process them in an streaming approach. Is there any way to do that with the `graphql` package? Otherwise it negates the advantages of using streams.

internal/service/resource_pool_object_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	// Filter results by 'cluster' property
+	h.graphqlVars.Filters = append(h.graphqlVars.Filters, &searchmodel.SearchFilter{
+		Property: "cluster",
+		Values:   []*string{&id},
+	})
+
+	// Fetch resource pools
+	resourcePools, err := NewResourcePoolFetcher(
+		h.logger, h.cloudID, h.backendURL, h.backendToken, h.graphqlQuery, h.graphqlVars,
+	).FetchItems(ctx)
```
jhernand, line 196: Is the resource pool fetcher stateless? Can we create one instance and reuse it for all calls to `fetchObject`?

internal/service/resource_object_mapper.go
```diff
@@ -0,0 +1,73 @@
...
+package service
+
+import (
+	"context"
+
+	"github.com/openshift-kni/oran-o2ims/internal/data"
+)
+
+//go:generate mockgen -source=resource_object_mapper.go -package=service -destination=resource_object_mapper_mock.go
+type ResourceObjectMapper interface {
```
jhernand, line 24: There is already a `Mapper` type here:

https://github.com/openshift-kni/oran-o2ims/blob/ccd39ba2f1c2f016566be75721cc824d6d8870c5/internal/streaming/stream.go#L145-L146

It is a simple function, so I think we can use that directly instead of creating a new interface and a mock for it.

If using the generic type is unconfortable we can add an alias in the `data` package:

```go
type Mapper = stream.Mapper[data.Object, data.Object]
```

internal/service/resource_pool_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return
+	}
+	return
+}
+
+// Map Cluster to an O2 ResourcePool object.
+func (r *resourcePoolFetcher) mapClusterItem(ctx context.Context,
+	from data.Object) (to data.Object, err error) {
+	to, err = NewResourcePoolObjectMapper(r.cloudID, r.resources).MapItem(ctx, from)
```
jhernand, line 144: Can we create the resource pool mapper only once and reuse it?

internal/service/resource_pool_fetcher.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return
+	}
+	return
+}
+
+// Map a Node to an O2 Resource object.
+func (rr *resourcePoolFetcher) mapNodeItem(ctx context.Context,
+	from data.Object) (to data.Object, err error) {
+	to, err = NewResourceObjectMapper().MapItem(ctx, from)
```
jhernand, line 154: Can we create the resource object mapper only once and reuse it?

internal/service/resource_pool_object_mapper.go
```diff
@@ -0,0 +1,99 @@
...
+		// TODO: no direct mapping to a property in Cluster object
+		"description":      "",
+		"location":         "",
+		"globalLocationID": "",
+	}
+	return
+}
+
+// Converts a specified data stream of O2 Resources (Nodes) to an array.
+func (r *resourcePoolObjectMapper) getResourceArray(ctx context.Context, items data.Stream, resourcePoolID string) (result []map[string]any) {
```
jhernand, line 78: In general we should try to avoid things that return arrays, because they negate the value of using streams. But if we do then we can combine the `Collect` and `Select` methods. For example:

```go
selector := func(ctx context.Context, item data.Object) (result bool, err error) {
  result = item["resourcePoolID"] == resourcePoolID
  return
}
result = streams.Collect(streams.Select(items, selector))
```

internal/service/resource_pool_fetcher.go
danielerez: It mainly makes the code less clean without it. Can be removed.

jhernand (approved):
Looks good to me, thanks @danielerez!

PR #13, opened 2023-11-14
Automatically update and check copied code

jhernand:
We copy the GraphQL models from the `stolostron/search-api-v2` project in order to avoid additional dependencies. This patch adds `go:generate` directives to automatically update that code when `make generate` is executed.

Related: https://issues.redhat.com/browse/MGMT-16108
Related: https://github.com/stolostron/search-v2-api/blob/main/graph/schema.graphqls
Related: https://github.com/stolostron/search-v2-api/blob/main/graph/model/models_gen.go

### DIFF
--- a/internal/searchapi/generate.go
+++ b/internal/searchapi/generate.go
@@ -1,0 +1,22 @@
+/*
+Copyright (c) 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in
+compliance with the License. You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License is
+distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+implied. See the License for the specific language governing permissions and limitations under the
+License.
+*/
+
+package searchapi
+
+// The models used by the ACM search GraphQL server are part of the 'stolostron/search-v2-api'
+// project, but we don't want to import it because that would bring many other dependencies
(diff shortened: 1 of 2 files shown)


jhernand:
FYI: @danielerez

internal/searchapi/generate.go
```diff
@@ -0,0 +1,22 @@
...
+License.
+*/
+
+package searchapi
+
+// The models used by the ACM search GraphQL server are part of the 'stolostron/search-v2-api'
+// project, but we don't want to import it because that would bring many other dependencies
+// that may be problematic in the future. Instead of that we copy that code into our project.
+//
+//go:generate curl -Lso searchapi.go https://raw.githubusercontent.com/stolostron/search-v2-api/234de1a26d82b3de2af74cf21aa1226db8e5f26a/graph/model/models_gen.go
```
danielerez, line 21: Isn't it tied to a specific commit? I.e. I would expect it would fetch from latest 'master' branch? Or is there some other mechanism for fetching the latest bits?

jhernand: It is tied to a specific commit on purpose. We could get the latest from the `master` branch with this:

https://github.com/stolostron/search-v2-api/blob/main/graph/model/models_gen.go

But that would make the build non-repeatable.

jhernand: Not that the point here is not really to automatically update, but rather to ensure that we don't manually change this file by accident. If we did, then the `check-generated-code` action would detect it and fail the build. Not super important, just a tad better.

danielerez: > Not that the point here is not really to automatically update, but rather to ensure that we don't manually change this file by accident. If we did, then the `check-generated-code` action would detect it and fail the build. Not super important, just a tad better.

Cool, makes sense.

PR #14, opened 2023-11-14
Add `--api-listener-...` flags to resource server

jhernand:
This patch adds the `--api-listener-...` flags to the resource server:

```
$ ./oran-o2ims start resource-server \
--api-listener-address=localhost:8002 \
```

This is intended for use in the reverse proxy configuration started with `proxy.sh`, as it requires different ports for different services.

Note that a side effect of this is that the default port is now `8000` instead of `8080`.

Related: https://issues.redhat.com/browse/MGMT-16113

### DIFF
--- a/proxy.sh
+++ b/proxy.sh
@@ -47,6 +47,19 @@ pids="${pids} $!"
 &
 pids="${pids} $!"
 
+# Start the resource server:
+./oran-o2ims start resource-server \
+--log-file="servers.log" \
+--log-level="debug" \
+--log-field="server=resource" \
+--log-field="pid=%p" \
+--api-listener-address="127.0.0.1:8002" \
+--cloud-id="123" \
+--backend-url="${BACKEND_URL}" \
+--backend-token="${BACKEND_TOKEN}" \
+&
+pids="${pids} $!"
+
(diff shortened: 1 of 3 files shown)


jhernand:
FYI: @danielerez

proxy.sh
```diff
@@ -47,6 +47,19 @@ pids="${pids} $!"
...
 
+# Start the resource server:
+./oran-o2ims start resource-server \
+--log-file="servers.log" \
+--log-level="debug" \
+--log-field="server=resource" \
+--log-field="pid=%p" \
+--api-listener-address="127.0.0.1:8002" \
+--cloud-id="123" \
+--backend-url="${BACKEND_URL}" \
```
danielerez, line 58: Not related to this PR, just wonder if we should support multiple backend URLs?
I.e. should we open a ticket for supporting something like RESOURCE_BACKEND_URL, in case the search api is routed using a different domain? Or is it an unlikely scenario...

jhernand: Yes, you are right, we should support multiple backend URLs, one for each kind of backend. Please open that ticket.

danielerez:
lgtm

PR #21, opened 2023-11-16
Replace O2ims_ with o2ims- in URL prefix

avishayt:


jhernand (commented):
Looks good. You will also have to update https://github.com/openshift-kni/oran-o2ims/blob/main/proxy.yaml.

jhernand:
@danielerez @irinamihai please be aware of this fix, you will need to use the new URLs in your testing.

jhernand (approved):

PR #27, opened 2023-11-20
Simplify deployment manager handler

jhernand:
This patch changes the implementation of the deployment manager service so that it has one single handler for the collection and for individual items. That is convenient because the code for both things is very similar, in particular the mapping code is identical. In order to do that simplification it is necessary to ensure that the methods of the `CollectionHandler` and `ObjectHandler` have different names, as otherwise it isn't possible to implement both in the same object. Currently both are named `Get`. This patch renames that method to `List`, merges the deployment manager code into one single handler, and adds tests to verify that it works correctly.


internal/service/resource_pool_object_handler.go
```diff
@@ -160,7 +160,7 @@ func (b *ResourcePoolObjectHandlerBuilder) Build() (
 
 // Get is part of the implementation of the collection handler interface.
 func (h *ResourcePoolObjectHandler) Get(ctx context.Context,
```
danielerez, line 162: Can we remove also this file now? I.e. use a single handler in start_resource_server

jhernand: Yes, I think so. Let me take a look at that after merging this.

PR #28, opened 2023-11-21
Simplify resource pool handler

jhernand:
This patch merges the resource pool collection and object handler into a single handler that implements both interfaces.


jhernand:
@danielerez please review.

PR #29, opened 2023-11-22
Add some unit tests for the resource handler

jhernand:
This patch adds some unit tests for the resource handler.


jhernand:
@danielerez I wanted to do some changes in the resource handler. In order to do it with a bit more confidence I thought that it would be good to add some unit tests. That is what this patch does. Please take a look.

PR #31, opened 2023-11-26
MGMT-16260: added resources endpoint

danielerez:
This change includes the following:
* Removed 'resources' property from resourcePool object
* Added 'resources' adapters (collection and object):
  * /resourcePools/{resourcePoolID}/resources
  * /resourcePools/{resourcePoolID}/resources/{resourceID}
* Added 'ParentID' to object/collection handlers
  * Required for fetching 'resourcePoolID' from request
* Handled a 'Not found' object in ObjectAdapter
* Renamed 'searchapi' package to 'model'
  * Rquired for avoiding 'sed' on generate (which differs on mac)
* Added 'vendor' dir to gitignore


internal/service/resource_fetcher.go
```diff
@@ -0,0 +1,302 @@
...
+	itemsReader := bytes.NewReader(items)
+
+	// Create reader for Nodes
+	nodes, err := k8s.NewStream().
+		SetLogger(r.logger).
+		SetReader(itemsReader).
+		Build()
+	if err != nil {
+		return
+	}
```
jhernand, line 185: I think this conversion to JSON in order to use the "k8s.NewStream" isn't necessary. If you already have an slice containing the input objects then you can create a stream from it using the [data.Pour](https://github.com/openshift-kni/oran-o2ims/blob/45562d9bc2afa0ff70b6d353dd0aaa53e5f0fc8f/internal/data/data.go#L90-L91) function:

```go
resources = data.Map(data.Pour(resultArr), r.mapNodeItems)
```

However having that `resultArr` slice is an indication that we aren't taking advantage of the streaming approach. If you look at the `k8s.Stream` type you will see that what it reads the input until it finds the "items" element, and then it reads those items one by one, without storing them in memory. Would it be possible to do the same for the "searchResults" field returned by the GraphQL server? That would reduce the number of JSON conversions that we have to do, and would also save memory, as only one item would need to be in memory at a time.

If you agree with that approach we can do it in a separate pull request. I'd rather merge this as is.

jhernand (approved):
Looks good, thanks @danielerez.

internal/service/resource_fetcher.go
danielerez: sure, will do.

PR #32, opened 2023-11-28
MGMT-16107: added resourceTypes endpoints

danielerez:
* Added the following endpoints:
  * /resourceTypes
  * /resourceTypes/{resourceTypeID}
* Populated resoruceTypeID property on Resource object.
  * Note: resourceTypeID is currently generated using properties from Resource objects. We'll probably need to reconsider it later on.


jhernand (approved):
Looks good to me, thanks @danielerez.

PR #35, opened 2023-12-04
[WIP] Demo scripts

jhernand:
Some changes and scripts for a demo.


jhernand:
@irinamihai I don't intend to merge this pull request as is, but I thought that the `template.yaml` file included in this pull request can give you some ideas of what the I expect the operator to do.

I will extract some of these changes to other pull requests that I think should be eventually merged, but not probably that template.

PR #39, opened 2023-12-05
MGMT-16351: Add support for configured extensions

jhernand:
Currently the population of the _extensions_ field is completely hardcoded. This patch adds support for populating it in a configurable way. A new `--extensions` command line option is added to the deployment manager server. The value will be a jq script that calculates extension fields. For example, if the managed cluster has a `country` label the following command line option will configure the server to add a corresponding `country` extension:

```
--extensions={
  "country": .metadata.labels["country"]
}
```

For a cluster where the value of the `contry` label is `ES` that will result in the following extensions:

```json
{
  "country": "ES"
}
```

Related: https://issues.redhat.com/browse/MGMT-16351

### DIFF
--- a/internal/cmd/server/flags.go
+++ b/internal/cmd/server/flags.go
@@ -16,5 +16,6 @@ package server
 
 // Names of command line flags:
 const (
-	cloudIDFlagName = "cloud-id"
+	cloudIDFlagName    = "cloud-id"
+	extensionsFlagName = "extensions"
 )
--- a/internal/cmd/server/start_deployment_manager_server.go
+++ b/internal/cmd/server/start_deployment_manager_server.go
@@ -58,6 +58,11 @@ func DeploymentManagerServer() *cobra.Command {
 		"",
 		"Token for authenticating to the backend server.",
 	)
+	_ = flags.StringArray(
+		extensionsFlagName,
+		[]string{},
+		"Extension to add to deployment managers.",
+	)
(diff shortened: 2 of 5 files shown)


danielerez:
Shouldn't we just automatically populate the extensions field with all labels? I.e. by default, go over all labels and add to the extensions map. Or do we prefer to keep it defined by the user for each label?

jhernand:
> Shouldn't we just automatically populate the extensions field with all labels? I.e. by default, go over all labels and add to the extensions map. Or do we prefer to keep it defined by the user for each label?

I think making it explicit is better. As the way to specifiy the extensions is a jq script the user could easily add all the labels to the extensions if she wanted:

```
--extensions=.metadata.labels
```

danielerez:
> > Shouldn't we just automatically populate the extensions field with all labels? I.e. by default, go over all labels and add to the extensions map. Or do we prefer to keep it defined by the user for each label?
> 
> I think making it explicit is better. As the way to specifiy the extensions is a jq script the user could easily add all the labels to the extensions if she wanted:
> 
> ```
> --extensions=.metadata.labels
> ```

So we might want to do the same in the Resource server later on. For now we simply add all labels to extensions.

jhernand:
> So we might want to do the same in the Resource server later on. For now we simply add all labels to extensions.

Yes, open a ticket about that, please.

PR #42, opened 2023-12-15
MGMT-16112: Add OpenAPI handler

jhernand:
This patch adds a new `/o2ims-infrastructureInventory/v1/openapi` endpoint that returns the OpenAPI specification of the service.

Currently this specification only covers the metadata and the deployment manager endpoints. Other endpoints will be added later.

Related: https://issues.redhat.com/browse/MGMT-16112

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,11 @@ lint:
 	golangci-lint --version
 	golangci-lint run --verbose --print-resources-usage --modules-download-mode=vendor --timeout=5m0s
 
+.PHONY: lint-openapi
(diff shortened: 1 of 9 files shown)


danielerez:
/lgtm
Note: just wonder whether the spec can be generated from code on every change.
Maybe using [huma](https://github.com/danielgtaylor/huma): https://github.com/danielgtaylor/huma/blob/8f089757a73fa97636099306a6549745e1f1bb8f/docs/docs/features/openapi-generation.md?plain=1#L9

PR #47, opened 2024-01-15
MGMT-16359: support configured extensions in resource server

danielerez:
Similarly to the support add in deployment manager server, this PR is adding a similar functionality to the resource server. The main difference between the implementations is the parsing logic of the objects retrieved from search API.

I.e. since each 'label' property is a single string consists of key=value pairs, the lables should be parsed before evaluating the expression with jq. Hence, added a custom jq function for parsing: `parse_labels`.

For example:
```
--extensions={
  "openshift_version": .label|parse_labels|.openshiftVersion,
  "cluster_set": .label|parse_labels|.\"cluster.open-cluster-management.io/clusterset\",
}
```


internal/service/resource_handler.go
```diff
@@ -397,11 +431,28 @@ func (r *ResourceHandler) mapNodeItem(ctx context.Context,
 		return
 	}
 
+	labels, err := data.GetString(from, "label")
+	if err != nil {
+		return
+	}
+	labelsMap := data.GetLabelsMap(labels)
+
+	r.logger.Error(fmt.Sprintf("r.extensions: %v", r.extensions))
```
jhernand, line 440: You will probably want to remove this, or replace it with `r.logger.Debug(...)`.

internal/service/resource_handler_test.go
```diff
@@ -361,6 +361,70 @@ var _ = Describe("Resource handler", func() {
...
+								Values: []*string{
+									ptr.To("node"),
+								},
+							},
+						},
+					}).
+					SetExtensions(
+						`{
+							"operation_system": .label|parse_labels|.os,
+							"architecture": .label|parse_labels|.arch
```
jhernand, line 410: This is really nice @danielerez!

internal/service/resource_handler.go
danielerez: Good catch:) Removed.

PR #50, opened 2024-01-22
MGMT-16410: OpenAPI spec - added resource endpoints

danielerez:
* Added resource endpoints (resourceTypes/resourcePools/resources) to the OpenAPI spec.
* Added a makefile target for validting the spec.yaml file.

### DIFF
--- a/internal/openapi/spec.yaml
+++ b/internal/openapi/spec.yaml
@@ -33,6 +33,15 @@ tags:
 - name: metadata
   description: |
     Service metadata, including versions and O-Cloud information.
+- name: resourceTypes
+  description: |
+    Information about resource types.
+- name: resourcePools
+  description: |
+    Information about resource pools.
+- name: resources
+  description: |
+    Information about resources.
 
 paths:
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,14 +51,19 @@ lint:
 	golangci-lint --version
 	golangci-lint run --verbose --print-resources-usage --modules-download-mode=vendor --timeout=5m0s
 
+.PHONY: validate-openapi
+validate-openapi:
+	@echo "Run OpenAPI validator"
+	go run github.com/getkin/kin-openapi/cmd/validate@v0.119.0 internal/openapi/spec.yaml
```
jhernand, line 57: There is already a test that runs the `spectral` tool:

https://github.com/openshift-kni/oran-o2ims/blob/5daf9b9676c32c5248a448f5edd41c45f32735a6/internal/openapi/handler_test.go#L83-L117

If you still prefer to have this Makefile target I won't object, but it is a bit redundant. If you prefer this please remove that test.

jhernand (commented):
Looks good, but you may want to address the comment about using both "spectral" and "kin-openapi".

Makefile
danielerez: Oh missed this test, indeed looks good enough. I'll remove the target.

PR #52, opened 2024-01-26
MGMT-16144: ORAN O2IMS operator

irinamihai:
Description:
- new CRD: orano2ims (orano2imses.oran.openshift.io)
  * Spec:
    * CloudId (string)
    * MetadataServer (bool)
    * DeploymentManagerServer (bool)
    * IngressHost (string)
    * BackendURL (string)
    * BackendToken (string)
    * BackendType (string:Enum=regular-hub;global-hub)
  * Status:
    * DeploymentsStatus:
       * DeploymentServerStatus
       * MetadataServerStatus
       * Conditions

- controller for managing the orano2ims CRs and deploying all the
  needed resources for making requests to the metadata and deployment manager
  server:
  * Deployments: metadata-server, deployment-manager-server
  * Ingress: same name as the orano2ims CR
  * ServiceAccount: client, deployment-manager-server, metadata-server
  * Service: metadata-server, deployment-manager-server
  * ConfigMap: authz
 
- the recociliation time is by default 5 minutes

- reconciliation logic:
  * the controller is restricted to a few namespaces: default, oran,
    o2ims, oran-o2ims
  * if the metadataServer/deploymentManagementServer is requested by the spec,
    it creates the metadata deployment and all the needed resources
  * the controller first checks if the needed resource exists. If not,
    it created it and if it exists it updates/patches it to match the
    Spec.
  * the orano2ims.Status is updated at the end of the reconciliation
    loop. If there is an error with the update, the reconciliation is
    set to happen after 30 seconds.


bin/controller-gen
jhernand, line 1: I guess you don't want to include this binary in the pull request. Maybe add it to .gitignore?

bin/kustomize
jhernand, line 1: Add to .gitignore?

go.mod
```diff
@@ -1,72 +1,84 @@
-module github.com/openshift-kni/oran-o2ims
+module oran-o2ims
```
jhernand, line 1: Why is this needed?

controllers/orano2ims_controller.go
jhernand, line 1: Can we move this `controllers` package to `internal/controllers`? In general, if possible, everything that it isn't intended for use in other packages should go into the `internal` directory.

controllers/orano2ims_controller.go
```diff
@@ -0,0 +1,600 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// ORANO2IMSReconciler reconciles a ORANO2IMS object
+type ORANO2IMSReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
```
jhernand, line 52: I guess that these `Log` and `Scheme` fields are intended for internal use only. Can they be `logger` and `scheme` to make them private?

controllers/orano2ims_controller.go
```diff
@@ -0,0 +1,600 @@
...
+						Args: []string{
+							"start",
+							"metadata-server",
+							"--log-level=debug",
+							"--log-file=stdout",
+							"--api-listener-address=0.0.0.0:8000",
+							"--api-listener-tls-crt=/secrets/tls/tls.crt",
+							"--api-listener-tls-key=/secrets/tls/tls.key",
+							fmt.Sprintf("--cloud-id=%s", orano2ims.Spec.CloudId),
+							//fmt.Sprintf("--external-address=https://%s", orano2ims.Spec.IngressHost),
```
jhernand, line 221: This remind me that I need to add the `--external-address` option, thanks @irinamihai.

jhernand (commented):
I think we can merge this once we remove the included binaries and clarify why we need to change the `module ...` line in the `go.mod` file. Other than that it looks good, thanks @irinamihai!

bin/controller-gen
irinamihai: That's right, I will actually remove the whole bin directory as it will automatically be created.

go.mod
irinamihai: I suspect a go.mod was generated by kubebuilder when I did the initial operator scaffolding. I will change everything back to include github.com/openshift-kni.

controllers/orano2ims_controller.go
irinamihai: This is part of the kubebuilder scaffolding and they're also used in cmd/main.go for setting up the controller. I suppose we could try and change these, although all the examples I've looked at in the ACM org have these public:
- https://github.com/stolostron/multiclusterhub-operator/blob/main/controllers/multiclusterhub_controller.go#L74
- https://github.com/stolostron/backplane-operator/blob/main/controllers/backplaneconfig_controller.go#L71
- https://github.com/stolostron/multicluster-observability-operator/blob/main/operators/multiclusterobservability/controllers/multiclusterobservability/multiclusterobservability_controller.go#L81

I propose to leave it as is and update it if it comes out as a concern later on.

controllers/orano2ims_controller.go
irinamihai: My initial thought was to separate the image code from the operator, but what you're saying also makes sense. I will make the change. Thank you for the suggestion.

danielerez:
/approve
Looks good to me.

PR #54, opened 2024-01-31
graphql: handle unknown property in MapFilter

danielerez:
Handling unknown GraphQL filter property in MapFilter func.
It is required to fallback into selector filtering when the specified property is not available in the search response.
E.g. ?filter=(eq,location,US) --> should filter using the selector.

### DIFF
--- a/internal/graphql/graphql.go
+++ b/internal/graphql/graphql.go
@@ -15,6 +15,7 @@ License.
 package graphql
 
 import (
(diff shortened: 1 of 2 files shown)


danielerez:
rebased

danielerez:
> New changes are detected. LGTM label has been removed.

handled rebase conficts.

PR #59, opened 2024-02-15
MGMT-16856: docs - extract acm config section

danielerez:
Extract 'Operators configuration' section from both env docs to a separate file.


docs/dev/env_operators.md
```diff
@@ -0,0 +1,199 @@
+# Operators configuration
```
irinamihai, line 1: Maybe instead of operators we should say **ACM**? I think it would be more clear. The file should be renamed as well. Keeping the **operators** name could cause confusion that this is for the actual O-RAN O2IMS operator.

danielerez: Yeah, good point. Will update.

PR #61, opened 2024-02-15
MGMT-16144: Controller unit tests

irinamihai:


internal/controllers/orano2ims_controller_test.go
```diff
@@ -0,0 +1,348 @@
...
+	testscheme.AddKnownTypes(oranv1alpha1.GroupVersion, &oranv1alpha1.ORANO2IMSList{})
+	testscheme.AddKnownTypes(networkingv1.SchemeGroupVersion, &networkingv1.Ingress{})
+	testscheme.AddKnownTypes(networkingv1.SchemeGroupVersion, &networkingv1.IngressList{})
+	testscheme.AddKnownTypes(corev1.SchemeGroupVersion, &corev1.ServiceAccount{})
+	testscheme.AddKnownTypes(corev1.SchemeGroupVersion, &corev1.ServiceAccountList{})
+	testscheme.AddKnownTypes(corev1.SchemeGroupVersion, &corev1.Service{})
+	testscheme.AddKnownTypes(corev1.SchemeGroupVersion, &corev1.ServiceList{})
+	testscheme.AddKnownTypes(appsv1.SchemeGroupVersion, &appsv1.Deployment{})
+	testscheme.AddKnownTypes(appsv1.SchemeGroupVersion, &appsv1.DeploymentList{})
+}
```
jhernand, line 56: Is there any chance that we can use the Ginkgo framework for this tests like we do for the rest of the tests of the project? For example, for instead of this `init` we could do something like this:

```go
var _ = BeforeAll(func () {
        testscheme.AddKnownTypes(...)
})

File: internal/controllers/orano2ims_controller_test.go [hunk 2]
```diff
@@ -0,0 +1,348 @@
...
+
+func getFakeClientFromObjects(objs ...client.Object) (client.WithWatch, error) {
+	return fake.NewClientBuilder().WithScheme(testscheme).WithObjects(objs...).WithStatusSubresource(&oranv1alpha1.ORANO2IMS{}).Build(), nil
+}
+
+func TestORANO2IMSReconciler(t *testing.T) {
```
jhernand, line 62: And instead of this long test with the custom test cases we could use the `DescribeTable`, for example:

```go
var _ = DescribeTable(
	"Reconciler",
	func(objs []client.Object, request reconcile.Request, validate func(result ctrl.Result, reconciler ORANO2IMSReconciler)) {
		// Declare the Namespace for tha O-RAN O2IMS resource.
		ns := &corev1.Namespace{
			ObjectMeta: metav1.ObjectMeta{
				Name: "oran-o2ims",
			},
		}

		// Update the testcase objects to include the Namespace.
		objs = append(objs, ns)

		// Get the fake client.
		fakeClient, err := getFakeClientFromObjects(objs...)
		Expect(err).ToNot(HaveOccurred())

		// Initialize the O-RAN O2IMS reconciler.
		r := &ORANO2IMSReconciler{
			Client: fakeClient,
			Log:    logr.Discard(),
			Scheme: fakeClient.Scheme(),
		}

		// Reconcile.
		result, err := r.Reconcile(context.TODO(), request)
		Expect(err).ToNot(HaveOccurred())

		validate(result, *r)
	},
	Entry(
		"Metadata server deployment is updated after edit",
		reconcile.Request{
			NamespacedName: types.NamespacedName{
				Namespace: utils.ORANO2IMSNamespace,
				Name:      "oran-o2ims-sample1",
			},
		},
		[]client.Object{ ... },
		func (result ctrl.Result, reconciler ORANO2IMSReconciler) { ... },
	),
	Entry(
		"Only the metadata server is required",
		...
	),
	...
)
```

Makefile
```diff
@@ -307,6 +307,8 @@ go-generate:
 test tests:
 	@echo "Run ginkgo"
 	ginkgo run -r $(ginkgo_flags)
+	@echo "Run controller unit test"
+	go test -v ./internal/controllers/...
```
jhernand, line 311: As I explain in other comments I'd suggest to use the Ginkgo framework for these tests, that way we don't have to different ways to run tests and display the results, and these two lines wouldn't be needed.

jhernand (commented):
I think it would be good to consider using the Ginkgo framework for these tests, to keep consistency across all the tests of the project.

jhernand:
Looks great, thanks @irinamihai!

/lgtm

PR #65, opened 2024-02-20
MGMT-16144: Operator - Add extension support

irinamihai:
Description:
- Add a new array type Spec field called Extensions
- As per O-RAN.WG6.O2IMS-INTERFACE-R003-v04.00, this new field is of key:value format and it has been implemented as a structure:
```
type Extension struct {
	Key   string `json:"key,omitempty"`
	Value string `json:"value,omitempty"`
}
```
- In the ORANO2IMS CR definition, the Extensions can be declared as  follows:
```
  spec:
    extensions:
    - key: country
      value: ".metadata.labels[\"country\"]"
    - key: my
      value: "{memory: .status.capacity.memory, k8s_version: .status.version.kubernetes}"
```


internal/controllers/orano2ims_controller.go
```diff
@@ -328,13 +330,7 @@ func (r *ORANO2IMSReconciler) deployManagerServer(ctx context.Context, orano2ims
...
 							fmt.Sprintf("--backend-token=%s", orano2ims.Spec.BackendToken),
 							fmt.Sprintf("--backend-type=%s", orano2ims.Spec.BackendType),
-							fmt.Sprintln(
-								// TODO: properly hold the extensions instead of hardcoding them.
-								"--extensions={\n",
-								"\"country\": .metadata.labels[\"country\"],\n",
-								"\"version\": .metadata.labels[\"openshiftVersion\"],\n",
-								"\"hub\": .metadata.annotations[\"global-hub.open-cluster-management.io/managed-by\"],",
-								"}"),
+							fmt.Sprintf("--extensions={%s}", extensions),
```
jhernand, line 333: If we do this we are forcing the user to put each extension within a dictionary. That limits somehow the flexibility, as there are some jq constructs that don't start with a curly bracket. For example, you can set variables and use them later:

```jq
.metadata.labels["name"] as $name |
{
  name: $name,
  alias: $name
}
```

If we put that between curly brackets it will result in a syntax error.

I think it is better to make the `extensions` field an array of strings, and translate it into multiple `--extensions` flag. So the CR could look like this:

```yaml
extensions:
- |
    .metadata.labels["name"] as $name |
    {
      name: $name,
      alias: $name
    }
- |
    { foo: "bar" }
- ...
```

That way we can tell users that the syntax of the value of the `extensions` field is just that of jq, without additional restrictions.

internal/controllers/utils/utils_test.go
```diff
@@ -0,0 +1,219 @@
...
+	It("Returns empty string when there are no extensions", func() {
+		extensions := []oranv1alpha1.Extension{}
+
+		extString := ExtensionToString(extensions)
+		Expect(extString).To(Equal(""))
+	})
+})
+
+var _ = Describe("DoesK8SResourceExist", func() {
+
```
jhernand, line 71: Do these tests belong here? The aren't related to extensions. May this be a copy & paste error?

irinamihai: I included some extra testcases in this PR. I can add them separately in another PR.

jhernand: That is fine, keep them.

internal/controllers/orano2ims_controller.go
irinamihai: Ok, that makes sense. I interpreted the O-RAN standard that is describing the extensions as key:value pair.
Thank you for the clarification, @jhernand.
Kindly take another look when you get the time and please let me know if this approach better suits the extension scenarios.

jhernand: Your interpretation of the specification is correct, but this is about how we are implementing the addition of configurable extensions with jq. It looks good to me now. Thanks @irinamihai!

PR #67, opened 2024-02-21
Add DELETE support

jhernand:
Add DELETE support

This patch adds support for the `DELETE` HTTP method. The adapter will now accept the `DELETE` method and call the `Delete` method defined in the `DeleteHandler` interface. For example, a simple implementation that stores the object in a map in memory could look like this:

```go
func (h *MyHandler) Delete(ctx context.Context, request *AddRequest) (respone *AddResponse, err error) {
        h.storeLock.Lock()
        defer h.storeLock.Unlock()
        delete(h.storeMap, request.Variables[0])
        return
}
```

### DIFF
--- a/internal/service/adapter.go
+++ b/internal/service/adapter.go
@@ -51,6 +51,7 @@ type Adapter struct {
 	listHandler        ListHandler
 	getHandler         GetHandler
 	addHandler         AddHandler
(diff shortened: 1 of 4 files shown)


jhernand:
Note that this shares a commit with #66. That should be merged before this, and then this should be rebased.

PR #73, opened 2024-03-13
MGMT-16977: Mapping Alerts to AlarmEventRecords

danielerez:
Added mapping logic to the AlarmFetcher:
* For each Alert retrieved from the AlertManager, map to an AlarmEventRecord object.
* Use the resource server url for fetching info about resources and pools (needed for resourceID/resourceTypeID properties).

Also added an API for fetching alarm probable causes:
* This API is not defined by O2ims spec, but we can use it for exposing a custom list of probable causes.
* This list is available in data folder (data/alarms/probable_causes.json), and can be customized and maintained by server admins as required.

Note: this PR does not include AlarmDictionary support (will be handled separately).

### DIFF
--- a/internal/service/alarm_fetcher.go
+++ b/internal/service/alarm_fetcher.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	neturl "net/url"
 
+	"github.com/imdario/mergo"
 	"github.com/openshift-kni/oran-o2ims/internal/data"
 	"github.com/openshift-kni/oran-o2ims/internal/jq"
 	"github.com/openshift-kni/oran-o2ims/internal/k8s"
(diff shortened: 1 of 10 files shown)


internal/service/alarm_fetcher.go
```diff
@@ -232,8 +255,157 @@ func (h *AlarmFetcher) doGet(ctx context.Context, path string,
...
+		"alarmDefinitionID":  alertName,
+		"probableCauseID":    alertName,
+		"perceivedSeverity":  AlarmSeverity(severity).MapProperty(),
+		"extensions":         extensionsMap,
+	}
+
+	return
+}
+
+func (r *AlarmFetcher) FetchResourcePool(ctx context.Context, clusterId string) (resourcePool data.Object, err error) {
```
jhernand, line 348: Do these functions need to be public?

jhernand:
I only have a minor question about why a function is public. Other than that it looks good. Feel free to merge.

internal/service/alarm_fetcher.go
danielerez: Not really, can be private, fixed. Good catch:)

danielerez:
> I only have a minor question about why a function is public. Other than that it looks good. Feel free to merge.

Can indeed be private, fixed. Thanks:)

PR #76, opened 2024-03-26
Clarify the need to clone assited repo on spoke deployment step.

tshefi:
Replace hardcoded "git" user on dev-script's path with a variable $USER.
Typo swapped a comma with period.

### DIFF
--- a/docs/dev/env_devscripts.md
+++ b/docs/dev/env_devscripts.md
@@ -32,7 +32,7 @@ Go to https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/, clic
 right, copy the login command, extract the token from the command and use it to set `CI_TOKEN` in `config_$USER.sh`.
 
 #### Get pull secret
-Convert the secret obtained from [cloud.openshift.com](https://cloud.redhat.com/openshift/install/pull-secret), originally in text format, into JSON format and save it as `pull_secret.json`. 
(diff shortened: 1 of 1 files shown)


docs/dev/env_devscripts.md
```diff
@@ -32,7 +32,7 @@ Go to https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/, clic
 right, copy the login command, extract the token from the command and use it to set `CI_TOKEN` in `config_$USER.sh`.
 
 #### Get pull secret
-Convert the secret obtained from [cloud.openshift.com](https://cloud.redhat.com/openshift/install/pull-secret), originally in text format, into JSON format and save it as `pull_secret,json`. 
+Convert the secret obtained from [cloud.openshift.com](https://cloud.redhat.com/openshift/install/pull-secret), originally in text format, into JSON format and save it as `pull_secret.json`.
```
danielerez, line 35: Not related to this change, but I just wonder what did you mean by 'conversion to JSON'? I.e. which operation exactly did you have to make? I've just tried to use the 'Copy pull secret' button, and the result seems the same as the downloaded file:/

tshefi: Correct, but I originally clicked on "Download pull secret", resulting in txt format, thus had to convert. 
Maybe I should tweak the sentence to reflect the fact that we should either use "copy pull secret"(json already) or if using the "Download pull secret"(txt) option you'd to convert to json.

PR #79, opened 2024-04-02
Add defaults for deployment manager backend URL and token

jhernand:
Currently the backend URL and token don't have defaults. That means that the user has to explicitly set them in the custom resource that represents the O2 IMS deployment. That is reasonable for the URL, but not for the token, because it is the token of a service account that is created by the operator. That means that the user would have to create the service account in advance, wait for the token to be created and then read it and update the custom resource. That isn't practical. To avoid that this patch changes the operator so that the token is the default URL of the Kubernetes API server and the token is taken from the `/run/secrets/kubernetes.io/serviceaccount/token` file.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -9,5 +9,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/imihai/oran-o2ims-operator
-  newTag: 4.16.0
+  newName: quay.io/jhernand/o2ims-operator
+  newTag: "2"
--- a/internal/controllers/utils/utils.go
+++ b/internal/controllers/utils/utils.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"slices"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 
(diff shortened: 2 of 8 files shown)


internal/controllers/utils/utils.go
```diff
@@ -150,40 +151,78 @@ func GetDeploymentVolumeMounts(serverName string) []corev1.VolumeMount {
...
+		if backendURL == "" {
+			backendURL = "https://kubernetes.default.svc"
+		}
+		result = append(
+			result,
+			fmt.Sprintf("--backend-url=%s", backendURL),
+		)
+
+		// If no backend token has been provided then use the token of the service account
+		// that is will eventually execute the server. Note that the file may not exist,
```
irinamihai, line 206:
```suggestion
		// that will eventually execute the server. Note that the file may not exist,
```

internal/controllers/utils/utils.go
```diff
@@ -150,40 +151,78 @@ func GetDeploymentVolumeMounts(serverName string) []corev1.VolumeMount {
...
+				result,
+				fmt.Sprintf("--backend-type=%s", orano2ims.Spec.BackendType),
+			)
+		}
+
+		// If no backend URL has been provided then use the default URL of the Kubernetes
+		// API server of the cluster:
+		backendURL := orano2ims.Spec.BackendURL
+		if backendURL == "" {
+			backendURL = "https://kubernetes.default.svc"
```
irinamihai, line 198: We could consider moving this to utils/constants.go.

internal/controllers/utils/utils.go
```diff
@@ -150,40 +151,78 @@ func GetDeploymentVolumeMounts(serverName string) []corev1.VolumeMount {
...
+		// but we can't check it here as that will be a different pod.
+		if orano2ims.Spec.BackendToken != "" {
+			result = append(
+				result,
+				fmt.Sprintf("--backend-token=%s", orano2ims.Spec.BackendToken),
+			)
+		} else {
+			result = append(
+				result,
+				"--backend-token-file=/run/secrets/kubernetes.io/serviceaccount/token",
```
irinamihai, line 216: Same, comment about moving this to utils/constants.go

internal/controllers/utils/utils.go
jhernand: Yes, I will do that.

internal/controllers/utils/utils.go
jhernand: Yes, will do.

config/manager/kustomization.yaml
```diff
@@ -9,5 +9,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/imihai/oran-o2ims-operator
-  newTag: 4.16.0
+  newName: quay.io/jhernand/o2ims-operator
```
irinamihai, line 12: Just a note that we will have to update these to the o2ims quay repo.

PR #81, opened 2024-04-09
Use assisted installer admin kubeconfig if exists

jhernand:
Currently the deployment manager service uses the admin kubeconfig that was used to register the cluster with ACM. But this will not exist if the cluster was created with assisted installer. To avoid that problem this patch changes the service so that it first check if the secret used by the assisted installer exists. If it does exist then it will use it.

### DIFF
--- a/internal/service/deployment_manager_handler.go
+++ b/internal/service/deployment_manager_handler.go
@@ -579,8 +579,24 @@ func (h *DeploymentManagerHandler) fetchProfile(ctx context.Context,
 		return
 	}
 
-	// Fetch the kubeconfig that was used to register the hub, and then use it to fetch the
-	// admin kubeconfig of the hub:
+	// First we will try to get the admin Kubeconfig from the namespace of the cluster. This
+	// will work if the cluster was provisioned using ACM and assisted installer.
+	assistedInstallerAdminKubeconfig, err := h.fetchAssistedInstallerAdminKubeconfig(
+		ctx,
+		h.hubClient,
+		cluster,
+	)
+	if err != nil {
+		return
+	}
+	if assistedInstallerAdminKubeconfig != nil {
+		h.logger.Info(
+			"Using assisted installer admin kubeconfig",
+			slog.String("cluster", cluster),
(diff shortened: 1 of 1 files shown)


internal/service/deployment_manager_handler.go
```diff
@@ -579,8 +579,20 @@ func (h *DeploymentManagerHandler) fetchProfile(ctx context.Context,
...
+	assistedInstallerAdminKubeconfig, err := h.fetchAssistedInstallerAdminKubeconfig(
+		ctx,
+		h.hubClient,
+		cluster,
+	)
+	if err != nil {
+		return
+	}
+	if assistedInstallerAdminKubeconfig != nil {
+		result, err = h.makeProfile(assistedInstallerAdminKubeconfig)
```
danielerez, line 593: Maybe worth having some log here?
I.e. a debug/info log that indicates which kubeconfig is used.

danielerez (commented):
Looks good, just a minor comment.

internal/service/deployment_manager_handler.go
jhernand: Makes sense, will do it.

internal/service/deployment_manager_handler.go
```diff
@@ -638,12 +654,58 @@ func (h *DeploymentManagerHandler) fetchProfile(ctx context.Context,
...
+	)
 
 	// Make the profile data from the cluster admin kubeconfig:
 	result, err = h.makeProfile(clusterAdminKubeconfig)
 	return
 }
 
+// fetchAssistedInstallerAdminKubeconfig uses the given Kubernetes API client to fetch the admin
+// kubeconfig that is created by the assisted installer when cluster installation finishes. It
+// returns the serialized kubeconfig, or nil if there it doesn't exist, for example if the cluster
```
irinamihai, line 669: Typo: extra "there".

jhernand: Thanks Irina :-) , fixing it ...

PR #82, opened 2024-04-09
Grant deployment manager server account permissions to read managed clusters and secrets

jhernand:
The service account that is used to run the deployment manager server needs permissions to read `ManagedCluster` objects and assisted installer admin kubeconfig secrets. This patch changes the operator so that it creates the corresponding cluster roles and cluster role bindings. In order to do so the operator needs to also have those permissions, as Kubernetes forbids granting more permissions than the service account granting them already has. So this patch also adds those permissions to the service account that is used to run the operator.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -9,5 +9,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/jhernand/o2ims-operator
-  newTag: "2"
+  newName: quay.io/imihai/oran-o2ims-operator
+  newTag: 4.16.0
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -9,5 +9,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/imihai/oran-o2ims-operator
+  newName: quay.io/openshift-kni/oran-o2ims-operator
   newTag: 4.16.0
--- a/internal/controllers/orano2ims_controller.go
+++ b/internal/controllers/orano2ims_controller.go
@@ -36,6 +36,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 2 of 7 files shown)


config/manager/kustomization.yaml
```diff
@@ -9,5 +9,5 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/jhernand/o2ims-operator
-  newTag: "2"
+  newName: quay.io/imihai/oran-o2ims-operator
```
irinamihai, line 12: Could we change this to the final name which would be `quay.io/openshift-kni/oran-o2ims-operator`?

internal/controllers/orano2ims_controller.go
```diff
@@ -224,6 +222,108 @@ func (r *ORANO2IMSReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		return err
+	}
+
+	// The role and binding are cluster level objects, so to make them unique we use the
+	// namespace as a prefix:
+	roleName := fmt.Sprintf("%s-%s", orano2ims.Namespace, utils.ORANO2IMSAlarmSubscriptionServerName)
+	bindingName := fmt.Sprintf("%s-%s", orano2ims.Namespace, utils.ORANO2IMSAlarmSubscriptionServerName)
+
+	// Grant the service account permissions to read the objects it needs, for example managed
```
irinamihai, line 244: We could consider moving the creation of the ClusterRole & ClusterRoleBinding to their own functions, like for the other resources create by the operator.

internal/controllers/orano2ims_controller.go
```diff
@@ -224,6 +222,108 @@ func (r *ORANO2IMSReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	// clusters:
+	role := &rbacv1.ClusterRole{
```
irinamihai, line 246: I think it would be a good idea to use the `utils.CreateK8sCR` function for creating these new resources for 2 reasons:
1. It sets the owner resource, so when the `orano2ims` CR is deleted, these resources will also be deleted.
2. If anyone is to update these resources, the operator will PATCH/UPDATE them to the expected values so that everything works as expected.

Please ignore these if they are not relevant to what you had in mind.

config/manager/kustomization.yaml
jhernand: Yes, will do.

internal/controllers/orano2ims_controller.go
jhernand: Yes, makes sense, will do it.

internal/controllers/orano2ims_controller.go
jhernand: I agree, will do it.

jhernand:
@irinamihai I think I addressed all your suggestions, please take another look.

PR #85, opened 2024-04-11
Reduce numer of parameters of `CreateK8sCR`

jhernand:
Currently this function has nine parameters. Two of them are the name and namespace of the object, but the object itself is also passed. Assuming that the `Name` and `Namespace` fields of the object will be populated it is not necessary to have additional parameters for that. This patch removes those parameters.


jhernand:
@irinamihai while addressing your suggestion to use `CreateK8sCR` I noticed these apparently redundant parameters. It is OK to remove them?

PR #86, opened 2024-04-11
Support creation of non namespaced objects in createk8scr

jhernand:
Kubernetes forbids the creation of object in one namespace with owners in another namespace, or the creation of non-namespaced objects with namespaced owners. But the `CreateK8sCR` function doesn't support. We will need that in order to create cluster roles and cluster role bindings. This patch changes the function so that it will not try to set an owner if the namespace of the object and the namespace of the owner are different.


jhernand:
Note that this pull request also includes the code in #85. That should be merged before this, and then this should be rebased.

PR #87, opened 2024-04-11
Remove `oldObject` parameter from `CreateK8sCR`

jhernand:
Currently the `CreateK8sCR` function receives an `oldObject` parameter that is needed because the function doesn't know how to create an empty object of the same type of the object that is going to be created. But that can be solved using reflection, something like this:

```go
// Create an empty object of the same type of the new object. We will use it to
// fetch the current state.
objectType := reflect.TypeOf(newObject).Elem()
oldObject := reflect.New(objectType).Interface().(client.Object)
```

This patch does that and removes the `oldObject` parameter, which simplifies the code that calls the `CreateK8sCR` function.

### DIFF
--- a/internal/controllers/utils/utils.go
+++ b/internal/controllers/utils/utils.go
@@ -20,7 +20,7 @@ import (
 var oranUtilsLog = ctrl.Log.WithName("oranUtilsLog")
 
 func CreateK8sCR(ctx context.Context, c client.Client,
-	newObject client.Object, ownerObject client.Object, oldObject client.Object,
+	newObject client.Object, ownerObject client.Object,
 	runtimeScheme *runtime.Scheme, operation string) (err error) {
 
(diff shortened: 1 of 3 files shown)


internal/controllers/utils/utils.go
```diff
@@ -31,7 +31,11 @@ func CreateK8sCR(ctx context.Context, c client.Client,
 		return err
 	}
 
-	// Check if the CR already exists.
+	// Create an empty object of the same type of the new object. We will use it to store the
+	// fetch the current state.
```
irinamihai, line 35:
```suggestion
	// fetch of the current state.
```

jhernand: Thanks @irinamihai, I will fix this.

jhernand: Done.

PR #88, opened 2024-04-11
Remove scheme parameter of `CreateK8sCR` and `ORANO2IMSReconciler`

jhernand:
It isn't necessary to explicitly pass the scheme paratemer to the `CreateK8sCR` function because it can obtain it calling the `Scheme()` function of the client.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -105,7 +105,6 @@ func main() {
 	if err = (&controllers.ORANO2IMSReconciler{
 		Client: mgr.GetClient(),
 		Log:    ctrl.Log.WithName("controller").WithName("ORAN-O2IMS"),
-		Scheme: mgr.GetScheme(),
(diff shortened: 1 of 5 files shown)


irinamihai:
Hmm, I do wonder why [kubebuilder](https://book.kubebuilder.io/cronjob-tutorial/controller-overview.html) adds `*runtime.Scheme` by default to the Reconciler structure if it's not actually needed :thinking:
The controller-runtime example does not use it, so I think we should be ok.

Kindly add more info on this (as a comment here or readme) if you have it, it would be greatly appreciated.

jhernand:
> Kindly add more info on this (as a comment here or readme) if you have it, it would be greatly appreciated.

I don't really have more information, just wanted to reduce the number of parameters of the `CreateK8sCR` function and noticed that the scheme can be obtained from the client. I will take a look at what kubebuilder does and update if I find some better explanation.

PR #90, opened 2024-04-14
Mention that namespaces should be created first

tshefi:
Else yaml apply fails with an excepted: namespaces "mgmt-spoke1" not found.


danielerez:
/ok-to-test
/lgtm
/approve

Thanks Tzach!

PR #96, opened 2024-04-26
Subscription server use ConfigMap as persistent storage as well as IPC

Jennifer-chen-rh:
This is WIP version for alarm subscription server using configmap for now.

### DIFF
--- a/internal/cmd/server/start_alarm_subscription_server.go
+++ b/internal/cmd/server/start_alarm_subscription_server.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift-kni/oran-o2ims/internal/authentication"
 	"github.com/openshift-kni/oran-o2ims/internal/authorization"
 	"github.com/openshift-kni/oran-o2ims/internal/exit"
+	"github.com/openshift-kni/oran-o2ims/internal/k8s"
 	"github.com/openshift-kni/oran-o2ims/internal/logging"
 	"github.com/openshift-kni/oran-o2ims/internal/metrics"
 	"github.com/openshift-kni/oran-o2ims/internal/network"
@@ -205,13 +206,28 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 	})
 	router.Use(metricsWrapper, authenticationWrapper, authorizationWrapper)
 
+	// create k8s client with kube(from env first)
+	//var config *rest.Config
+	kubeClient, err := k8s.NewClient().SetLogger(logger).SetLoggingWrapper(loggingWrapper).Build()
+
+	if err != nil {
+		logger.ErrorContext(
+			ctx,
+			"Failed to create kubeClient",
+			"error", err,
+		)
+		return exit.Error(1)
+	}
+
 	// Create the handler:
 	handler, err := service.NewAlarmSubscriptionHandler().
 		SetLogger(logger).
 		SetLoggingWrapper(loggingWrapper).
 		SetCloudID(cloudID).
 		SetExtensions(extensions...).
-		Build()
+		SetKubeClient(kubeClient).
+		Build(ctx)
+
 	if err != nil {
 		logger.ErrorContext(
 			ctx,
--- a/internal/k8s/client.go
+++ b/internal/k8s/client.go
@@ -29,9 +29,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clnt "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 // Client builder contains the data and logic needed to create a Kubernetes API client that
@@ -51,6 +53,13 @@ type ClientBuilder struct {
 type Client struct {
 	logger   *slog.Logger
 	delegate clnt.WithWatch
+}
+
+// set fake client to be used for test.go
+func NewFakeClient() *Client {
+	client := Client{}
+	client.delegate = fake.NewFakeClient()
+	return &client
 }
 
 // NewClient creates a builder that can then be used to configure and create a Kubernetes API client
@@ -283,6 +292,11 @@ func (c *Client) GroupVersionKindFor(obj runtime.Object) (schema.GroupVersionKin
 
 func (c *Client) IsObjectNamespaced(obj runtime.Object) (bool, error) {
 	return c.delegate.IsObjectNamespaced(obj)
+}
+
+// watch
+func (c *Client) Watch(ctx context.Context, obj clnt.ObjectList, opts ...clnt.ListOption) (watch.Interface, error) {
+	return c.delegate.Watch(ctx, obj, opts...)
 }
 
 // Close closes the client and releases all the resources it is using. It is specially important to
--- a/internal/service/alarm_subscription_handler.go
+++ b/internal/service/alarm_subscription_handler.go
@@ -24,20 +24,30 @@ import (
 
 	"github.com/google/uuid"
 	jsoniter "github.com/json-iterator/go"
+
 	"github.com/openshift-kni/oran-o2ims/internal/data"
 	"github.com/openshift-kni/oran-o2ims/internal/jq"
+	"github.com/openshift-kni/oran-o2ims/internal/k8s"
 	"github.com/openshift-kni/oran-o2ims/internal/search"
-	"golang.org/x/exp/maps"
+
+	"github.com/openshift-kni/oran-o2ims/internal/persiststorage"
 )
 
-// DeploymentManagerHandlerBuilder contains the data and logic needed to create a new deployment
+const (
+	TestNamespace     = "orantest"
+	TestConfigmapName = "orantestconfigmapalarmsub"
+	FieldOwner        = "oran-o2ims"
+)
+
+// alarmSubscriptionHandlerBuilder contains the data and logic needed to create a new deployment
 // manager collection handler. Don't create instances of this type directly, use the
-// NewDeploymentManagerHandler function instead.
+// NewAlarmSubscriptionHandler function instead.
 type alarmSubscriptionHandlerBuilder struct {
 	logger         *slog.Logger
 	loggingWrapper func(http.RoundTripper) http.RoundTripper
 	cloudID        string
 	extensions     []string
+	kubeClient     *k8s.Client
 }
 
 // alarmSubscriptionHander knows how to respond to requests to list deployment managers.
@@ -48,11 +58,13 @@ type alarmSubscriptionHandler struct {
 	loggingWrapper           func(http.RoundTripper) http.RoundTripper
 	cloudID                  string
 	extensions               []string
+	kubeClient               *k8s.Client
 	jsonAPI                  jsoniter.API
 	selectorEvaluator        *search.SelectorEvaluator
 	jqTool                   *jq.Tool
 	subscritionMapMemoryLock *sync.Mutex
-	subscriptionMap          map[string]data.Object
+	subscriptionMap          *map[string]data.Object
+	persistStore             *persiststorage.KubeConfigMapStore
 }
 
 // NewAlarmSubscriptionHandler creates a builder that can then be used to configure and create a
@@ -90,8 +102,15 @@ func (b *alarmSubscriptionHandlerBuilder) SetExtensions(
 	return b
 }
 
+// SetExtensions sets the fields that will be added to the extensions.
+func (b *alarmSubscriptionHandlerBuilder) SetKubeClient(
+	kubeClient *k8s.Client) *alarmSubscriptionHandlerBuilder {
+	b.kubeClient = kubeClient
+	return b
+}
+
 // Build uses the data stored in the builder to create anad configure a new handler.
-func (b *alarmSubscriptionHandlerBuilder) Build() (
+func (b *alarmSubscriptionHandlerBuilder) Build(ctx context.Context) (
 	result *alarmSubscriptionHandler, err error) {
 	// Check parameters:
 	if b.logger == nil {
@@ -100,6 +119,11 @@ func (b *alarmSubscriptionHandlerBuilder) Build() (
 	}
 	if b.cloudID == "" {
 		err = errors.New("cloud identifier is mandatory")
+		return
+	}
+
+	if b.kubeClient == nil {
+		err = errors.New("kubeClient is mandatory")
 		return
 	}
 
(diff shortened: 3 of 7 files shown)


Jennifer-chen-rh:
@jhernand This is WIP version. I will work/double check the corresponding go-test codes. Also need the multiple configmap instances to make it scale better. 

For the go route loop, I currently use sleeping. do we have other approach not using sleeping? Also when we support multiple configmap instances, what's general practice here:  use single thread to watch all, or each thread watch single instance of configmap?

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+		//there is error and err is not notfound error
+		//panic("unexpected error")
+		return
+	}
+
+	saved_data := configmap.Data
+
+	//configmap does not exist
+	if saved_data == nil {
+		saved_data = make(map[string]string)
```
jhernand, line 428: Consider using `savedData = map[string]string{}`, seems to be the preferred idiom in Go.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: TESTNAMESPACE,
+			Name:      TESTCONFIGMAPNAME,
+		},
+		Data: make(map[string]string),
+	}
+
+	configmap.Data = saved_data
```
jhernand, line 444: Instead of that you can put `Data: savedData` above.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+
+	//apply new data
+
+	//u := &unstructured.Unstructured{}
+	//u.Object, _ = runtime.DefaultUnstructuredConverter.ToUnstructured(configmap)
+	err = h.hubClient.Patch(ctx, configmap, clnt.Apply, clnt.ForceOwnership, clnt.FieldOwner("field-owner"))
```
jhernand, line 450: If we are going to use field ownership then the name should probably be something like `oran-o2ims` isntead of `field-owner`.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+	_, ok := configmap.Data[entry_key]
+
+	//entry not found
+	if !ok {
+		return
+	}
+	delete(configmap.Data, entry_key)
+
+	// Note: there is a case configmap is empty shall we remove the configmap
+	err = h.hubClient.Update(ctx, configmap, &clnt.UpdateOptions{})
```
jhernand, line 516: I think we should use here the same `Patch` method that you used in `ConfigMapAddEntry`, otherwise the `Update` method may generate conflicts and we would need to detect and handle them.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+	}
+
+	for mapkey, value := range configmap.Data {
+		var object data.Object
+		err = jsoniter.Unmarshal([]byte(value), &object)
+		if err != nil {
+			continue
+		}
+
+		h.subscriptionMap[mapkey] = object
```
jhernand, line 549: Remember that we may have multiple goroutines running this code simultaneously, and maps are not safe in that situation. Should we use locking here?

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+		return
+	}
+
+	if configmap.Data == nil {
+		return
+	}
+
+	for mapkey, value := range configmap.Data {
+		var object data.Object
+		err = jsoniter.Unmarshal([]byte(value), &object)
```
jhernand, line 544: Consider using `h.jsonAPI.Unmarshal`, it shares configuration and buffers with other parts of the handler.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+	}
+
+	go func() {
+		for {
+			event, open := <-watcher.ResultChan()
+			if open {
+				switch event.Type {
+				case watch.Added:
+					fallthrough
+				case watch.Modified:
```
jhernand, line 575: Using `fallthrough` is OK, but I'd rather just put both options in the same case branch:

```go
case watch.Added, watch.Modified:
```

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+					h.subscritionMapMemoryLock.Lock()
```
jhernand, line 576: When you acquire a lock it is good practice to ensure that it will be released even if there are errors. Here, for example, you have a `continue` statement that will end the iteration without releasing the lock. One possible solution for that is to create an anonymous function and use the `defer` mechanism:

```go
func () {
    h.subscriptionMapMemoryLock.Lock()
    defer h.subscriptionMemoryLock.Unlock()
    ...
}()
```

You can also use a method instead of an anonymous function, if there are not many local variables to pass.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+							err = jsoniter.Unmarshal([]byte(value), &object)
+							if err != nil {
+								continue
+							}
+							h.subscriptionMap[k] = object
+						}
+					}
+					h.subscritionMapMemoryLock.Unlock()
+				case watch.Deleted:
+					h.subscritionMapMemoryLock.Lock()
```
jhernand, line 591: Same here, use the `defer` mechanism even if it is not strictly needed. In this case this code can't possible fail (unless if memory is exhausted) but it may be changed in the future and cause a locking bug.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+					h.subscriptionMap = make(map[string]data.Object)
+					h.subscritionMapMemoryLock.Unlock()
+
+				default:
+
+				}
+
+			}
+			//sleep
+			time.Sleep(10 * time.Millisecond)
```
jhernand, line 601: I think we don't need to check if the channel delivered a result or was closed: iterating it would be enough, and then there is no need for the sleep.

Anyhow as I commented somewhere else this is not the preferred way to watch for Kubernetes events: consider using a controller and a reconciler.

internal/service/handlers.go
```diff
@@ -155,11 +155,52 @@ type DeleteHandler interface {
...
+
+type RecoveryFromConfigMapHandler interface {
+	//The handler read to DB and build/recovery datastructures in memory
+	RecoveryFromConfigMap(ctx context.Context) (err error)
+}
+
+type WatchConfigMapHandler interface {
+	//The handler read to DB and build/recovery datastructures in memory
+	WatchConfigMap(ctx context.Context) (err error)
+}
```
jhernand, line 190: In general it is good practice in Go to make interfaces small, but I'd say that this is taking that too far. These methods are closely related so I think they belong in a single interface. Also the names of the methods are two configmap/kubernetes specific. What if we want to make an implementation that uses a relational database? Can you try to come up with something more generic?

internal/cmd/server/start_alarm_subscription_server.go
```diff
@@ -187,6 +187,11 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string)
 		return exit.Error(1)
 	}
 
+	//read from configmap
+	handler.RecoveryFromConfigMap(ctx)
+	//watch configmap changes
+	handler.WatchConfigMap(ctx)
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Addressed.

internal/service/alarm_subscription_handler.go
```diff
@@ -21,13 +21,25 @@ import (
...
 	"github.com/openshift-kni/oran-o2ims/internal/data"
 	"github.com/openshift-kni/oran-o2ims/internal/jq"
+	"github.com/openshift-kni/oran-o2ims/internal/k8s"
 	"github.com/openshift-kni/oran-o2ims/internal/search"
-	"golang.org/x/exp/maps"
+)
+
+const (
+	TESTNAMESPACE     string = "orantest"
+	TESTCONFIGMAPNAME string = "orantestconfigmap"
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Addressed

internal/service/alarm_subscription_handler.go
```diff
@@ -140,6 +154,14 @@ func (b *alarmSubscriptionHandlerBuilder) Build() (
 		}
 	}
 
+	// create k8s client with kube(from env first)
+	//var config *rest.Config
+	hubClient, err := k8s.NewClient().SetLogger(b.logger).SetLoggingWrapper(b.loggingWrapper).Build()
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Addressed

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: Changed.

internal/service/handlers.go
Jennifer-chen-rh: Addressed

internal/k8s/client.go
```diff
@@ -285,6 +286,11 @@ func (c *Client) IsObjectNamespaced(obj runtime.Object) (bool, error) {
 	return c.delegate.IsObjectNamespaced(obj)
 }
 
+// watch
+func (c *Client) Watch(ctx context.Context, obj clnt.ObjectList, opts ...clnt.ListOption) (watch.Interface, error) {
+	return c.delegate.Watch(ctx, obj, opts...)
```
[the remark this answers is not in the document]
Jennifer-chen-rh: As we discussed, will pick up retryWatch

internal/service/alarm_subscription_handler.go
```diff
@@ -53,6 +65,8 @@ type alarmSubscriptionHandler struct {
 	jqTool                   *jq.Tool
 	subscritionMapMemoryLock *sync.Mutex
 	subscriptionMap          map[string]data.Object
+	//place for configmap handle (kube client)
+	hubClient *k8s.Client
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Addressed

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: As this is infinite loop, handle the error path differently.

internal/service/handlers.go
```diff
@@ -155,11 +155,52 @@ type DeleteHandler interface {
...
 	GetHandler
 	AddHandler
 	DeleteHandler
+	//configmap interfaces
+	ConfigMapChangeNotifyHandler
+	ConfigMapReadEntryHandler
+	ConfigMapAddEntryHandler
+	ConfigMapDeleteEntryHandler
+	RecoveryFromConfigMapHandler
+	WatchConfigMapHandler
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Addressed

internal/service/handlers.go
Jennifer-chen-rh: Made the changes.

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: As we discussed.

internal/service/alarm_subscription_handler.go
```diff
@@ -271,7 +294,13 @@ func (h *alarmSubscriptionHandler) fetchItem(ctx context.Context,
 func (h *alarmSubscriptionHandler) fetchItems(ctx context.Context) (result data.Stream, err error) {
 	h.subscritionMapMemoryLock.Lock()
 	defer h.subscritionMapMemoryLock.Unlock()
-	ar := maps.Values(h.subscriptionMap)
+
+	ar := make([]data.Object, 0, len(h.subscriptionMap))
```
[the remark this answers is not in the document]
Jennifer-chen-rh: To avoid duplicated subscription ID value in the configmap, I removed the subscription ID in data.object. This is the place we loop the datamap and encode the subscript ID back. later "stream" is used to build the http packet. Please let me know.

internal/service/alarm_subscription_handler.go
```diff
@@ -320,6 +349,19 @@ func (h *alarmSubscriptionHandler) addItem(
...
 
 	subId = h.getSubcriptionId()
+
+	//save the subscription in configuration map
+	value, err := jsoniter.MarshalIndent(&input_data.Object, "", " ")
+	//value, err := jsoniter.Marshal(input_data.Object)
+	if err != nil {
+		return
+	}
+	h.ConfigMapAddEntry(ctx, subId, string(value))
```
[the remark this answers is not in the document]
Jennifer-chen-rh: changed.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
 	//TBD only save related attributes in the future
 	return input, nil
 }
+
+func (h *alarmSubscriptionHandler) ConfigMapAddEntry(
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Changed.

internal/service/alarm_subscription_handler.go
```diff
@@ -358,3 +402,206 @@ func (h *alarmSubscriptionHandler) mapItem(ctx context.Context,
...
+	if saved_data == nil {
+		saved_data = make(map[string]string)
+	}
+	saved_data[entry_key] = value
+
+	configmap = &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
+		},
```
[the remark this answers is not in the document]
Jennifer-chen-rh: When we use "patch", it seems needed

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: changed.

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: changed.

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: Changed.

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: This is called  during daemon initialization before any event could cause datamap update.

internal/service/alarm_subscription_handler.go
Jennifer-chen-rh: Changed.

internal/service/alarm_subscription_handler.go
```diff
@@ -336,6 +378,8 @@ func (h *alarmSubscriptionHandler) addItem(
 func (h *alarmSubscriptionHandler) deleteItem(
 	ctx context.Context, delete_req DeleteRequest) (err error) {
 
+	h.ConfigMapDeleteEntry(ctx, delete_req.Variables[0])
```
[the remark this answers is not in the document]
Jennifer-chen-rh: addressed.

PR #99, opened 2024-05-27
WIP: Add plugin API

jhernand:
This patch adds the definition of the Kubernetes custom resource definitions that will be used by the O2 IMS implementation to interact with the hardware manager.


api/v1alpha1/node_order_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// values of the condition type are `Fulfilled` and `Failed`.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// BMC contains the details of the baseboard management controller of the node.
+	BMC BMC `json:"bmc"`
+}
+
+// BMC contains the details needed to connect to the baseboard management controller of a node.
+type BMC struct {
+	URL string `json:"url"`
```
danielerez, line 48: Just a reminder, we probably want a reference to the creds secret here.
E.g.
`
CredentialsSecretName string
`

api/v1alpha1/deployment_manager_template_types.go
```diff
@@ -0,0 +1,108 @@
...
+type DeploymentManagerTemplateNodeProfile struct {
+	// Name is the name of the profile.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// FirmwareSettings contains the firmware settings for the nodes that use this profile.
+	//
+	// +kubebuilder:validation:Optional
+	FirmwareSettings map[string]string `json:"firmwareSettings,omitempty"`
```
danielerez, line 80: IIRC, there was a requirement to support separate extensions maps for FirmwareSettings and FirmwareVersions.
I.e. we can consider having these as structs instead of maps.

api/v1alpha1/condition_types.go
```diff
@@ -0,0 +1,25 @@
...
+implied. See the License for the specific language governing permissions and limitations under the
+License.
+*/
+
+package v1alpha1
+
+const (
+	// OrderFulfilledCondition is the condition that indicates that an order has been
+	// completely and successfully completed.
+	OrderFulfilledCondition = "Fullfilled"
```
irinamihai, line 20: Typo: 'Fulfilled'

api/v1alpha1/deployment_manager_order_types.go
```diff
@@ -0,0 +1,79 @@
...
+	// Template is the name of the deployment manager template that will be used to create the
+	// deployment manager.
+	//
+	// +kubebuilder:validation:Required
+	Template string `json:"template,omitempty"`
+
+	// Location is the geographical location of the requested deployment manager.
+	//
+	// +kubebuilder:validation:Required
+	Location string `json:"profile"`
```
irinamihai, line 37: For consistency, shouldn't the golang variable and the json key have the same name, `location` / `profile` in this case? Or is the difference here intended?

api/v1alpha1/deployment_manager_order_types.go
```diff
@@ -0,0 +1,79 @@
...
+	// Name is the name of the requested deployment manager.
+	//
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Template is the name of the deployment manager template that will be used to create the
+	// deployment manager.
+	//
+	// +kubebuilder:validation:Required
+	Template string `json:"template,omitempty"`
```
irinamihai, line 32: I don't think we need `omitempty` if the field is Required.

api/v1alpha1/deployment_manager_template_types.go
```diff
@@ -0,0 +1,108 @@
...
+// manager.
+//
+// +kubebuilder:resource:shortName=dmtemplate
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+type DeploymentManagerTemplate struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// NodeProfiles is a collection of named node profiles that will be ued in the template.
```
irinamihai, line 31: Typo: `used`

api/v1alpha1/deployment_manager_template_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Profile string `json:"profile"`
+
+	// Extensions contains additional information associated with the node set that can't
+	// be expressed in the other fields.
+	//
+	// +kubebuilder:validation:Optional
+	Extensions map[string]string `json:"extensions,omitempty"`
+}
+
+// DeploymentManagerTemplateNodeProfile associates a name with a set of node node settings.
```
irinamihai, line 70: Extra `node`

api/v1alpha1/node_order_types.go
jhernand: Yes, I added that in the latest version of the patch.

api/v1alpha1/deployment_manager_template_types.go
jhernand: Actually I think I will remove these fields in favor of the extensions. Vendors can there add whatever they need. For example:

```yaml
extensions:
- oran.acme.com/firmwareSettings: |
    {
       "CStates": "C6",
       "VirtualizationEnabled": "true"
    }
- oran.acme.com/firmwareVersions: |
    {
      "UEFI": "1.3",
      "NIC": "2.7"
    }
```

In the future if we see that different vendors use the same data structures we can re-add these fields.

api/v1alpha1/condition_types.go
jhernand: Thanks Irinal, will do.

api/v1alpha1/deployment_manager_template_types.go
danielerez: Nice, sounds like a good idea. Just probably worth to document that example for the keys then.

api/v1alpha1/deployment_manager_order_types.go
jhernand: Yes, this is a plain mistake, I will fix the JSON key.

api/v1alpha1/deployment_manager_order_types.go
jhernand: OK, will remove it.

api/v1alpha1/node_order_types.go
```diff
@@ -0,0 +1,96 @@
...
+	BMC corev1.LocalObjectReference `json:"bmc,omitempty"`
+
+	// Conditions represents the observations of the current state of the template. Possible
+	// values of the condition type are `Fulfilled` and `Failed`.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// BMCDetails contains the details needed to connect to the baseboard management controller of
+// a node.
+type BMCDetails struct {
```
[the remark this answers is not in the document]
jhernand: It is used in the next version of the patch.

api/v1alpha1/deployment_manager_types.go
```diff
@@ -0,0 +1,49 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DeploymentManager is the schema a deployment manager.
+//
+// +kubebuilder:resource:shortName=dm
```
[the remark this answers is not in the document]
jhernand: It is more descriptive, but it would make other short names longer and harder to type, for example `dmgrtemplate` instead of `dmtemplate`. I will keep this for now, unless you have a strong objection.

PR #102, opened 2024-06-10
Supports for alarm notification server:

Jennifer-chen-rh:
Create the alarm notification server.
   Manage in memory alarm subscriptions base on persist storage (populated by alarm subscription server).
   Prebuilt subscription filter terms + limited optimization
   Accept the alarms, match the filters and build and send out notification packet based on matched subscription.

This PR supports basic alert packet flow. In future the filter match will have more optimization and the alert to notification the attributes may need fine tuned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,8 +185,12 @@ $ ./oran-o2ims start infrastructure-inventory-subscription-server \
 --log-field="server=resource-subscriptions" \
 --api-listener-address=localhost:8004 \
 --metrics-listener-address=localhost:8008 \
+--namespace="test" \
+--configmap-name="testInventorySubscriptionConfigmap" \
 --cloud-id=123
 ```
+
+Note: By default, the namespace of "orantest" and "oran-infra-inventory-sub" are used currently.
 
 For more information about other command line flags use the `--help` command:
 
@@ -332,6 +336,8 @@ $./oran-o2ims start alarm-subscription-server \
 --log-field="pid=%p" \
 --api-listener-address="127.0.0.1:8006" \
 --metrics-listener-address="127.0.0.1:8008" \
+--namespace="test" \
+--configmap-name="testConfigmap" \
 --cloud-id="123" 
 ```
 
@@ -340,6 +346,8 @@ be conflicts if you try to run multiple servers in the same machine. The
 `--api-listener-address` and `--metrics-listener-address` options are used to select a port number that isn't in use.
 
 The `cloud-id` is any string that you want to use as identifier of the O-Cloud instance.
+
+By default, the namespace of "orantest" and configmap-name of "oran-o2ims-alarm-subscriptions" are used.
 
 For more information about other command line flags use the `--help` command:
 
@@ -354,5 +362,57 @@ $ curl -s http://localhost:8001/o2ims-infrastructureMonitoring/v1/alarmSubscript
 ```
 Above example will get a list of existing alarm subscriptions
 
+```
+$ curl -s -X POST --header "Content-Type: application/json" -d @subscription.json http://localhost:8000/o2ims-infrastructureMonitoring/v1/alarmSubscriptions
+```
+Above example will post an alarm subscription defined in subscription.json file 
+
 Inside _VS Code_ use the _Run and Debug_ option with the `start
 alarm-subscription-server` [configuration](.vscode/launch.json).
+
+#### Alarm Notification server
+
+The alarm-notification-server should use together with alarm subscription server. The alarm subscription sever accept and manages the alarm subscriptions. The alarm notificaton servers synch the alarm subscriptions via perisist storage. To use the configmap to persist the subscriptions, the namespace "orantest" should be created at hub cluster for now (will use official oranims namespace in future). Alarm subscripton server and corresonding alarm notification server should have same namespace and configmap-name. The alarm notification server accept the alerts, match the subscription filter, build and send out the alarm notification based on url in the subscription.
+
+The required Resource server URL and token can be obtained as follows:
+
+```
+$ export RESOURCE_SERVER_URL=http://localhost:8002/o2ims-infrastructureInventory/v1/
+$ export RESOURCE_SERVER_TOKEN=$(
+  oc whoami --show-token
+)
+```
+
+Start the alarm notification server with a command like this:
+
+```
+$./oran-o2ims start alarm-notification-server \
+--log-file="servers.log" \
+--log-level="debug" \
+--log-field="server=alarm-notification" \
+--log-field="pid=%p" \
+--api-listener-address="127.0.0.1:8010" \
+--metrics-listener-address="127.0.0.1:8011" \
+--cloud-id="123" \
+--namespace="test" \
+--configmap-name="testConfigmap" \
+--resource-server-url="${RESOURCE_SERVER_URL}"
+--resource-server-token="${RESOURCE_SERVER_TOKEN}" \
+```
+
+Note that by default all the servers listen on `localhost:8000`, so there will
+be conflicts if you try to run multiple servers in the same machine. The
+`--api-listener-address` and `--metrics-listener-address` options are used to select a port number that isn't in use.
+
+The `cloud-id` is any string that you want to use as identifier of the O-Cloud instance.
+
+By default, the namespace of "orantest" and configmap-name of "oran-o2ims-alarm-subscriptions" are used.
+
+For more information about other command line flags use the `--help` command:
+
+```
+$ ./oran-o2ims start alarm-notification-server --help
+```
+
+Inside _VS Code_ use the _Run and Debug_ option with the `start
+alarm-notification-server` [configuration](.vscode/launch.json).
--- a/internal/cmd/server/flags.go
+++ b/internal/cmd/server/flags.go
@@ -133,11 +133,15 @@ func GetTokenFlag(
 
 // Names of command line flags:
 const (
-	backendTypeFlagName      = "backend-type"
-	backendTokenFlagName     = "backend-token"
-	backendTokenFileFlagName = "backend-token-file"
-	backendURLFlagName       = "backend-url"
-	cloudIDFlagName          = "cloud-id"
-	extensionsFlagName       = "extensions"
-	externalAddressFlagName  = "external-address"
+	backendTypeFlagName               = "backend-type"
+	backendTokenFlagName              = "backend-token"
+	backendTokenFileFlagName          = "backend-token-file"
+	backendURLFlagName                = "backend-url"
+	cloudIDFlagName                   = "cloud-id"
+	extensionsFlagName                = "extensions"
+	externalAddressFlagName           = "external-address"
+	namespaceFlagName                 = "namespace"
+	subscriptionConfigmapNameFlagName = "configmap-name"
+	resourceServerURLFlagName         = "resource-server-url"
+	resourceServerTokenFlagName       = "resource-server-token"
 )
--- a/internal/cmd/server/start_alarm_notification_server.go
+++ b/internal/cmd/server/start_alarm_notification_server.go
@@ -34,12 +34,12 @@ import (
 	"github.com/openshift-kni/oran-o2ims/internal/service"
 )
 
-// Server creates and returns the `start alarm-subscription-server` command.
-func AlarmSubscriptionServer() *cobra.Command {
-	c := NewAlarmSubscriptionServer()
+// Server creates and returns the `start alarm-notification-server` command.
+func AlarmNotificationServer() *cobra.Command {
+	c := NewAlarmNotificationServer()
 	result := &cobra.Command{
-		Use:   "alarm-subscription-server",
-		Short: "Starts the alarm Subscription Server",
+		Use:   "alarm-notification-server",
+		Short: "Starts the alarm Notification Server",
 		Args:  cobra.NoArgs,
 		RunE:  c.run,
 	}
@@ -55,11 +55,6 @@ func AlarmSubscriptionServer() *cobra.Command {
 		"",
 		"O-Cloud identifier.",
 	)
-	_ = flags.StringArray(
-		extensionsFlagName,
-		[]string{},
-		"Extension to add to alarm subscriptions.",
-	)
 	_ = flags.String(
 		namespaceFlagName,
 		"",
@@ -70,22 +65,32 @@ func AlarmSubscriptionServer() *cobra.Command {
 		"",
 		"The configmap name used by alarm subscriptions ",
 	)
+	_ = flags.String(
+		resourceServerURLFlagName,
+		"",
+		"The resource server URL used by alarm subscriptions ",
+	)
+	_ = flags.String(
+		resourceServerTokenFlagName,
+		"",
+		"The resource server token used by alarm subscriptions ",
+	)
 	return result
 }
 
-// alarmSubscriptionServerCommand contains the data and logic needed to run the `start
-// alarm-subscription-server` command.
-type AlarmSubscriptionServerCommand struct {
+// alarmNotificationServerCommand contains the data and logic needed to run the `start
+// alarm-notification-server` command.
+type AlarmNotificationServerCommand struct {
 }
 
-// NewAlarmSubscriptionServer creates a new runner that knows how to execute the `start
-// alarm-subscription-server` command.
-func NewAlarmSubscriptionServer() *AlarmSubscriptionServerCommand {
-	return &AlarmSubscriptionServerCommand{}
+// NewAlarmNotificationServer creates a new runner that knows how to execute the `start
+// alarm-notification-server` command.
+func NewAlarmNotificationServer() *AlarmNotificationServerCommand {
+	return &AlarmNotificationServerCommand{}
 }
 
-// run executes the `start alarm-subscription-server` command.
-func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) error {
+// run executes the `start alarm-notification-server` command.
+func (c *AlarmNotificationServerCommand) run(cmd *cobra.Command, argv []string) error {
 	// Get the context:
 	ctx := cmd.Context()
 
@@ -107,12 +112,10 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		)
 		return exit.Error(1)
 	}
-
 	// Get the cloud identifier:
 	cloudID, err := flags.GetString(cloudIDFlagName)
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Failed to get cloud identifier flag",
 			"flag", cloudIDFlagName,
 			"error", err.Error(),
@@ -120,34 +123,15 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		return exit.Error(1)
 	}
 	if cloudID == "" {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Cloud identifier is empty",
 			"flag", cloudIDFlagName,
 		)
 		return exit.Error(1)
 	}
-	logger.InfoContext(
-		ctx,
+	logger.Info(
 		"Cloud identifier",
 		"value", cloudID,
-	)
-
-	// Get the extensions details:
-	extensions, err := flags.GetStringArray(extensionsFlagName)
-	if err != nil {
-		logger.ErrorContext(
-			ctx,
-			"Failed to extension flag",
-			"flag", extensionsFlagName,
-			"error", err.Error(),
-		)
-		return exit.Error(1)
-	}
-	logger.InfoContext(
-		ctx,
-		"alarm subscription extensions details",
-		slog.Any("extensions", extensions),
 	)
 
 	// Create the logging wrapper:
@@ -156,8 +140,7 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		SetFlags(flags).
 		Build()
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Failed to create transport wrapper",
 			"error", err.Error(),
 		)
@@ -170,8 +153,7 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		SetFlags(flags).
 		Build()
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Failed to create authentication wrapper",
 			slog.String("error", err.Error()),
 		)
@@ -182,8 +164,7 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		SetFlags(flags).
 		Build()
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Failed to create authorization wrapper",
 			slog.String("error", err.Error()),
 		)
@@ -193,7 +174,7 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 	// Create the metrics wrapper:
 	metricsWrapper, err := metrics.NewHandlerWrapper().
 		AddPaths(
-			"/o2ims-infrastructureMonitoring/-/alarmSubscriptions/-",
+			"/o2ims-infrastructureMonitoring/-/alarmNotifications/-",
 		).
 		SetSubsystem("inbound").
 		Build()
@@ -229,7 +210,35 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		return exit.Error(1)
 	}
 
-	// Get the namespace:
+	// Get the resource server details:
+	resourceServerURL, err := flags.GetString(resourceServerURLFlagName)
+	if err != nil {
+		logger.ErrorContext(
+			ctx,
+			"Failed to get resource server URL flag",
+			"flag", resourceServerURLFlagName,
+			"error", err.Error(),
+		)
+		return exit.Error(1)
+	}
+	if resourceServerURL == "" {
+		logger.ErrorContext(
+			ctx,
+			"Resource server URL is empty",
+			slog.String("flag", resourceServerURLFlagName),
+		)
+		return exit.Error(1)
+	}
+	resourceServerToken, err := flags.GetString(resourceServerTokenFlagName)
+	if err != nil || resourceServerToken == "" {
+		logger.ErrorContext(
+			ctx,
+			"Resource server URL is empty",
+			slog.String("flag", resourceServerTokenFlagName),
+		)
+		return exit.Error(1)
+	}
+
 	namespace, err := flags.GetString(namespaceFlagName)
 	if err != nil {
 		logger.ErrorContext(
@@ -243,7 +252,6 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 	if namespace == "" {
 		namespace = service.DefaultNamespace
 	}
-
 	// Get the configmapName:
 	subscriptionsConfigmapName, err := flags.GetString(subscriptionConfigmapNameFlagName)
 	if err != nil {
@@ -259,49 +267,30 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		subscriptionsConfigmapName = service.DefaultAlarmConfigmapName
 	}
 
-	handler, err := service.NewSubscriptionHandler().
+	//create handler
+	handler, err := service.NewAlarmNotificationHandler().
 		SetLogger(logger).
 		SetLoggingWrapper(loggingWrapper).
 		SetCloudID(cloudID).
-		SetExtensions(extensions...).
 		SetKubeClient(kubeClient).
-		SetSubscriptionIdString(service.SubscriptionIdAlarm).
 		SetNamespace(namespace).
 		SetConfigmapName(subscriptionsConfigmapName).
+		SetResourceServerURL(resourceServerURL).
+		SetResourceServerToken(resourceServerToken).
 		Build(ctx)
 
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
+		logger.Error(
 			"Failed to create handler",
 			"error", err,
 		)
 		return exit.Error(1)
 	}
 
-	// Create the routes:
-	adapter, err := service.NewAdapter().
-		SetLogger(logger).
-		SetPathVariables("alarmSubscriptionID").
-		SetHandler(handler).
-		Build()
-	if err != nil {
-		logger.ErrorContext(
-			ctx,
-			"Failed to create adapter",
-			"error", err,
-		)
-		return exit.Error(1)
-	}
 	router.Handle(
-		"/o2ims-infrastructureMonitoring/{version}/alarmSubscriptions",
-		adapter,
-	).Methods(http.MethodGet, http.MethodPost)
-
-	router.Handle(
-		"/o2ims-infrastructureMonitoring/{version}/alarmSubscriptions/{alarmSubscriptionID}",
-		adapter,
-	).Methods(http.MethodGet, http.MethodDelete)
+		"/",
+		handler,
+	).Methods(http.MethodPost)
 
 	// Start the API server:
 	apiListener, err := network.NewListener().
@@ -309,9 +298,8 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		SetFlags(flags, network.APIListener).
 		Build()
 	if err != nil {
-		logger.ErrorContext(
-			ctx,
-			"Failed to create API listener",
+		logger.Error(
+			"Failed to to create API listener",
 			slog.String("error", err.Error()),
 		)
 		return exit.Error(1)
@@ -325,6 +313,7 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string) 
 		Addr:    apiListener.Addr().String(),
 		Handler: router,
 	}
+
 	exitHandler.AddServer(apiServer)
 	go func() {
 		err = apiServer.Serve(apiListener)
--- a/internal/persiststorage/persiststorage_configmap.go
+++ b/internal/persiststorage/persiststorage_configmap.go
@@ -2,6 +2,7 @@ package persiststorage
 
 import (
 	"context"
+	"errors"
 	"sync"
 
 	jsoniter "github.com/json-iterator/go"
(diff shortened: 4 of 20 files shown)


README.md
```diff
@@ -282,4 +282,37 @@ $ curl -s http://localhost:8001/o2ims-infrastructureMonitoring/v1/alarmSubscript
...
 
 Inside _VS Code_ use the _Run and Debug_ option with the `start
-alarm-subscription-server` [configuration](.vscode/launch.json).
\ No newline at end of file
+alarm-subscription-server` [configuration](.vscode/launch.json).
+
+
+#### Alarm Notification server
+
+The alarm-notification-server should use together with alarm subscription server. The alarm subscription severs accept and manage the alarm subscriptions. The alarm notificaton servers synch the alarm subscriptions via perisist storage. To use the configmap to persist the subscriptions, the namespace "orantest" should be created at hub cluster for now. The alarm notification server accept the alerts, match the subscription filter, build and send out the alarm notification based on url in the subscription.
```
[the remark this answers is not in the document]
Jennifer-chen-rh: thanks for the comments. Changed. Please see updated version.

internal/service/alarm_notification_dispatcher.go
```diff
@@ -0,0 +1,97 @@
...
+		if !ok {
+			h.logger.Debug(
+				"alarmNotificationHandler failed to get subinfo key",
+				slog.String(": ", key),
+			)
+
+			continue
+		}
+
+		/*
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Cleaned up.

internal/service/alarm_notification_dispatcher.go
```diff
@@ -0,0 +1,97 @@
...
+				"alarmEventRecord": $alarmEvent
+			}`,
+			request.Object, &obj,
+			jq.String("$alarmConsumerSubId", subInfo.consumerSubscriptionId),
+			jq.String("$objRef", eventRecordId),
+			jq.String("$subId", key),
+			jq.Any("$alarmEvent", request.Object),
+		)
+
+		//stream := data.Pour(obj)
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Cleaned up

internal/service/alarm_notification_dispatcher.go
```diff
@@ -0,0 +1,97 @@
...
+
+		go func(pkt data.Object) {
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Added the comment before the function. Keep like this so can easily to view the context of what passed in the routine.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,306 @@
...
+func (h *alarmNotificationHandler) watchPersistStore(ctx context.Context) (err error) {
+	err = persiststorage.ProcessChangesWithFunction(h.persistStore, ctx, ProcessStorageChanges)
+
+	if err != nil {
+		panic("failed to launch watcher")
+	}
+	return
+}
+
+/*
```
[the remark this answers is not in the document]
Jennifer-chen-rh: cleaned up.

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,233 @@
...
+		b.logger.Error(
+			"failed to create filter expression parser: ",
+			slog.String("error: ", err.Error()),
+		)
+		return
+	}
+	b.selectorParser = selectorParser
+}
+
+/*
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Cleaned up.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,306 @@
...
+		SetName(TestConfigmapName).
+		SetFieldOwnder(FieldOwner).
+		SetJsonAPI(&jsonAPI).
+		SetClient(b.kubeClient)
+
+	// http client to send out notification
+	// use 2 sec first
+	httpClient := http.Client{Timeout: 2 * time.Second}
+
+	if singleAlarmNotificationHandle != nil {
```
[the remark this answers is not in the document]
Jennifer-chen-rh: This is good question. I added 'single' here to label the singleton of the alarmNotificationHandler object. Let's wait for 3rd people to break the even. I do not mind to change it if more people think 'alarmNotificationHandler' is cleaner.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}
+
+	alarmSubscriptionSearcher := newAlarmSubscriptionSearcher()
+	alarmSubscriptionSearcher.SetLogger(b.logger).SetJqTool(jqTool).build()
+
+	// create persist storeage option
+	persistStore := persiststorage.NewKubeConfigMapStore().
+		SetNameSpace(TestNamespace).
+		SetName(TestConfigmapName).
```
danielerez, line 173: TestConfigmapName is a constant just for current testings, right? It probably should be customizable later on.

internal/persiststorage/persiststorage_configmap.go
```diff
@@ -249,3 +249,45 @@ func (s *KubeConfigMapStore) ProcessChanges(ctx context.Context, dataMap **map[s
...
+		Namespace: s.nameSpace,
+		Name:      s.name,
+	})
+	opt := clnt.ListOptions{}
+	opt.Raw = &raw_opt
+
+	watcher, err := s.client.Watch(ctx, &corev1.ConfigMapList{}, &opt)
+
+	go func() {
+		for {
```
danielerez, line 264: Do we need some tests for this logic?

internal/service/alarm_notification_dispatcher.go
```diff
@@ -0,0 +1,89 @@
...
+	err = h.jqTool.Evaluate(`.alarmEventRecordId`, request.Object, &eventRecordId)
+	if err != nil {
+		h.logger.Debug(
+			"alarm does not have alarmEventRecordId included ",
+		)
+	}
+
+	subIdSet := h.getSubscriptionIdsFromAlarm(ctx, request.Object)
+
+	//now look id_set and send http packets to URIs
```
danielerez, line 32: typo? e.g. "loop over IDs"

internal/service/alarm_notification_dispatcher.go
```diff
@@ -0,0 +1,89 @@
...
+			h.logger.Debug(
+				"alarmNotificationHandler failed to get subinfo key",
+				slog.String(": ", key),
+			)
+
+			continue
+		}
+
+		var obj data.Object
+		// alarmNotificationType needs to be added
```
danielerez, line 46: worth adding 'TODO: '

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return
+	}
+
+	subInfo := (*b.subscriptionInfoMap)[subId]
+	subInfo.filters = *result
+	(*b.subscriptionInfoMap)[subId] = subInfo
+
+	//for now use path 0 only
+	//to be fixed with full path for quicker search
```
danielerez, line 108: consider adding 'TODO: '

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	for key, value := range *subscriptionMap {
+
+		(*b.subscriptionInfoMap)[key] = subscriptionInfo{}
+
+		subInfo := (*b.subscriptionInfoMap)[key]
+		//get uris
+		var uris string
+		err = jqTool.Evaluate(`.callback`, value, &uris)
+		if err != nil {
+			panic("the callback is mandary for subscription")
```
danielerez, line 134: s/mandary/mandatory

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+				" error: ", err.Error(),
+			)
+		}
+	}
+	return
+}
+
+func (h *alarmNotificationHandler) getSubscriptionIdsFromAlarm(ctx context.Context, alarm data.Object) (result alarmSubIdSet) {
+
+	h.subscriptionMapMemoryLock.RLock()
```
danielerez, line 170: Why do we need the lock here actually? Maybe add some comment for explanation.

danielerez (commented):
Looks good. Mostly minor comments.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/openshift-kni/oran-o2ims/internal/persiststorage"
+	"github.com/openshift-kni/oran-o2ims/internal/search"
+)
+
+// singlton alarm notification handler
+var singleAlarmNotificationHandle *alarmNotificationHandler = nil
+
+// AlarmNotificationManagerHandlerBuilder contains the data and logic needed to create a new alarm notification
+// collection handler. Don't create instances of this type directly, use the
+// NewAlarmNotificationHandler function instead.
```
jhernand, line 39: This comment isn't accurate, probably the result of copy & paste?

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+type alarmNotificationHandlerBuilder struct {
+	logger         *slog.Logger
+	loggingWrapper func(http.RoundTripper) http.RoundTripper
+	cloudID        string
```
jhernand, line 43: Is the `cloudID` actually needed by this handler? If not then we should remove it.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	extensions     []string
+	kubeClient     *k8s.Client
+}
+
+// key string is uuid
+type alarmSubIdSet map[string]struct{}
+
+// alarmNotificationHander knows how to respond to requests to list deployment managers.
+// Don't create instances of this type directly, use the NewAlarmNotificationHandler function
+// instead.
```
jhernand, line 53: This comment isn't accurate.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+var singleAlarmNotificationHandle *alarmNotificationHandler = nil
+
+// AlarmNotificationManagerHandlerBuilder contains the data and logic needed to create a new alarm notification
+// collection handler. Don't create instances of this type directly, use the
+// NewAlarmNotificationHandler function instead.
+type alarmNotificationHandlerBuilder struct {
+	logger         *slog.Logger
+	loggingWrapper func(http.RoundTripper) http.RoundTripper
+	cloudID        string
+	extensions     []string
```
jhernand, line 44: I believe extensions aren't needed here either. Can you check and remove them if they aren't used?

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	//structures for notification
+	subscriptionMapMemoryLock *sync.RWMutex
+	subscriptionMap           *map[string]data.Object
+	persistStore              *persiststorage.KubeConfigMapStore
+	subscriptionSearcher      *alarmSubscriptionSearcher
+	httpClient                http.Client
+}
+
+// NewAlarmNotificationHandler creates a builder that can then be used to configure and create a
+// handler for the collection of deployment managers.
```
jhernand, line 72: The comment isn't accurate.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+		SetJsonAPI(&jsonAPI).
+		SetClient(b.kubeClient)
+
+	// http client to send out notification
+	// use 2 sec first
+	httpClient := http.Client{Timeout: 2 * time.Second}
+
+	if singleAlarmNotificationHandle != nil {
+		panic("There is an existing handler instance")
+	}
```
jhernand, line 184: I think we should not do this check: it is perfectly OK to create multiple alarm notification handlers. Maybe one will receive notifications from Prometheus alert manager and another will receive them from something else, a hardware manager, for example.

If we really needed to ensure that there is only one instance it shouldn't be at this level, but at the level of the code that creates the handler. And to do it correctly we would need to be careful with how we do the check: using a nil check like here doesn't work well when there are multiple goroutines working in parallel.

But lets forget about that: just remove the check, I'd say.

jhernand: Ah, and we shouldn't be using "panic" anyhow, better return an error.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	singleAlarmNotificationHandle = result
+
+	b.logger.Debug(
+		"alarmNotificationHandler build:",
+		"CloudID", b.cloudID,
+	)
+
+	err = result.recoveryFromPersistStore(ctx)
```
jhernand, line 210: If we are doing additional work after creating the object, and that work may fail, like here, then it is better to use a temporary variable to make sure that we will never return a result that is half baked. For example:

```go
// Create and populate the object:
handler := &alarmNotificationHandler{...}

// Do other things with the "handler" variable, which mail fail.

// When all preparations succeeded copy the temporary variable
// to the result:
result = handler
return
```

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		b.logger.Error(
+			"alarmNotificationHandler failed to recovery from persistStore ",
+			slog.String(": ", err.Error()),
```
jhernand, line 214: This will return in JSON log message like this:

```json
{
  ": ": "the details of the error"
}
```

That looks weird. Can we use `slog.String("error", err.Error())` instead?

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+
+	}
+	return
+}
+
+func (h *alarmNotificationHandler) watchPersistStore(ctx context.Context) (err error) {
+	err = persiststorage.ProcessChangesWithFunction(h.persistStore, ctx, ProcessStorageChanges)
+
+	if err != nil {
+		panic("failed to launch watcher")
```
jhernand, line 248: Try to avoid panics, return the error instead.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+			"alarmNotificationHandler failed building the indexes ",
+			slog.String(": ", err.Error()),
+		)
+
+	}
+	return
+}
+
+func (h *alarmNotificationHandler) watchPersistStore(ctx context.Context) (err error) {
+	err = persiststorage.ProcessChangesWithFunction(h.persistStore, ctx, ProcessStorageChanges)
```
jhernand, line 245: `ProcessStorageChanges` should be a private method of the handler, something like this:

```go
func (h *alarmNotificationHandler) processStorageChanges(...) {
        ....
}
```

Then here you can pass that method as parameter:

```go
err = persiststorage.ProcessChangesWithFunction(h.persistStore, ctx, h.processStorageChanges)
```

internal/service/alarm_notification_handler_test.go
```diff
@@ -0,0 +1,169 @@
...
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2/dsl/core"
+	. "github.com/onsi/gomega"
+	"github.com/openshift-kni/oran-o2ims/internal/data"
+	"github.com/openshift-kni/oran-o2ims/internal/k8s"
+)
+
+var _ = Describe("alarm Notification handler", func() {
```
jhernand, line 26: Great to have tests from the beginning, thanks @Jennifer-chen-rh!

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	"log/slog"
+
+	"github.com/openshift-kni/oran-o2ims/internal/data"
+	"github.com/openshift-kni/oran-o2ims/internal/jq"
+	"github.com/openshift-kni/oran-o2ims/internal/search"
+)
+
+type subscriptionInfo struct {
+	filters                search.Selector
+	uris                   string
```
jhernand, line 29: I assume this `uris` is the callback URL, is it?

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	parser *search.SelectorParser) *alarmSubscriptionSearcher {
+	b.selectorParser = parser
+	return b
+}
+
+func newAlarmSubscriptionSearcher() *alarmSubscriptionSearcher {
+	return &alarmSubscriptionSearcher{}
+}
+
+func (b *alarmSubscriptionSearcher) build() {
```
jhernand, line 69: If we are going to use the builder pattern for this object then I'd appreciate if we do like in the other objects: separate types for the builder and the object.

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	consumerSubscriptionId string
+}
+
+// This file contains oran alarm notification serer search for matched subscriptions
+// at 1st step apply linear search
+type alarmSubscriptionSearcher struct {
+	logger *slog.Logger
+	jqTool *jq.Tool
+	//maps with prebuilt selector
+	subscriptionInfoMap *map[string]subscriptionInfo
```
jhernand, line 39: Why pointers to maps?

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	selectorParser, err := search.NewSelectorParser().
+		SetLogger(b.logger).
+		Build()
+	if err != nil {
+		b.logger.Error(
+			"failed to create filter expression parser: ",
+			slog.String("error: ", err.Error()),
+		)
+		return
+	}
```
jhernand, line 84: Looks like you are both passing a selector parser as parameter, and then building a new one here. I'd say it is better to build it here, so consider removing the `SetSelectorParser` method from the builder.

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	jqTool *jq.Tool) (err error) {
+
+	for key, value := range *subscriptionMap {
+
+		(*b.subscriptionInfoMap)[key] = subscriptionInfo{}
+
+		subInfo := (*b.subscriptionInfoMap)[key]
+		//get uris
+		var uris string
+		err = jqTool.Evaluate(`.callback`, value, &uris)
```
jhernand, line 132: He have been using jq to extract fields from objects because till know we didn't need to know much about the fields and what they mean. But I'd say that is different for subscriptions: this code needs to know what "callback", "filter" and "consumerSubscriptionId" mean. So I'd say we should instead have a new `Subscription` type (probably in a new `models` package) and then use it instead of using jq.

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+				" does not have filter included",
+			)
+		}
+
+		err = b.getSubFilters(filter, key)
+		if err != nil {
+			b.logger.Debug(
+				"pocessSubscriptionMapForSearcher ",
+				"subscription: ", key,
+				" error: ", err.Error(),
```
jhernand, line 161: Please consider using a human readable string as the log message. And the rest should be typical JSON field names. For example `subscription` or `error`; try to avoid spaces or punctuation.

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+	defer h.subscriptionMapMemoryLock.RUnlock()
+	result = *h.subscriptionSearcher.noFilterSubsSet
+
+	subCheckedSet := alarmSubIdSet{}
+
+	for path, subSet := range *h.subscriptionSearcher.pathIndexMap {
+
+		var alarmPath string
+		path = fmt.Sprintf(`.%s`, path)
+		err := h.jqTool.Evaluate(path, alarm, &alarmPath)
```
jhernand, line 180: What is this trying to achieve? Why can't we evaluate the filter on the complete alarm object?

internal/service/alarm_notification_searcher.go
```diff
@@ -0,0 +1,225 @@
...
+
+		if err == nil {
```
jhernand, line 182: Looks like if there is an error it is silently ignored. I'd suggest to first check if there is an error, handle it (maybe just log it) and then continue with the loop.

internal/service/alarm_notification_handler.go
Jennifer-chen-rh: I made it singleton for the configmap update call back. It is also workable to make global array/arrays. Not big deal.

For now I will change not panic but log an error. (change to handle multiple instances only when needed)

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: Yes, I will add the comments.

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: Yes

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: Maybe leftover from my attempts using reflect and unsafe package. I gave up as golang is not pointer/cast flexible. will change.

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: I am not sure we passed in the selector to handler. The parser was created in the Adapter codes that is part of server package and not accessible to handler in service package. Good thing the parser self just contains logger object and we only duplicate a pointer variable that point to the same logger object. As we are changing not using Adaptor for this handler, I think we need to keep like this now. Please let me know if you have other thought.

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: This is at different code path. let's discuss it on Wednesday

internal/persiststorage/persiststorage_configmap.go
Jennifer-chen-rh: You mean test.go? The subscription server will write the configmap. Alarm server will update the new map in the memory. I did test manually with 2 server instances (1 subscription, 1 notification) by adding/deleting the subscriptions. In the test.go, can we support 2 handler instances and mimic the configmap change? Please let me know.

internal/service/alarm_notification_handler.go
Jennifer-chen-rh: It is temperate for testConfigurationName. It will have a official name. Let me update this and propose a potential official name. Please see next commit/update.

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+	"github.com/openshift-kni/oran-o2ims/internal/search"
+)
+
+// singlton alarm notification handler
+var singleAlarmNotificationHandle *alarmNotificationHandler = nil
+
+// AlarmNotificationManagerHandlerBuilder contains the data and logic needed to create a new alarm notification
+// collection handler. Don't create instances of this type directly, use the
+// NewAlarmNotificationHandler function instead.
+type alarmNotificationHandlerBuilder struct {
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Actually it built. But will make the change

internal/service/alarm_notification_handler.go
Jennifer-chen-rh: Yes copy paste. Will update.

internal/service/alarm_notification_handler.go
Jennifer-chen-rh: will update

internal/service/alarm_notification_handler.go
```diff
@@ -0,0 +1,287 @@
...
+// Don't create instances of this type directly, use the NewAlarmNotificationHandler function
+// instead.
+type alarmNotificationHandler struct {
+	logger            *slog.Logger
+	loggingWrapper    func(http.RoundTripper) http.RoundTripper
+	cloudID           string
+	extensions        []string
+	jsonAPI           jsoniter.API
+	selectorEvaluator *search.SelectorEvaluator
+	jqTool            *jq.Tool
```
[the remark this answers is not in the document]
Jennifer-chen-rh: Yes, we use this and will keep it.

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: let's talk about this during code walkthrough

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: The filter came from the alarm subscription, and apply to the alert/alarm from alertmanager

internal/service/alarm_notification_searcher.go
Jennifer-chen-rh: Add the comment in the code. This is not error just not applied to the path

internal/cmd/server/start_alarm_notification_server.go
```diff
@@ -0,0 +1,319 @@
...
+		logger.Error(
+			"Failed to create adapter",
+			"error", err,
+		)
+		return exit.Error(1)
+	}
+	router.Handle(
+		"/o2ims-infrastructureMonitoring/{version}/alarmNotifications",
+		adapter,
+	).Methods(http.MethodPost)
```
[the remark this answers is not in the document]
Jennifer-chen-rh: let's discuss this more during code walk through.

README.md
```diff
@@ -354,5 +354,57 @@ $ curl -s http://localhost:8001/o2ims-infrastructureMonitoring/v1/alarmSubscript
...
 
+```
+$ curl -s -X POST --header "Content-Type: application/json" -d @subscription.json http://localhost:8000/o2ims-infrastructureMonitoring/v1/alarmSubscriptions
+```
+Above example will post an alarm subscription defined in subscription.json file 
+
 Inside _VS Code_ use the _Run and Debug_ option with the `start
 alarm-subscription-server` [configuration](.vscode/launch.json).
+<<<<<<< HEAD
+=======
```
jhernand, line 365: These ^  are leftovers from a merge conflict, fix and remove.

README.md
```diff
@@ -354,5 +354,57 @@ $ curl -s http://localhost:8001/o2ims-infrastructureMonitoring/v1/alarmSubscript
...
+#### Alarm Notification server
+
+The alarm-notification-server should use together with alarm subscription server. The alarm subscription sever accept and manages the alarm subscriptions. The alarm notificaton servers synch the alarm subscriptions via perisist storage. To use the configmap to persist the subscriptions, the namespace "orantest" should be created at hub cluster for now (will use official oranims namespace in future). The alarm notification server accept the alerts, match the subscription filter, build and send out the alarm notification based on url in the subscription.
+
+The required Resource server URL and token can be obtained as follows:
+
+```
+$ export RESOURCE_SERVER_URL=http://localhost:8002/o2ims-infrastructureInventory/v1/
+$ export RESOURCE_SERVER_TOKEN=$(
+  oc create token -n openshift-oauth-apiserver oauth-apiserver-sa --duration=24h
```
jhernand, line 377: What is the reason to use this service account? As this is for the development environment it is probably OK, but strange. Can we instead use `oc whoami --show-token`?

api/v1alpha1/orano2ims_types.go
```diff
@@ -64,8 +64,6 @@ type ResourceServerConfig struct {
 	//+optional
 	Extensions []string `json:"extensions,omitempty"`
 }
-
-// AlarmSubscriptionServerConfig contains the configuration for the alarm subscription server.
```
jhernand, line 68: Why are you removing this? It isn't a very informative comment, but doesn't hurt either.

internal/cmd/server/flags.go
```diff
@@ -133,11 +133,15 @@ func GetTokenFlag(
...
-	extensionsFlagName       = "extensions"
-	externalAddressFlagName  = "external-address"
+	backendTypeFlagName       = "backend-type"
+	backendTokenFlagName      = "backend-token"
+	backendTokenFileFlagName  = "backend-token-file"
+	backendURLFlagName        = "backend-url"
+	cloudIDFlagName           = "cloud-id"
+	extensionsFlagName        = "extensions"
+	externalAddressFlagName   = "external-address"
+	namespace                 = "namespace"
```
jhernand, line 143: To be consistent with other constants this should be `namespaceFlagName`. Same for the other constants below: should be `...FlagName` for consistency.

api/v1alpha1/orano2ims_types.go
Jennifer-chen-rh: This caused PR sanity check failure.

README.md
Jennifer-chen-rh: Addressed. I have the extension for read.me and it did not show there. Used plain editor to fix it.

api/v1alpha1/orano2ims_types.go
Jennifer-chen-rh: add back

internal/cmd/server/flags.go
Jennifer-chen-rh: addressed.

README.md
Jennifer-chen-rh: @danielerez This is related to alert to alarm mapping. Can you please answer Juan's question?

danielerez: Yeah, I guess it should work as well. But it'll require a bit more explanations

> What is the reason to use this service account? As this is for the development environment it is probably OK, but strange. Can we instead use `oc whoami --show-token`?

Yeah, I guess it should work as well. But it'll require a bit more explanation to login first, since an issue as follows could be encountered: https://access.redhat.com/solutions/6063651

internal/cmd/server/start_alarm_notification_server.go
```diff
@@ -0,0 +1,367 @@
...
+			ctx,
+			"Resource server URL is empty",
+			"flag", resourceServerTokenFlagName,
+		)
+		return exit.Error(1)
+	}
+
+	// Create the handler:
+	// Get the namespace:
+	o2imsNamespace, err := flags.GetString(namespaceFlagName)
```
jhernand, line 244: The `o2ims` prefix here isn't necessary, should just be `namespace`.

internal/cmd/server/start_alarm_notification_server.go
```diff
@@ -0,0 +1,367 @@
...
+	if err != nil {
+		logger.DebugContext(
+			ctx,
+			"Failed to get o2ims namespace flag",
+			"flag", o2imsNamespace,
+			"error", err.Error(),
+		)
```
jhernand, line 251: Missing `return exit.Error(1)`.

internal/cmd/server/start_alarm_notification_server.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	// Get the configmapName:
+	subscriptionsConfigmapName, err := flags.GetString(subscriptionConfigmapNameFlagName)
+	if err != nil {
+		logger.DebugContext(
+			ctx,
+			"Failed to get alarm subscription configmap name flag",
+			"flag", subscriptionsConfigmapName,
+			"error", err.Error(),
+		)
```
jhernand, line 265: Same, should halt the program.

internal/cmd/server/start_alarm_notification_server.go
```diff
@@ -0,0 +1,367 @@
...
+		logger.Error(
+			"Failed to create handler",
+			"error", err,
+		)
+		return exit.Error(1)
+	}
+
+	router.HandleFunc(
+		"/o2ims-infrastructureMonitoring/{version}/alarmNotifications",
+		handler.ProcessPost,
```
jhernand, line 293: I think this handler should implement the `SeveHTTP` method, and then here it should be like this:

```go
router.Handle(
        "...",
        handler,
)
```

internal/cmd/server/start_alarm_notification_server.go
Jennifer-chen-rh: Oh the namespace is treated optional here. Will use "default" when it is not supplied. Will change the log to debugContext.

internal/cmd/server/start_alarm_notification_server.go
Jennifer-chen-rh: Oh the namespace is treated optional here. Will use "default" when it is not supplied. Will change the log to debugContext.

internal/cmd/server/start_alarm_subscription_server.go
```diff
@@ -219,14 +229,42 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string)
...
 
-	// Create the handler:
+	// Get the namespace:
+	o2imsNamespace, err := flags.GetString(namespaceFlagName)
+	if err != nil {
+		logger.DebugContext(
+			ctx,
+			"Failed to get o2ims namespace flag",
+			"flag", o2imsNamespace,
+			"error", err.Error(),
```
jhernand, line 239: Try to use the type safe interface of `slog`:

```go
loggger.DebugContext(
        ctx,
        "Failed to ....",
        slog.String("flag", namespaceFlagName),
        slog.Strng("error", err.Error(),
)
```

Note that the value of the "flag" here should be the flag name: the `o2imsNamespace` variable will always be empty because inside this block it failed to get it.

internal/cmd/server/start_alarm_subscription_server.go
```diff
@@ -219,14 +229,42 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string)
 		return exit.Error(1)
 	}
 
-	// Create the handler:
+	// Get the namespace:
+	o2imsNamespace, err := flags.GetString(namespaceFlagName)
```
jhernand, line 233: Rename to `namespace`.

internal/persiststorage/persiststorage_configmap.go
```diff
@@ -249,3 +253,45 @@ func (s *KubeConfigMapStore) ProcessChanges(ctx context.Context, dataMap **map[s
 
 	return
 }
+
+func (s *KubeConfigMapStore) ProcessChangesWithFunction(ctx context.Context, function ProcessFunc) (err error) {
+	raw_opt := metav1.SingleObject(metav1.ObjectMeta{
```
jhernand, line 258: Try to use `camelCase` for variables: `rawOpt`.

internal/persiststorage/persiststorage_configmap.go
```diff
@@ -249,3 +253,45 @@ func (s *KubeConfigMapStore) ProcessChanges(ctx context.Context, dataMap **map[s
...
+		for {
+			event, open := <-watcher.ResultChan()
+			if open {
+				switch event.Type {
+				case watch.Added, watch.Modified:
+					configmap, _ := event.Object.(*corev1.ConfigMap)
+					newMap := map[string]data.Object{}
+					for k, value := range configmap.Data {
+						var object data.Object
+						err = (*s.jsonAPI).Unmarshal([]byte(value), &object)
```
jhernand, line 277: Why the explicit pointer dereference? Should work without it: `s.jsonAPI.Unmarshal(...)`

internal/service/alarm_fetcher.go
```diff
@@ -225,8 +224,19 @@ func (r *AlarmFetcher) FetchItems(
 		return
 	}
 
+	alarmMapper, err := NewAlarmMapper().
```
jhernand, line 227: This is great!

jhernand: And, can we create it once and reuse it?

internal/service/alarm_mapper.go
```diff
@@ -0,0 +1,332 @@
...
+	logger        *slog.Logger
+	backendClient *http.Client
+	//transportWrapper    func(http.RoundTripper) http.RoundTripper
+	resourceServerURL   string
+	resourceServerToken string
+	extensions          []string
+}
+
+// AlarmMapper creates a builder that can then be used to configure
+// and create a handler for the AlarmMapper.
```
jhernand, line 53: Review the comments: this isn't accurate, it talks about a "handler" but there is no handler here.

internal/service/alarm_mapper.go
```diff
@@ -0,0 +1,332 @@
...
+	result *AlarmMapper, err error) {
+	// Check parameters:
+	if b.logger == nil {
+		err = errors.New("logger is mandatory")
+		return
+	}
+	if b.backendClient == nil {
+		err = errors.New("backend client is mandatory")
+		return
+	}
```
jhernand, line 104: What about the resource server URL and token? Aren't those required?

internal/service/alarm_mapper.go
```diff
@@ -0,0 +1,332 @@
...
+		resourceServerURL:   b.resourceServerURL,
+		resourceServerToken: b.resourceServerToken,
+		extensions:          b.extensions,
+		jqTool:              jqTool,
+	}
+	return
+}
+
+// Map an Alert to an O2 Alarm object.
+func (r *AlarmMapper) MapAlertItem(ctx context.Context,
```
jhernand, line 144: As this is an alert mapper there is no need to repeat it in the name of the method. I'd suggest just `Map`. Alternatively you could just return the function from the builder, and then there is no need for another name.

internal/service/alarm_notification_handler.go
Jennifer-chen-rh: Changed

internal/persiststorage/persiststorage_configmap.go
Jennifer-chen-rh:
```internal/persiststorage/persiststorage_configmap.go:277:23: s.jsonAPI.Unmarshal undefined (type *jsoniter.API is pointer to interface, not interface)
make: *** [Makefile:292: binary] Error 1```
The persistent storage keep a pointer to the same jsonAPI object used by its owner handler

Comment (jhernand) [reply internal/persiststorage/persiststorage_configmap.go:277]: Ah, I understand now. `s.jsonAPI` should not be a pointer because the target is an interface and in Go variables of interface types are already pointers. So you should to change the `jsonAPI` field of the `KubeConfigMapStore` struct from `*jsoniter.API` to just `jsoniter.API`. Then here you can use directly `s.jsonAPI.Unmarshal(...)`.

File: internal/persiststorage/persiststorage_configmap.go [hunk 53]
```diff
@@ -15,8 +15,8 @@ import (
 )
 
 type KubeConfigMapStore struct {
-	Name       string
```
jhernand, line 18: I think it wasn't introduced in this patch, but I just noticed that this type isn't using the builder pattern like in all other types of the project. Please take note and fix it (in a different pull request).

internal/service/alarm_mapper.go
Jennifer-chen-rh: good point.

internal/persiststorage/persiststorage_configmap.go
Jennifer-chen-rh: OK. Will do that.

internal/service/alarm_mapper.go
Jennifer-chen-rh: @danielerez can you please answer Juan's question?

danielerez: So a validation for the URL would is indeed needed (as it's mandatory). But the token is optional, as it fallbacks to the backend token if missing. See:
* https://github.com/openshift-kni/oran-o2ims/blob/f9c1a5e7b17c1bafb32f471edb1092e8597fa8c9/internal/service/alarm_handler.go#L148
* https://github.com/openshift-kni/oran-o2ims/blob/ae2a0bb4765e273d06d9befdecf1ae463457716e/internal/cmd/server/start_alarm_server.go#L214

internal/service/alarm_mapper.go
```diff
@@ -0,0 +1,332 @@
...
+	resourceServerToken string
+	extensions          []string
+	jqTool              *jq.Tool
+}
+
+// AlarmMapperBuilder contains the data and logic needed to create a new AlarmMapper.
+type AlarmMapperBuilder struct {
+	logger        *slog.Logger
+	backendClient *http.Client
+	//transportWrapper    func(http.RoundTripper) http.RoundTripper
```
danielerez, line 46: Can be removed.

internal/service/alarm_mapper.go
danielerez: Right. Can be something like: NewAlarmMapper creates a builder that can be used to create a mapper of Alertmanager alerts to AlarmEventRecord objects.

internal/cmd/server/start_alarm_notification_server.go
Jennifer-chen-rh: changed

internal/cmd/server/start_alarm_notification_server.go
Jennifer-chen-rh: Changed.

internal/persiststorage/persiststorage_configmap.go
jhernand: I do have an strong opinion . Not that it introduces a bug, but it is really uncommon and unnecessary to use a pointer to an interface in general, and in this case in particular.

When you declare a variable like `var jsonAPI jsoniter.API` what the Go compiler will do is create a structure with two pointers: one pointer to the table describing the methods of the interface (know as an _itable_) and another pointer to the object itself. When you then do something like `jsonAPI = x` what you are doing is copying those two pointers, not creating a new object.

If you do `var jsonAPI *jsoniter.API` and then `jsonAPI = &x` what you are doing is copying the address of the `x` variable. That isn't necessary in general, and in particular not necessary in this case. I addition it means that the `x` object can't be stored in the stack because the Go compiler can't know if it will be used after the current scope is closed: that is known in Go as "escaping" to the heap. Finally it means that you can do `*jsonAPI = &y` and the result will be that the code that uses `x` would now surprisingly be using `y` instead.

Jennifer-chen-rh: @jhernand Thanks for long explanation. Really appreciate it! Will change it later today.

PR #103, opened 2024-06-18
Added test suite for servers API testing

Samudelacruz77:
test suite to test API responses from all servers


oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,326 @@
+package main_test
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"io"
+	"net/http"
+  "crypto/tls"
```
jhernand, line 10: Looks like this code hasn't been automatically formatted. Try to do so: `make fmt`

Samudelacruz77:
Ok, will do as requested

tshefi:
As Juan pointed out already adding readme would be great.

Samudelacruz77:
Added required folders and a README about usage

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
...
+	"testing"
+	"crypto/tls"
+	"io"
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestOranO2ims(t *testing.T) {
```
jhernand, line 15: Can you rename this to `TestAPI`?

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "OranO2ims Test Suite")
```
jhernand, line 17: And then here the name can just be `API`.

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+var _ = BeforeSuite(func() {
+	Expect(os.Getenv("TEST_HOST")).NotTo(BeZero(), "Please make sure TEST_HOST is set correctly.")
```
jhernand, line 21: At this point we don't have the infrastructure to run these tests in Prow or GitHub actions, so we will need to skip them when the environment variable isn't set, something like this:

```go
if os.Getenv("TEST_HOST") == "" {
        Skip("API tests were skipped because environment variable 'TEST_HOST' isn't set")
}
```

Can you do that change?

Samudelacruz77:
Yeah, I can do that.

test/api/oran_o2ims_suite_test.go
Samudelacruz77: Changed name to TestAPI

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
+package main_test
```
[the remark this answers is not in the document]
jhernand: And regarding the name of the file, would it make sense to have separate files for each major part of the API that is tested? For example, could we have these files?

- `suite_test.go` - For the setup of the test suite only (that is what we do in other packages).
- `metadata_test.go` - For tests of the metadata API.
- `resource_test.go` - For tests of the resources API.
- ...

I'd expect these tests to grow in the future, so having separate files can help reduce the size of each file.

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+var _ = BeforeSuite(func() {
+	Expect(os.Getenv("TEST_HOST")).NotTo(BeZero(), "Please make sure TEST_HOST is set correctly.")
+})
+
+var _ = Describe("Metadata Server API testing", func() {
+	Context("When getting infrastructure Inventory API version", func() {
+		It("should return OK in the response and json response should match reference json", func() {
+			requestBody := []byte(`{"uriPrefix":"/o2ims-infrastructureInventory/v1","apiVersions":[{"version":"1.0.0"}]}`)
+			request, _ := http.NewRequest("GET", "https://"+os.Getenv("TEST_HOST")+"/o2ims-infrastructureInventory/api_versions", bytes.NewBuffer([]byte(requestBody)))
```
jhernand, line 31: The code to get the environment variable is repeated may times. You may want to get it in the `BeforeSuite` block, and save it to a global variable:

```go
var testHost string

var _ = BeforeSuite(func() {
        testHost = os.Getenv("TEST_HOST")
        ...
}
```

Then use it int he tests instead of repeating the call to `os.Getenv`: it reduces the chances that you use the wrong environment variable name.

test/api/oran_o2ims_suite_test.go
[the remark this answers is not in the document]
Samudelacruz77: Ok, got on it, agree with the requirement

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,324 @@
...
+			request.Header.Set("Content-Type", "application/json")
+
+			By("Executing https petition")
+			tr := &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			}
+			client := &http.Client{Transport: tr}
```
jhernand, line 38: The creation of the transport and client is repeated multiple times. You may want to use a `BeforeEach` block to write it only once:

```go
var _ = Describe("...", func() {
        var client *http.Client

        BeforeEach(func() {
                tr := &http.Transport{...}
                client = &http.Client{
                        Transport: tr
                }
        })
}
```

Then you can use the `client` variable in the tests.

test/api/oran_o2ims_suite_test.go
```diff
@@ -0,0 +1,327 @@
...
+			response, err := client.Do(request)
+			Expect(err).NotTo(HaveOccurred())
+			By("Checking OK status response")
+			Expect(response.StatusCode).To(Equal(http.StatusOK))
+
+			By("Checking response JSON is equal to expected JSON")
+			responseBody, _ := io.ReadAll(response.Body)
+			Expect(responseBody).To(MatchJSON(requestBody))
```
jhernand, line 68: Checking complete JSON matches is fragile. For example, if a new version of the software adds a new field then this will fail. An alternative to that complete JSON matching can be to check for the presence and values of specific fields. For example, you can use our `MatchJQ` custom matcher to extract a field (using a jq expression) and check its value:

```go
Expect(responseBody).To(MatchJQ(`.name`, "OpenShift O-Cloud"))
```

You can have multiple likes like that ^ to check multiple fields. It may be material for a separate pull request, but please consider it.

test/api/alarm_test.go
```diff
@@ -0,0 +1,39 @@
...
+	BeforeEach(func() {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+		client = &http.Client{Transport: tr}
+	})
+	Context("When getting infrastructure Inventory API version", func() {
+		It("should return OK in the response and json response should match reference json", func() {
+			requestBody := []byte(``)
+			request, _ := http.NewRequest("GET", "https://"+os.Getenv("TEST_HOST")+"/o2ims-infrastructureInventory/api_versions", bytes.NewBuffer([]byte(requestBody)))
```
irinamihai, line 27: I second Juan's [comment](https://github.com/openshift-kni/oran-o2ims/pull/103#discussion_r1645795954) about moving `os.Getenv...` to `BeforeSuite`.

irinamihai, line 27: I think the tests would benefit in general from multi-line formatting:
```
http.NewRequest(
    "GET",
    "https://"+os.Getenv("TEST_HOST")+"/o2ims-infrastructureInventory/api_versions",
    bytes.NewBuffer([]byte(requestBody)
)
```

irinamihai, line 27: Please consider having the ORAN URIs (like `/o2ims-infrastructureInventory/api_versions`) as constants. Hardcoding them multiple times increases the risk of having typos.

test/api/alarm_test.go
```diff
@@ -0,0 +1,39 @@
...
+			request.Header.Set("Content-Type", "application/json")
+			By("Executing https petition")
+			response, err := client.Do(request)
+			Expect(err).NotTo(HaveOccurred())
+			By("Checking OK status response")
+			Expect(response.StatusCode).To(Equal(http.StatusOK))
+			By("Checking response JSON matches condition")
+			responseBody, _ := io.ReadAll(response.Body)
+			Expect(responseBody).To(MatchJQ(`version`, "1.0.0"))
```
irinamihai, line 36: The version can also be a constant.

irinamihai:
@Samudelacruz77 , is this PR to set up the initial framework? If so, I think it should be clearly stated in the PR's name / 1st comment.

In my opinion, if any parts of the tests are as per WG6 standard, the specific chapter should be specified + the WG6 chapter that's actually being tested.

Kindly link the Jira # as well.

test/api/oran_o2ims_suite_test.go
Samudelacruz77: Done

test/api/alarm_test.go
Samudelacruz77: Done

Samudelacruz77: Done

test/api/alarm_test.go
Samudelacruz77: Done

Samudelacruz77:
> @Samudelacruz77 , is this PR to set up the initial framework? If so, I think it should be clearly stated in the PR's name / 1st comment.
> 
> In my opinion, if any parts of the tests are as per WG6 standard, the specific chapter should be specified + the WG6 chapter that's actually being tested.
> 
> Kindly link the Jira # as well.

Is not the case for now, we are only having the curls in the readme and demos in a test suite, on next stages telco team will have test as per specifications, we are testing right now the integration with openshift as per individual components

donpenney:
Closing as stale

PR #107, opened 2024-06-21
MGMT-16066: Resource Subscription Server

irinamihai:
Description:
- update the alarm subscription handler to a generic subscription handler that can also be used for resource subscriptions
- add a new command for the resource subscription server:
`oran-o2ims start infrastructure-inventory-subscription-server `
- update the alarm and resource subscription servers to also set the subscription type


internal/service/constants.go
```diff
@@ -0,0 +1,13 @@
...
+
+const (
+	TestNamespace               = "orantest"
+	AlarmConfigMapName          = "oran-alarms-sub"
+	InfraInventoryConfigMapName = "oran-infra-inventory-sub"
+	FieldOwner                  = "oran-o2ims"
+)
+
+const (
+	SUBSCRPTION_TYPE_ALARM                     = "ALARM"
```
jhernand, line 11: Should these be `SubscriptionTypeAlarm` and `SubscriptionTypeInfrastructureInventory` to be consistent with the typical naming for constants in Go?

internal/cmd/server/start_alarm_subscription_server.go
```diff
@@ -220,12 +220,13 @@ func (c *AlarmSubscriptionServerCommand) run(cmd *cobra.Command, argv []string)
...
 
 	// Create the handler:
-	handler, err := service.NewAlarmSubscriptionHandler().
+	handler, err := service.NewSubscriptionHandler().
 		SetLogger(logger).
 		SetLoggingWrapper(loggingWrapper).
 		SetCloudID(cloudID).
 		SetExtensions(extensions...).
 		SetKubeClient(kubeClient).
+		SetSubscriptionType("ALARM").
```
jhernand, line 229: I wonder if the storage should be created outside of the handler and then passed as a parameter instead of forcing the subscription handler to know the type of subscription. Is there any difference in the behaviour other than the namespace/name of the configmap where the subscriptions are stored?

Jennifer-chen-rh: I made the configuration map configurable in my current PR. but not committed.

internal/persiststorage/persiststorage_configmap.go
```diff
@@ -206,8 +206,8 @@ func (s *KubeConfigMapStore) ReadAllEntries(ctx context.Context) (result map[str
 
 func (s *KubeConfigMapStore) ProcessChanges(ctx context.Context, dataMap **map[string]data.Object, lock *sync.Mutex) (err error) {
 	raw_opt := metav1.SingleObject(metav1.ObjectMeta{
-		Namespace: s.nameSpace,
```
Jennifer-chen-rh, line 209: I made similar change in my current notification server PR.

internal/cmd/server/start_alarm_subscription_server.go
irinamihai: There are also 2 other instances where we need to translate the reponse. The Alarm and Infrastructure subscription responses differ a bit. In [decodeSubId](https://github.com/openshift-kni/oran-o2ims/pull/107/files#diff-8bac4a7cd1e160e980214c0ee0cad277404fb16c3891ddca485a411965c31502R484-R491) and [encodeSubId](https://github.com/openshift-kni/oran-o2ims/pull/107/files#diff-8bac4a7cd1e160e980214c0ee0cad277404fb16c3891ddca485a411965c31502R447-R471). In these cases the handler needs to know the subscription type so that it translates correctly.

irinamihai: @Jennifer-chen-rh , that's great. Would it be ok if we tried to merge the current PR first? Then yours will actually be shorter. The rebase shouldn't be a headache.

Jennifer-chen-rh: Sure. Mine has more changes and comparatively harder to be a rebase code base.

jhernand:
Looks good to me.

internal/service/subscription_handler.go
```diff
@@ -403,59 +428,74 @@ func (h *alarmSubscriptionHandler) deleteToSubscriptionMap(key string) {
...
-			"callback": .callback,
-			"filter": .filter
-		}`,
-		input, &output,
-		jq.String("$alarmSubId", subId),
-	)
+
+	err = nil
+
+	if h.persistStore.Name == AlarmConfigMapName {
```
danielerez, line 447: nit: we can consider switch/case here for future expansion

internal/service/subscription_handler.go
```diff
@@ -403,59 +428,74 @@ func (h *alarmSubscriptionHandler) deleteToSubscriptionMap(key string) {
...
+				"filter": .filter
+			}`,
+			input, &output,
+			jq.String("$alarmSubId", subId),
+		)
+	} else if h.persistStore.Name == InfraInventoryConfigMapName {
+		// get consumer name, subscriptions
+		err = h.jqTool.Evaluate(
+			`{
+				"subscriptionId": $subId,
```
danielerez, line 463: Just to keep the convention, maybe we can rename it to something like `$resourceSubId`

internal/service/subscription_handler.go
```diff
@@ -403,59 +428,74 @@ func (h *alarmSubscriptionHandler) deleteToSubscriptionMap(key string) {
...
+				"callback": .callback,
+				"filter": .filter
+			}`,
+			input, &output,
+			jq.String("$subId", subId),
+		)
+	}
+
 	if err != nil {
 		return
```
danielerez, line 474: Do we want some log here? If not, seems lines 473-475 are redundant.

internal/service/subscription_handler.go
```diff
@@ -403,59 +428,74 @@ func (h *alarmSubscriptionHandler) deleteToSubscriptionMap(key string) {
...
-			"alarmSubscriptionId": $alarmSubId,
-			"consumerSubscriptionId": .consumerSubscriptionId,
-			"callback": .callback,
-			"filter": .filter
-		}`,
-		input, &output,
-		jq.String("$alarmSubId", subId),
-	)
+
+	err = nil
```
danielerez, line 445: Just for readability?

internal/cmd/server/start_infrastructure_inventory_subscription_server.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// Create the handler:
+	handler, err := service.NewSubscriptionHandler().
+		SetLogger(logger).
+		SetLoggingWrapper(loggingWrapper).
+		SetCloudID(cloudID).
+		SetExtensions(extensions...).
+		SetKubeClient(kubeClient).
+		SetSubscriptionType("INFRA-ENV").
```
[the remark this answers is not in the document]
irinamihai: You're right. Initially I saw they were in different packages, but I noticed the service package is already imported here, so I will use it. Thanks!

internal/service/subscription_handler.go
irinamihai: I  wanted to keep it in line with the O-RAN standard (where it's `subscriptionId`) and the alarm subscription case that's using `alarmSubscriptionId`, just as in the standard.
Please let me know if you agree.

internal/service/subscription_handler.go
irinamihai: Will remove.

internal/service/subscription_handler.go
danielerez: Oh, I was actually referring to the name of the variable (`$subId`). It just looked a bit weird at first glance since it's generic compared to `$alarmSubId`. But feel free to ignore, it makes better sense now.

internal/service/subscription_handler.go
```diff
@@ -403,59 +428,72 @@ func (h *alarmSubscriptionHandler) deleteToSubscriptionMap(key string) {
...
 }
 
-func (h *alarmSubscriptionHandler) decodeSubId(ctx context.Context,
+func (h *SubscriptionHandler) decodeSubId(
 	input data.Object) (output string, err error) {
-	//get cluster name, subscriptions
-	err = h.jqTool.Evaluate(
-		`.alarmSubscriptionId`, input, &output)
+
+	err = nil
```
danielerez, line 480: Same comments from `encodeSubId` - consider to apply also for this func.

PR #111, opened 2024-06-27
Provisioning POC: Operator boilerplate

irinamihai:
Description:
- Add the boilerplate for the operator provisioning logic:
  * ClusterTemplate & ClusterRequest CRs
  * ClusterTemplate & ClusterRequest controllers
 
CRD samples:
```
$  oc get clustertemplate clustertemplate-sample -oyaml
apiVersion: oran.openshift.io/v1alpha1
kind: ClusterTemplate
metadata:
  name: clustertemplate-sample
  namespace: oran-o2ims
spec:
  ...
   

$  oc get clusterrequest clusterrequest-sample -oyaml
apiVersion: oran.openshift.io/v1alpha1
kind: ClusterRequest
metadata:
  name: clusterrequest-sample
  namespace: oran-o2ims
spec:
  ...
```

From operator logs:
```
{"time":"2024-06-26T22:07:57Z","level":"INFO","msg":"Starting Controller","controller":"orano2ims-cluster-template","controllerGroup":"oran.openshift.io","controllerKind":"ClusterTemplate"}
{"time":"2024-06-26T22:07:58Z","level":"INFO","msg":"Starting workers","controller":"orano2ims-cluster-request","controllerGroup":"oran.openshift.io","controllerKind":"ClusterRequest","worker count":1}
{"time":"2024-06-26T22:07:58Z","level":"INFO","msg":"Starting workers","controller":"orano2ims-cluster-template","controllerGroup":"oran.openshift.io","controllerKind":"ClusterTemplate","worker count":1}
```

/cc @bartwensley @Missxiaoguo @browsell


jhernand:
Looks like Nephio is proposing to put the "ClusterRequest" type in the "o2ims.provisioning.oran.org" group. Doesn't look completely correct, should rather be "provisioning.o2ims.oran.org". But I'd suggest you don't use an "openshift.io" group for this, as it is supposed eventually be part of the specification.

For the same reason I'd suggest you create a separate Go module for these types, that way people writing clients in Go will be able to import just the API types, without having to bring in all the other dependencies that we use for the implementation.

Another thing you may want to consider is having two CRDs for each type, in different groups. For example an external "ClusterTemplate.provisioning.o2ims.oran.org" and an internal "ClusterTemplate.oran.openshift.io". That will give you the freedom to put whatever you need in the spec and status of the internal template, while keeping in external template backwards compatible and containing only the bare minimum that is needed by the client. It also makes it possible to have different permissions. For example, the client will need to read the external templates, but it doesn't need to have permissions to see the internal details. The controller can watch the internal cluster template, and automatically create the external cluster template, containing only the attributes that the client needs. Same for the cluster request, but in the other direction: the client creates the external request and the controller creates the internal request (if it needs additional information in the status or spec).

irinamihai:
@jhernand , thank you for the suggestions.
This is currently for an initial POC (which is also why it's on a separate branch). I think we can leave them as is until we have more information on what the standards decide for the API group naming.
Once that part is decided I will include your suggestions.

PR #117, opened 2024-07-03
NO-ISSUE: Move selector logic from handlers to adapter

jhernand:
Currently we apply the selector (the `filter` query parameter) in all the handlers. This patch moves the logic to the adapter, which is where it belongs.


danielerez (approved):
Nice! It indeed simplifies the handlers and removes redundant code:)
Maybe we can think of a similar improvement for the `jsonAPI` that we're creating in every handler. I actually think it mostly can be removed(?)

/lgtm

PR #118, opened 2024-07-04
CNF-13446: ClusterRequest schema match with template

irinamihai:
Description:
- match the clusterTemplateInput with the inputDataSchema of the clusterTemplateRef
- automatically trigger clusterRequest reconciliation if clusterTemplates used by clusterRequests are updated or deleted
- unit tests


internal/controllers/clusterrequest_controller.go
```diff
@@ -108,7 +117,31 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (nextReconcile c
 	if err != nil {
 		t.logger.ErrorContext(
 			ctx,
-			"Failed to update status for ClusterRequest",
+			"Failed to update the ClusterTemplateInputValidation status for ClusterRequest",
+			slog.String("name", t.object.Name),
+		)
+		nextReconcile = ctrl.Result{RequeueAfter: 30 * time.Second}
```
Missxiaoguo, line 123: I think if `err` exists, it requeues on the error regardless of the reconcile time set here. Is this expected or should be err set to nil to requeue on the interval?

irinamihai: I actually think it would make sense to not even reconcile if there's an error and wait for the user to update either the ClusterRequest or the used ClusterTemplate which would trigger a reconciliation anyway. I'll make the update.

internal/cmd/operator/start_controller_manager.go
```diff
@@ -124,22 +123,22 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
...
-	defaultNamespaces := make(map[string]cache.Config)
+	//namespaces := [...]string{"default", "oran", "o2ims", "oran-o2ims"} // List of Namespaces
+	//defaultNamespaces := make(map[string]cache.Config)
 
-	for _, ns := range namespaces {
-		defaultNamespaces[ns] = cache.Config{}
-	}
+	//for _, ns := range namespaces {
+	//	defaultNamespaces[ns] = cache.Config{}
+	//}
```
bartwensley, line 131: @irinamihai - I assume this was done because we're using a new namespace for each cluster template. Is this going to be an issue? Do we need to look at what would be required to put the cluster templates all in one of the ims namespaces?

irinamihai: > this was done because we're using a new namespace for each cluster template
Yes. And these lines were restricting the namespaces that are watched by the controllers.

It wasn't a requirement to have just these namespaces, so I think we can remove these lines completely in the future.

PR #119, opened 2024-07-07
NO-ISSUE: remove redundant jsonAPI from handlers

danielerez:
Removed unsed `jsonAPI` from handlers (can be added ad-hoc when needed).


jhernand:
Very nice, thanks @danielerez.

/approve
/lgtm

PR #123, opened 2024-07-10
NO-ISSUE: Create only one image

jhernand:
Currently we create two images:

- `quay.io/openshift-kni/oran-o2ims` - For the servers.
- `quay.io/openshift-kni/oran-o2ims-operator` - For the operator.

But both contain the same `oran-o2ims` binary. The servers and the operator are just sub-commands of that binary. For example, the deployment manager server is the `start deployment-manager-server` sub-command and the operator is the `start controller-manager` command.

To simplify things this patch removes one of the images. The preserved image name is `oran-o2ims-operator`.

Related: https://github.com/openshift/release/pull/54191


donpenney:
Due to the way the CI works, I think we should probably do this in multiple steps.
https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image

Since the root image build will use the docker files from HEAD, I think it would be safest to leave the Containerfile and Containerfile.tools files in place in this PR to get the first step merged. Then we should be able to merge the corresponding openshift/release PR https://github.com/openshift/release/pull/54191 safely. After that, we can have a followup PR here to delete the Containerfile and Containerfile.tools, once they're no longer being referenced by the CI.

donpenney:
These changes have been rebased and adopted in https://github.com/openshift-kni/oran-o2ims/pull/187

PR #127, opened 2024-07-24
CNF-13706: [POC2] ACM PolicyGenerator Schema

irinamihai:
Description:
- sample for ACM policy generator that would be used in the ClusterTemplate spec.inputDataSchema.policyTemplateSchema
- the current sample is subject to change

### DIFF
--- a/docs/samples/policyTemplateSchema.yaml
+++ b/docs/samples/policyTemplateSchema.yaml
@@ -1,0 +1,73 @@
+# All the values from this schema are of string type as they will
+# end up as key:value pairs in a ConfigMap.
+description: ACM PolicyGenerator schema
+properties:
+
+  # site specific config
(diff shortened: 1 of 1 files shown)


docs/samples/policyTemplateSchema.yaml
```diff
@@ -0,0 +1,73 @@
+# All the values from this schema are of string type as they will
+# end up as key:value pairs in a ConfigMap.
+description: ACM PolicyGenerator schema
+properties:
+
+  # site specific config
+  site: # this is the name of the ConfigMap holding site specific configuration
```
Missxiaoguo, line 7: Based on the CaaS provision doc, it seems like this top level field represents both the configmap name and  cluster label.

Missxiaoguo (commented):
/lgtm

PR #131, opened 2024-07-26
CNF-13710: Add ClusterTemplate validation and status condition

Missxiaoguo:
- `ClusterTemplateValidated` status condition added for cluster template
- Utility function added for configmap validation
- xvalidation rule added to reject any spec updates to the cluster template that has been validated
- Set the referenced configmaps to immutable for a valid cluster template
- Failed validations that would cause `ClusterTemplateValidated` status to `False`:
            - `ClusterInstanceDefaults` configmap referenced in the cluster template is missing
            - `ClusterInstanceDefaults` configmap is explictilty set to Immutable:false
            - `ClusterInstanceDefaults` configmap is missing expected key
            - The data of `ClusterInstanceDefaults` configmap is not a valid YAML format
 - Reconcilation tiggers for cluster template
            - Creation of cluster template and `spec` updates to the cluster template
            - Creation, deletion, or `data` updates the referenced configmaps (Because of the added validation rule, APIserver rejects the data updates if the cluster template is valid already)
            - Auto-requeue for invalid cluster template after 5mins just in case that event was not triggered somehow
- Unittests added


internal/controllers/clustertemplate_controller.go
```diff
@@ -96,31 +121,177 @@ func (r *ClusterTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	} else {
+		// Check if the expected key is present in the configmap data
+		defaults, exists := existingConfigmap.Data[utils.ClusterInstanceTemplateDefaultsConfigmapSuffix]
+		if !exists {
+			validationErr = fmt.Sprintf(
+				"the expected key %s does not exist in the ConfigMap %s data", expectedKey, name)
+			return validationErr, nil
+		}
+
+		// Check if the data value is either valid JSON or YAML
```
bartwensley, line 219: Just wondering why we would still allow JSON if we are recommending YAML format?

Missxiaoguo: No parituclar reason. It supported both json and yaml string. I just didn't change it. Sure, we could reject JSON string and only allow YAML string in the configmap data value.

Missxiaoguo: I changed to use `yaml.Unmarshal` to parse the string, but I just found that yaml parser in Go can handle json strings as well that means any valid Json string can also be parsed as valid yaml string by `yaml.Unmarshal`. It's a superset of Json.  We should just simply remove the JSON word from the comment/error message for avoid misleading.

api/v1alpha1/clustertemplate_types.go
```diff
@@ -51,26 +51,19 @@ type InputDataSchema struct {
...
-	// Tells if the schema provided in spec.inputDataSchema is a valid JSON.
-	ClusterTemplateValidation ClusterTemplateValidation `json:"clusterTemplateValidation"`
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 
 // ClusterTemplate is the Schema for the clustertemplates API
+// +kubebuilder:validation:XValidation:message="Spec changes are not allowed for the ClusterTemplate that has passed the validation", rule="!has(oldSelf.status) || oldSelf.status.conditions.exists(c, c.type=='ClusterTemplateValidated' && c.status=='False') || oldSelf.spec == self.spec"
```
bartwensley, line 66: nit - wording: ...for a ClusterTemplate that has passed validation...

internal/controllers/clustertemplate_controller.go
```diff
@@ -96,31 +122,181 @@ func (r *ClusterTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			utils.CTconditionReasons.Failed,
+			metav1.ConditionFalse,
+			errMsg,
+		)
+	} else {
+		utils.SetStatusCondition(&t.object.Status.Conditions,
+			utils.CTconditionTypes.Validated,
+			utils.CTconditionReasons.Completed,
+			metav1.ConditionTrue,
+			"The cluster template validation is succeeded",
```
bartwensley, line 291: nit: remove "is"

bartwensley (approved):
Looks good

api/v1alpha1/clustertemplate_types.go
Missxiaoguo: Will update in another PR

internal/controllers/clustertemplate_controller.go
Missxiaoguo: Will update in another PR

internal/controllers/clustertemplate_controller.go
```diff
@@ -96,31 +122,181 @@ func (r *ClusterTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+}
+
+// validateUniqueConfigmapReferences ensures that the Configmaps referenced in the ClusterTemplate are unique across all ClusterTemplates.
+// The required unique configmap - ClusterInstanceDefaults
+func (t *clusterTemplateReconcilerTask) validateUniqueConfigmapReferences(ctx context.Context) (string, error) {
+	var validationErr string
+	var ciDefaultsCmCts []string
+
+	clusterTemplates := &oranv1alpha1.ClusterTemplateList{}
+	err := t.client.List(ctx, clusterTemplates)
```
irinamihai, line 254: I think here we should get the clusterTemplates in the same namespace rather than all the existing clusterTemplates in the cluster.

Missxiaoguo:
/hold
updating unit tests

Missxiaoguo:
/unhold

unittests updated as well

PR #133, opened 2024-08-01
CNF-13707 [POC2] ClusterRequest controller logic for PG values

irinamihai:
- `policyTemplateDefaults` added to `clusterTemplate.spec.templates` to point to a configMap in the same namespace as the ClusterTemplate to hold default values for the templated PGs
- `policyTemplateInput` added to `clusterRequest.spec.clusterTemplateInput` to override default values
- `policyTemplateSchema` added to `clusterTemplate.spec.inputDataSchema` to define the schema for the PGs values
- a new configMap is generated for each ClusterRequest with the name `<cluster-request>-pg`
- each data entry of the configMap will contain the original key and the version of the PG (contained by the `<cluster-template>-policy` label on the managed cluster)
- the `<cluster-template>-policy`  and `cluster-version` labels are mandatory
- the namespace for the policies and the associated ConfigMap is `ztp-<base-cluster-name>-<cluster-version>`

In the PolicyGenerator file, the following template can be used:
`vlan: '{{hub fromConfigMap "" (printf "%s-pg" .ManagedClusterName) (printf "%s-sriov-network-vlan-1" (index .ManagedClusterLabels "cluster-template-a-policy")) | toInt hub}}'` where `cluster-template-a` is the name of the clusterTemplate


internal/controllers/clusterrequest_controller.go
```diff
@@ -652,26 +695,34 @@ func (t *clusterRequestReconcilerTask) getCrClusterTemplateRef(ctx context.Conte
...
 				t.object.Spec.ClusterTemplateRef, jsonString))
 		return err
 	}
 
 	// Check that the clusterTemplateInput matches the inputDataSchema from the ClusterTemplate.
-	err = utils.ValidateJsonAgainstJsonSchema(
-		string(clusterTemplateRef.Spec.InputDataSchema.ClusterInstanceSchema.Raw), string(jsonString))
+	if schemaType == utils.ClusterInstanceSchema {
+		err = utils.ValidateJsonAgainstJsonSchema(
+			string(clusterTemplateRef.Spec.InputDataSchema.ClusterInstanceSchema.Raw), string(jsonString))
```
bartwensley, line 721: I thought the clusterInstanceSchema was now using YAML? Should ValidateJsonAgainstJsonSchema be renamed?

PR #139, opened 2024-08-09
Set transport defaults when creating HTTP backend transport

alegacy:
The zero value of HttpTransport doesn't include any default attribute values for setting up the HTTP transport.  This prevents using the HTTPS_PROXY environment variable (and other related proxy variables) to redirect HTTP requests to a proxy.

We could set this explicitly on our own, but that wouldn't pick up any future changes to the defaults made upstream in the library.

This is useful when testing as a standalone binary and connecting to a system that is behind a proxy.

### DIFF
--- a/internal/authentication/handler_wrapper.go
+++ b/internal/authentication/handler_wrapper.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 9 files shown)


bartwensley (approved):
Seems reasonable to me.

PR #145, opened 2024-08-13
Exclude vendor directory from gofmt input parameters

alegacy:
Now that the `vendor` directory is no longer excluded from git we need to adjust the `gofmt` input parameters to exclude it since it currently looks at any and all *.go files within the entire repo.  This causes it to run on files within the vendor directory which, as it turns out, contains some *.go files that do not conform to the go formatting guidelines.

Alternatively we could have explicitly listed the required source directories, but figured that this dynamic approach would be more "future-proof" and require less code maintenance.


donpenney (commented):
/lgtm

PR #159, opened 2024-08-13
CNF-13955: Enable golangci-lint gocritic test

donpenney:
This update addresses a variety of issues identified by the golangci-lint gocritic test.

PR #161, opened 2024-08-14
CNF-13978: Fix some more gocritic lints

leo8a:
Follow-up PR of #159. Ran the `ci-job` target in my env and pushed some more fixes that were highlighted.

/cc @donpenney


internal/controllers/utils/constants.go
```diff
@@ -7,20 +7,20 @@ const (
...
-	ORANO2IMSMetadata          = "metadata"
-	ORANO2IMSDeploymentManager = "deployment-manager"
-	ORANO2IMSResource          = "resource"
-	ORANO2IMSAlarmSubscription = "alarm-subscription"
-	ORANO2IMSAlarmNotification = "alarm-notification"
+	ORANO2IMSMetadata          = "metadata-server"
+	ORANO2IMSDeploymentManager = "deployment-manager-server"
+	ORANO2IMSResource          = "resource-server"
+	ORANO2IMSAlarmSubscription = "alarm-subscription-server"
+	ORANO2IMSAlarmNotification = "alarm-notification-server"
```
donpenney, line 14: Why are these being modified?

leo8a: linting was complaining about those in my env with below issue

```
internal/controllers/utils/constants.go:19:61: string `-server` has 5 occurrences, make it a constant (goconst)
        ORANO2IMSMetadataServerName          = ORANO2IMSMetadata + "-server"
```  

version in my env is
```
-> golangci-lint --version
golangci-lint has version 1.55.2 built with go1.21.3 from e3c2265f on 2023-11-03T12:59:25Z
```

donpenney: The solution to that, though, would be to add a serverSuffix const that gets used in place of "-server", rather than changing the base const value

donpenney (approved):

PR #162, opened 2024-08-14
NO-ISSUE: Use dev script to pin golangci-lint version for users

donpenney:


hack/golangci-lint.sh
```diff
@@ -0,0 +1,42 @@
...
+    curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -d -b "${bindir}" "v${VERSION}"
+
+    if [ $? -ne 0 ]; then
+        echo "Install from script failed. Trying go install"
+        go install "github.com/golangci/golangci-lint/cmd/golangci-lint@v${VERSION}"
+        if [ $? -ne 0 ]; then
+            echo "Install of golangci-lint failed"
+            exit 1
+        else
+            golangci_lint=$(which golangci-lint)
```
alegacy, line 26: This location may not necessarily be where `go install` put it into... should be, but not necessarily.

donpenney: Updated to use `go env` to get the path

PR #170, opened 2024-08-16
CNF-13963: golangci-lint: Enable wrapcheck in linter configs

leo8a:
Enables wrapcheck in `.golangci.yml` and fixes issues highlighted by the linter.

/cc @donpenney @alegacy

### DIFF
--- a/internal/controllers/clusterrequest_controller.go
+++ b/internal/controllers/clusterrequest_controller.go
@@ -50,7 +50,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-// clusterRequestReconciler reconciles a ClusterRequest object
+// ClusterRequestReconciler reconciles a ClusterRequest object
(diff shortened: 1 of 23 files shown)


internal/controllers/clusterrequest_controller.go
```diff
@@ -1506,20 +1527,21 @@ func (t *clusterRequestReconcilerTask) updateHardwareProvisioningStatus(
...
 			t.logger.ErrorContext(
 				ctx,
 				"Failed to update the NodePool status",
 				slog.String("name", nodePool.Name),
+				slog.Any("specificError", err),
 			)
+			return fmt.Errorf("failed to update NodePool status: %w", err)
 		}
 	}
 	return err
```
donpenney, line 1540: Now that you've got errors being returned in the previous code blocks, this should just be `return nil`, I think?

donpenney (commented):
/lgtm

PR #177, opened 2024-08-20
CNF-13958: Enable TLS verification for in-cluster deployments

alegacy:
This configures the TLS clients using the service account root CA bundles so that we can properly verify the identity of backend servers.

### DIFF
--- a/internal/controllers/utils/utils.go
+++ b/internal/controllers/utils/utils.go
@@ -3,13 +3,19 @@ package utils
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
 	"slices"
 	"strconv"
 	"strings"
 	"text/template"
+
+	"k8s.io/apimachinery/pkg/util/net"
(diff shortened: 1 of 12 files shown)


internal/controllers/utils/utils.go
```diff
@@ -720,3 +725,60 @@ func GetTLSSkipVerify() bool {
...
+		// should have disabled the TLS verification which would prevent this function from being invoked.
+		return fmt.Errorf("failed to read CA bundle '%s': %w", defaultBackendCABundle, err)
+		// This should not happen, but if it does continue anyway
+	} else {
+		// This will enable accessing public facing API endpoints signed by the default ingress controller certificate
+		config.RootCAs.AppendCertsFromPEM(data)
+	}
+
+	if data, err := os.ReadFile(defaultServiceCAFile); err != nil {
+		return fmt.Errorf("failed to read service CA rile '%s': %w", defaultServiceCAFile, err)
```
irinamihai, line 746: Typo: `rile`

alegacy: thanks, fixed.

donpenney (commented):
/lgtm

PR #178, opened 2024-08-20
Add NodePool Provisioning Status Condition Constants

tliu2021:
Use a single condition type for tracking the NodePool provisioning status, with various condition reasons to represent different states in the provisioning process. This simplifies the condition handling by focusing on one type with multiple reasons for clarity and consistency.

### DIFF
--- a/api/hardwaremanagement/v1alpha1/conditions.go
+++ b/api/hardwaremanagement/v1alpha1/conditions.go
@@ -23,3 +23,22 @@ const (
 	// fulfilled.
 	FailedCondition = "Failed"
 )
+
+type ConditionType string
+
+// The following constants define the different types of conditions that will be set
+const (
+	Provisioned ConditionType = "Provisioned"
+	Unknown     ConditionType = "Unknown" // indicates the condition has not been evaluated
+)
+
+type ConditionReason string
+
+// The following constants define the different reasons that conditions will be set for
+const (
+	InProgress     ConditionReason = "InProgress"
+	Completed      ConditionReason = "Completed"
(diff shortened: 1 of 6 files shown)


Missxiaoguo (approved):

api/hardwaremanagement/v1alpha1/conditions.go
```diff
@@ -23,3 +23,22 @@ const (
...
+type ConditionReason string
+
+// The following constants define the different reasons that conditions will be set for
+const (
+	InProgress     ConditionReason = "InProgress"
+	Completed      ConditionReason = "Completed"
+	Unprovisioned  ConditionReason = "Unprovisioned"
+	Failed         ConditionReason = "Failed"
+	NotInitialized ConditionReason = "NotInitialized"
+)
```
donpenney, line 44: These would be more convenient if setup similar to the condition types and reasons as in:
https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/utils/conditions.go#L8-L48

Something like:
```
// ConditionReason is a string representing the condition's reason
type ConditionReason string

// The following constants define the different reasons that conditions will be set for the resource
var ResourceConditionReasons = struct {
	Completed  ConditionReason
	InProgress ConditionReason
	Failed     ConditionReason
}{
	Completed:  "Completed",
	InProgress: "InProgress",
	Failed:     "Failed",
}
```

Rather than a random collection of constants, they're grouped and referenced like:
```
hwmgmtv1alpha1.ResourceConditionReasons.InProgress
```

PR #187, opened 2024-08-29
NO-ISSUE: Create only one image

donpenney:
Currently we create two images:

- `quay.io/openshift-kni/oran-o2ims` - For the servers.
- `quay.io/openshift-kni/oran-o2ims-operator` - For the operator.

But both contain the same `oran-o2ims` binary. The servers and the operator are just sub-commands of that binary. For example, the deployment manager server is the `start deployment-manager-server` sub-command and the operator is the `start controller-manager` command.

To simplify things this patch removes one of the images. The preserved image name is `oran-o2ims-operator`.

Related: https://github.com/openshift/release/pull/54191


donpenney:
Rebased and adapted commit of https://github.com/openshift-kni/oran-o2ims/pull/123

bartwensley (approved):

PR #191, opened 2024-09-06
add delete secrets and list policies in manager role

edcdavid:


internal/controllers/clusterrequest_controller.go
```diff
@@ -97,6 +97,7 @@ const (
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;create;update;patch;watch
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;create;update;patch;watch
 //+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=policies,verbs=list
```
Missxiaoguo, line 100: `watch` is needed as well

PR #192, opened 2024-09-06
Update the ConfigurationApplied condition to completed only if all the policies are compliant

edcdavid:
This change force the clusterrequest object to update the ConfigurationApplied condition to completed only if all the policies are compliant.

TODO:
~System test~

### DIFF
--- a/internal/controllers/clusterrequest_controller_test.go
+++ b/internal/controllers/clusterrequest_controller_test.go
@@ -1663,4 +1663,104 @@ defaultHugepagesSize: "1G"`,
 		Expect(conditions[5].Reason).To(Equal(string(utils.CRconditionReasons.InProgress)))
 		Expect(conditions[5].Message).To(Equal("The configuration is still being applied"))
 	})
+
+	It("Updates ClusterRequest ConfigurationApplied condition to InProgress when the cluster is "+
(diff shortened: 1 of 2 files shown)


internal/controllers/clusterrequest_controller_test.go
```diff
@@ -1937,4 +1937,104 @@ defaultHugepagesSize: "1G"`,
...
+					PolicyName:        "v1-subscriptions-policy",
+					PolicyNamespace:   "ztp-clustertemplate-a-v4-16",
+					RemediationAction: "inform",
+				},
+			},
+		))
+
+		// Check the status conditions.
+		conditions := CRTask.object.Status.Conditions
+		Expect(conditions[4].Type).To(Equal(string(utils.CRconditionTypes.ConfigurationApplied)))
```
Missxiaoguo, line 2035: This [PR](https://github.com/openshift-kni/oran-o2ims/pull/190#event-14187424860) was just merged and may impact this testcase!  Could you please pull the updates and confirm this? Thanks!
```suggestion
		Expect(conditions[5].Type).To(Equal(string(utils.CRconditionTypes.ConfigurationApplied)))
```

edcdavid: Fixed.

PR #195, opened 2024-09-10
CNF-14301: Change the rendered ClusterInstance's type to siteconfig.ClusterInstance

Missxiaoguo:
The ClusterInstance template is explicitly defined and delivered within the IMS operator. When new fields are available in the Siteconfig operator, we will need to update the ClusterInstance template and upversion the IMS operator to support the new fields.

Given this, we could switch to rendering the ClusterInstance into the ClusterInstance struct type imported from the siteconfig package instead of using the unstructured type, which will simplify data manipulation. It's safe to directly manipulate with the ClusterInstance data, as the non-empty parameters have been verified in the ClusterInstance template and validated with dry-run previously.

When we need to update the ClusterInstance template to support new fields, the imported siteconfig package should be upversioned as well.

### DIFF
--- a/internal/controllers/clusterrequest_controller.go
+++ b/internal/controllers/clusterrequest_controller.go
@@ -30,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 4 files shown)


internal/controllers/clusterrequest_controller.go
```diff
@@ -436,7 +435,8 @@ func (t *clusterRequestReconcilerTask) renderClusterInstanceTemplate(
 		"Cluster": t.clusterInput.clusterInstanceData,
 	}
 
-	renderedClusterInstance, err := utils.RenderTemplateForK8sCR(
+	renderedClusterInstance := &siteconfig.ClusterInstance{}
+	renderedClusterInstanceUntyped, err := utils.RenderTemplateForK8sCR(
```
irinamihai, line 439: Could we consider renaming `renderedClusterInstanceUntyped` to `renderedClusterInstanceUnstruct`? I think it conveys its purpose a bit better.

tliu2021:
/lgtm
Good stuff! Using a struct is much cleaner.

PR #199, opened 2024-09-12
CNF-14356: Add timeout for policy configuration

irinamihai:
Add a timeout mechanism for determining policies that stay `NonCompliant`/`Pending` for longer than the configured `Spec.Timeout.Configuration`.

Changes & notes:
- only policies in `enforce` are taken into consideration when computing the timeout
- a timeout is registered in the `ConfigurationApplied` condition as follows:

>   \- lastTransitionTime: "2024-09-11T22:13:24Z"
>      message: The configuration is still being applied, but it timed out
>      reason: TimedOut
>      status: "False"
>      type: ConfigurationApplied

- the first time IMS detects an enforce policy different from Compliant, it will set `status.clusterInstanceRef.NonCompliantAt`
- if there are policies with compliance different from `Compliant` in `enforce`, IMS will requeue every minute to check for their status
- if there is a timeout registered, but then the policies get Compliant, the timeout is deleted, `status.clusterInstanceRef.NonCompliantAt` removed and the `ConfigurationApplied` condition moves to `Completed`

/cc @Missxiaoguo @bartwensley @edcdavid


internal/controllers/clusterrequest_controller.go
```diff
@@ -805,35 +832,115 @@ func (t *clusterRequestReconcilerTask) handleClusterPolicyConfiguration(ctx cont
...
 			)
 		} else {
-			utils.SetStatusCondition(&t.object.Status.Conditions,
-				utils.CRconditionTypes.ConfigurationApplied,
-				utils.CRconditionReasons.OutOfDate,
-				metav1.ConditionFalse,
-				"The configuration is out of date",
-			)
+			if nonCompliantPolicyInEnforce {
+				policyTimedOut := t.hasPolicyConfigurationTimedOut(ctx)
```
edcdavid, line 915: Why not move the clusterIsReadyForPolicyConfig call/condition here? this way we could list the policies that are matched to this cluster instance before they are getting enforced.

irinamihai: Because there are no child policies created if the cluster is not yet ready, so `targetPolicies` would always be empty.

internal/controllers/clusterrequest_controller.go
```diff
@@ -805,35 +832,115 @@ func (t *clusterRequestReconcilerTask) handleClusterPolicyConfiguration(ctx cont
...
+		t.object.Status.ClusterInstanceRef.NonCompliantAt = metav1.Now()
+		return policyTimedOut
+	}
+
+	// If the current status of the Condition is false.
+	if configurationAppliedCondition.Status == metav1.ConditionFalse {
+		switch configurationAppliedCondition.Reason {
+		case string(utils.CRconditionReasons.InProgress):
+			// Check if the configuration application has timed out.
+			if time.Since(t.object.Status.ClusterInstanceRef.NonCompliantAt.Time) > time.Duration(t.object.Spec.Timeout.Configuration) {
```
edcdavid, line 867: Multiply time.Duration(t.object.Spec.Timeout.Configuration) by unit (*time.Second?) also indicate unit in definition of t.object.Spec.Timeout.Configuration object. It is currently in nanosconds.

irinamihai: Yup, will add *time.Minute

internal/controllers/clusterrequest_controller.go
```diff
@@ -235,17 +235,16 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (ctrl.Result, er
 		}
 	}
 
-	crProvisionedCond := meta.FindStatusCondition(
-		t.object.Status.Conditions, string(utils.CRconditionTypes.ClusterProvisioned))
-	if crProvisionedCond != nil {
```
Missxiaoguo, line 240: I think we should keep this check to ensure the ordering and avoid setting the configurationApplied condition when hw provisioning is still in-progress and no ClusterInstance has been created (no ClusterProvisioned status condition)

irinamihai: We are setting the condition to status `False` with a `ClusterNotReady` reason. @edcdavid and I discussed that it's nice to know the configuration before it's actually being applied, especially since all the policies are in enforce.

Missxiaoguo: Yeah, I saw the ClusterNotReady condition you added. I think it makes sense when the cluster has started provisioning, so we have conditions like HWProvisioned, ClusterProvisioned-InProgress, and ConfigurationApplied-NotReady.

However, it feels a bit weird to me that when the ClusterInstance hasn’t been created yet because HW provisioning is still in progress, we will have the conditions HWProvisioned-InProgress and ConfigurationApplied=ClusterNotReady. This doesn’t align with how other conditions are handled, where we only add the condition when the previous step is completed.

irinamihai: Let's just keep `if crProvisionedCond != nil`  then and not check for the condition reason.

config/crd/bases/oran.openshift.io_clusterrequests.yaml
```diff
@@ -104,6 +104,11 @@ spec:
                   name:
                     description: Contains the name of the created ClusterInstance.
                     type: string
+                  nonCompliantAt:
+                    description: Holds the first timestamp when the configuration
+                      was found NonCompliant for the cluster.
+                    format: date-time
+                    type: string
```
bartwensley, line 111: It doesn't feel like this should be part of the clusterInstanceRef - should we rename that to be more general or create a new property for the policy related items?

irinamihai: I'll rename it to `ClusterDetails` since it's more general.

edcdavid (approved):

Missxiaoguo (approved):

PR #200, opened 2024-09-13
Aligning with latest from siteconfig (using extraLabels instead of clusterLabels)

edcdavid:
Align with latest from siteconfig to used extraLabels instead of clusterLabels
Also added missing permission for nodepool/status
TODO:
- ~system tests~
- ~unit tests~
- ~code cleanup~

### DIFF
--- a/config/samples/oran.openshift.io_v1alpha1_clusterrequest.yaml
+++ b/config/samples/oran.openshift.io_v1alpha1_clusterrequest.yaml
@@ -22,10 +22,6 @@ spec:
         - NTP.server1
         - 1.1.1.1
       baseDomain: example.com
-      clusterLabels:
-        common: "true"
-        group-du-sno: test
-        sites: site-sno-du-1
       clusterName: site-sno-du-1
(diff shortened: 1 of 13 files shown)


config/samples/oran.openshift.io_v1alpha1_clusterrequest.yaml
```diff
@@ -22,17 +22,17 @@ spec:
...
-        group-du-sno: test
-        sites: site-sno-du-1
       clusterName: site-sno-du-1
       extraAnnotations:
         AgentClusterInstall:
           extra-annotation-key: extra-annotation-value
       extraLabels:
         AgentClusterInstall:
           extra-label-key: extra-label-value
+        ManagedCluster:
```
irinamihai, line 32: I would update these to labels that we are actually using, to have an example that's closer to reality:
```
    cluster-version: v4.16
    clustertemplate-a-policy: v1
```

Missxiaoguo (approved):

PR #201, opened 2024-09-13
CNF-14426: Add Interfaces Handling to Hardware Provisioning

tliu2021:
This update enhances the hardware provisioning process by adding support for handling network interfaces. The hardware manager will ink a label to each interface on the server as part of the hardware onboarding process, which is managed outside the scope of ORAN use cases. Template authors will need to be aware of these label mappings to associate them with the appropriate ports on the server.

Key updates include:
- Passing the node network interfaces to the nodePool for each group.
- Retrieving each interface’s MAC address from the node object’s status field after the nodePool has been provisioned.
- Adding the boot interface label to the hardware template, which is used to select the boot interface and populate the bootMACAddress field in the cluster instance configuration.

In addition, the test cases have been updated to validate the new interface handling and ensure that the boot interface selection is properly reflected in the provisioning process.


internal/controllers/utils/utils.go
```diff
@@ -844,3 +846,83 @@ func GetDefaultBackendTransport() (http.RoundTripper, error) {
 
 	return net.SetTransportDefaults(&http.Transport{TLSClientConfig: tlsConfig}), nil
 }
+
+// Helper function to compare two slices of interfaces
+func IsInterfacesEqual(a, b []string) bool {
```
edcdavid, line 851: we could also use https://pkg.go.dev/slices#Equal

tliu2021: Changed.

api/hardwaremanagement/v1alpha1/node.go
```diff
@@ -39,7 +46,7 @@ type BMC struct {
 type NodeStatus struct {
 	BMC *BMC `json:"bmc,omitempty"`
 
-	BootMACAddress string `json:"bootMACAddress,omitempty"`
+	Interfaces []*Interface `json:"interfaces,omitempty"`
```
donpenney, line 49: Going forward, we'll need to be careful about changes like this, coordinating changes with the hwmgr plugins (where we currently only have the Test Plugin). For example, we may need to keep stale fields around for a period to avoid breaking the plugin(s). Since the changes to enable the Test Plugin integration are still under review, I'd be the only one impacted, but this would definitely be a breaking change for the Test Plugin, ie. a Non-Upward Compatible change (NUC)

tliu2021: Agree.

internal/controllers/utils/constants.go
```diff
@@ -118,8 +118,9 @@ const (
 
 // Hardeware template constants
 const (
-	HwTemplatePluginMgr = "hwMgrId"
-	HwTemplateNodePool  = "node-pools-data"
+	HwTemplatePluginMgr      = "hwMgrId"
+	HwTemplateNodePool       = "node-pools-data"
+	HwTemplateBootIfaceLabel = "bootIntefaceLabel"
```
donpenney, line 123: typo: bootIntefaceLabel

tliu2021: Fixed.

donpenney (approved):

PR #202, opened 2024-09-13
CNF-14303: Support integration with hardware manager plugin

donpenney:
This update includes changes to support using the Test Plugin.

### DIFF
--- a/internal/controllers/clusterrequest_controller.go
+++ b/internal/controllers/clusterrequest_controller.go
@@ -1201,18 +1201,6 @@ func (t *clusterRequestReconcilerTask) hwMgrPluginNamespaceExists(
 }
 
 func (t *clusterRequestReconcilerTask) createNodePoolResources(ctx context.Context, nodePool *hwv1alpha1.NodePool) error {
-	chkNodepool := &hwv1alpha1.NodePool{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      nodePool.Name,
-			Namespace: nodePool.Namespace,
-		},
-	}
-	if exists, err := utils.DoesK8SResourceExist(ctx, t.client, nodePool.Name, nodePool.Namespace, chkNodepool); err != nil {
-		return fmt.Errorf("failed check is nodepool exists %s: %w", nodePool.Name, err)
-	} else if exists {
-		return nil
-	}
-
 	// Create the hardware plugin namespace.
 	pluginNameSpace := nodePool.ObjectMeta.Namespace
 	if exists, err := t.hwMgrPluginNamespaceExists(ctx, pluginNameSpace); err != nil {
@@ -1228,19 +1216,19 @@ func (t *clusterRequestReconcilerTask) createNodePoolResources(ctx context.Conte
 		return fmt.Errorf("specified hardware manager plugin namespace does not exist: %s", pluginNameSpace)
 	}
 
-	// Create the clusterInstance namespace.
+	// Create/update the clusterInstance namespace, adding ClusterRequest labels to the namespace
 	err := t.createClusterInstanceNamespace(ctx, nodePool.GetName())
 	if err != nil {
 		return err
 	}
 
-	// Create/update the node pool resource
-	createErr := utils.CreateK8sCR(ctx, t.client, nodePool, t.object, utils.UPDATE)
+	// Create the node pool resource
+	createErr := utils.CreateK8sCR(ctx, t.client, nodePool, t.object, "")
 	if createErr != nil {
 		t.logger.ErrorContext(
 			ctx,
 			fmt.Sprintf(
-				"Failed to create/update the NodePool %s in the namespace %s",
+				"Failed to create the NodePool %s in the namespace %s",
 				nodePool.GetName(),
 				nodePool.GetNamespace(),
 			),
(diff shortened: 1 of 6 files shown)


internal/controllers/clusterrequest_controller.go
```diff
@@ -214,8 +217,10 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (ctrl.Result, er
 				renderedNodePool.GetNamespace(),
 			),
 		)
-		// TODO: enable Requeue after the hardware plugin is ready
-		// return ctrl.Result{RequeueAfter: time.Second * 30}, nil
+		// TODO: Remove this check once hwmgr plugin(s) are fully utilized
+		if renderedNodePool.ObjectMeta.Namespace != utils.TempDellPluginNamespace && renderedNodePool.ObjectMeta.Namespace != utils.UnitTestHwmgrNamespace {
```
tliu2021, line 221: The nodePool-related unit tests should pass without this check. The failed cases are in the policyManagement tests, which are missing the nodePool provisioned status. We plan to add this status in those test cases when requeue is enabled.

Since all test cases use "hwmgr" as the hwMgrId, the policyManagement test cases are passed with this check.

internal/controllers/clusterrequest_controller.go
```diff
@@ -1172,23 +1179,63 @@ func (t *clusterRequestReconcilerTask) createHwMgrPluginNamespace(
...
+
 	// Create the clusterInstance namespace.
 	err := t.createClusterInstanceNamespace(ctx, nodePool.GetName())
 	if err != nil {
 		return err
 	}
 
 	// Create/update the node pool resource
-	createErr = utils.CreateK8sCR(ctx, t.client, nodePool, t.object, utils.UPDATE)
+	createErr := utils.CreateK8sCR(ctx, t.client, nodePool, t.object, utils.UPDATE)
```
Missxiaoguo, line 1238: I realized that if you pass an empty string to the "operation". It will not update if already exists
```suggestion
	createErr := utils.CreateK8sCR(ctx, t.client, nodePool, t.object, "")
```

internal/controllers/clusterrequest_controller.go
```diff
@@ -1172,23 +1179,63 @@ func (t *clusterRequestReconcilerTask) createHwMgrPluginNamespace(
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nodePool.Name,
+			Namespace: nodePool.Namespace,
+		},
+	}
+	if exists, err := utils.DoesK8SResourceExist(ctx, t.client, nodePool.Name, nodePool.Namespace, chkNodepool); err != nil {
+		return fmt.Errorf("failed check is nodepool exists %s: %w", nodePool.Name, err)
+	} else if exists {
+		return nil
+	}
```
tliu2021, line 1214: We can replace "utils.UPDATE" with an empty string "" in all utils.CreateK8sCR() calls for creating nodePools and namespaces.  If we make this replacement, the check to see if the CR exists will not be needed, as CreateK8sCR() will no longer update the CR if it already exists.

internal/controllers/clusterrequest_controller.go
```diff
@@ -214,8 +217,10 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (ctrl.Result, er
 				renderedNodePool.GetNamespace(),
 			),
 		)
-		// TODO: enable Requeue after the hardware plugin is ready
-		// return ctrl.Result{RequeueAfter: time.Second * 30}, nil
+		// TODO: Remove this check once hwmgr plugin(s) are fully utilized
+		if renderedNodePool.ObjectMeta.Namespace != utils.TempDellPluginNamespace && renderedNodePool.ObjectMeta.Namespace != utils.UnitTestHwmgrNamespace {
+			return ctrl.Result{RequeueAfter: time.Second * 30}, nil
```
tliu2021, line 222: Do we want to enable the logic to wait for the nodePool to be provisioned in this update? The code related to createClusterInstanceBMCSecrets() will be removed when we do.

internal/controllers/clusterrequest_controller.go
tliu2021: @Missxiaoguo thanks for sharing this info with me today!

internal/controllers/clusterrequest_controller.go
donpenney: Are you asking me to make some change in test cases in order to remove this check? I added the check in order to both avoid breaking the unit tests and to avoid changes to anyone's current workflow who may be using the "dell-hwmgr" tag, until such time as folks transition to using the test plugin.

internal/controllers/clusterrequest_controller.go
donpenney: My goal was to enable support for using the test plugin while trying to avoid impacting anyone's current workflow

internal/controllers/clusterrequest_controller.go
donpenney: Done.

internal/controllers/clusterrequest_controller.go
donpenney: Done.

internal/controllers/clusterrequest_controller.go
donpenney: Sounds good

PR #203, opened 2024-09-16
MGMT-18924: Add RBAC role to Resource Server

alegacy:
This adds a clusterrole to the Resource Server so that it has permission to access data via the ACM search-api.


internal/controllers/orano2ims_controller.go
```diff
@@ -471,7 +493,98 @@ func (t *reconcilerTask) createDeploymentManagerClusterRoleBinding(ctx context.C
...
+
+func (t *reconcilerTask) createResourceServerClusterRole(ctx context.Context) error {
+	role := &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf(
+				"%s-%s", t.object.Namespace, utils.ORANO2IMSResourceServerName,
+			),
+		},
+		Rules: []rbacv1.PolicyRule{
+			// We need to read manged clusters, as that is the main source for the
```
irinamihai, line 510: Typo" `manged`

PR #207, opened 2024-09-18
CNF-13703: Immutable fields check for ClusterInstance

Missxiaoguo:
Updates:
- Use r3labs/diff package to compare the old and new ClusterInstance as it provides detailed information about the specific fields that have been changed
- After Cluster starts provisioning, or reaches a final status(Completed or Failed), updates to immutable fields in the ClusterInstance are disallowed. Only `extraAnnotations` and `extraLabels` can be modified. Any attempt to change immutable fields will result in the ClusterInstanceRendered status condition being marked as failed with an error message such as: `Failed to render and validate ClusterInstance: detected changes in immutable fields: baseDomain, nodes.0.hostName`.
- Unittests


internal/controllers/utils/utils.go
```diff
@@ -771,6 +772,68 @@ func CheckClusterLabelsForPolicies(
...
 }
 
+// FindClusterInstanceImmutableFieldUpdates identifies updates made to immutable fields
+// in the ClusterInstance spec. It returns a list of paths for the updated fields that
+// are considered immutable and should not be modified.
+func FindClusterInstanceImmutableFieldUpdates(
+	oldClusterInstance, newClusterInstance *unstructured.Unstructured) ([]string, error) {
+
+	// Non-immutable ClusterInstance fields at the cluster-level
+	allowedClusterFields := []string{
```
alegacy, line 782: I would suggest pulling these 2 lists out into `constants.go` so that they were separate from the code.

internal/controllers/utils/utils.go
```diff
@@ -771,6 +772,68 @@ func CheckClusterLabelsForPolicies(
...
+
+		oranUtilsLog.Info(
+			fmt.Sprintf(
+				"Detected field change in the newly rendered ClusterInstance(%s) type: %s, path: %s, from: %+v, to: %+v",
+				oldClusterInstance.GetName(), diff.Type, strings.Join(diff.Path, "."), diff.From, diff.To,
+			),
+		)
+
+		if diff.Path[0] == "nodes" {
+			if len(diff.Path) == 2 || (len(diff.Path) > 2 && !slices.Contains(allowedNodeFields, diff.Path[2])) {
```
alegacy, line 826: Can you double-check the logic here?  It seems odd that if the `len(diff.Path) == 2` that we are assuming that the field should be added to `updatedFields` regardless of its name.

internal/controllers/utils/utils.go
Missxiaoguo: I didn’t put it in constants.go initially because I thought only `const` types should be added there, but it looks like we already have `var` in that file as well. I’ll update it.

internal/controllers/utils/utils.go
Missxiaoguo: Yes, in the case of len(diff.Path) == 2, a new node is being added to the ClusterInstance, and it must contain immutable fields as they are mandatory in the template.

However, I do want to update something else—we should allow new nodes to be added after successful cluster provisioning (even though we’re not officially considering scaling in/out at the moment).

alegacy: ok

internal/controllers/utils/constants.go
```diff
@@ -87,11 +87,23 @@ const (
 	ClusterInstanceTemplateName                 = "ClusterInstance"
 	ClusterInstanceTemplatePath                 = "controllers/clusterinstance-template.yaml"
 	ClusterInstanceTemplateConfigmapName        = "sc-clusterinstance-template"
-	ClusterInstanceTemplateConfigmapNamespace   = InventoryNamespace
```
Missxiaoguo, line 90: Removed as it's not used

internal/controllers/utils/utils_test.go
```diff
@@ -859,121 +859,127 @@ nodes:
...
-										"type":  "bond",
-									},
+								"dhcp":    false,
+								"enabled": true,
+							},
+							"ipv6": map[string]interface{}{
+								"address": []map[string]interface{}{
+									{"ip": "2620:52:0:10e7:e42:a1ff:fe8a:601", "prefix-length": 64},
+									{"ip": "2620:52:0:10e7:e42:a1ff:fe8a:602", "prefix-length": 64},
+									{"ip": "2620:52:0:10e7:e42:a1ff:fe8a:603", "prefix-length": 64},
```
bartwensley, line 920: We shouldn't be using real IP addresses in unit tests. Would prefer to used the addresses reserved for documentation see https://en.wikipedia.org/wiki/Reserved_IP_addresses

PR #208, opened 2024-09-18
Add placeholders for values that come from NodePool

Missxiaoguo:
bmcAddress, bmcCredentialsName, bootMACAddress, and nodeNetwork.interfaces[*].macAddress are meant to come from the NodePool (hardware) and should not be exposed in the ClusterRequest when hw plugin is fully utilized.

Placeholders for these fields have been added to pass the ClusterInstance dry-run validation since they are required fields. These placeholders will be replaced with data retrieved from NodePool later on.

Before hw plugin is fully utilized, these fields can still be provided through the ClusterRequest.

Remove the creation of ClusterInstance template configmap since the configmap is useless.

### DIFF
--- a/bundle/manifests/oran-o2ims.clusterserviceversion.yaml
+++ b/bundle/manifests/oran-o2ims.clusterserviceversion.yaml
@@ -374,9 +374,6 @@ metadata:
                                     "type": "string"
                                   }
                                 },
-                                "required": [
-                                  "macAddress"
-                                ],
                                 "type": "object"
                               },
                               "minItems": 1,
--- a/config/samples/v1alpha1_clustertemplate.yaml
+++ b/config/samples/v1alpha1_clustertemplate.yaml
@@ -186,8 +186,6 @@ spec:
                           description: 'nic name used in the yaml, which relates
                             1:1 to the mac address. Name in REST API: logicalNICName'
                           type: string
-                      required:
-                      - macAddress
                       type: object
                     minItems: 1
                     type: array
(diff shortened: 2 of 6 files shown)


bartwensley:
> Placeholders for these fields have been added to pass the ClusterInstance dry-run validation since they are required fields. These placeholders will be replaced with data retrieved from NodePool later on.

Are these placeholders permanent? i.e. will we always need them in order to pass dry-run validation? If so, this should be commented well in the code.

Missxiaoguo:
> > Placeholders for these fields have been added to pass the ClusterInstance dry-run validation since they are required fields. These placeholders will be replaced with data retrieved from NodePool later on.
> 
> Are these placeholders permanent? i.e. will we always need them in order to pass dry-run validation? If so, this should be commented well in the code.

Yeah, they are permanent. I was considering to add a comment in this yaml file but it will also ne included in the created configmap which is visible to everyone, but I think that might be fine?

bartwensley:
> Yeah, they are permanent. I was considering to add a comment in this yaml file but it will also ne included in the created configmap which is visible to everyone, but I think that might be fine?

If we aren't using the created configmap for anything, I think it is best to just remove it. And then it would be good to add comments to the yaml file explaining why these defaults are required.

Missxiaoguo:
> If we aren't using the created configmap for anything, I think it is best to just remove it. And then it would be good to add comments to the yaml file explaining why these defaults are required.

Sounds good. We are not actually using the configmap. It's created just for information which seems unnecessary. I will remove it and add comments in this yaml file.

bartwensley (approved):
lgtm

PR #210, opened 2024-09-19
Proposal: Using Finite State Machine library to expose state diagram for configuration applied transitions

edcdavid:
Proposing to use https://github.com/qmuntal/stateless to expose the implicit state diagram used for The configuration applied status
- With each calls to reconcile, the state machine is initialized with the current state saved in configurationAppliedCondition.Reason
- The FsmHelper collects all the data required by the state machine to make its transition decisions. It is initialized  during reconciliation with .InitFromCluster()
- updateConfigurationAppliedStatus pushes updates to the status back to the cluster

The fsm_test.go  tests the state machine transitions independently from kubenetes.
The ci-job process updates the state diagram in documentation (dev/config_fsm.mg [link](https://github.com/openshift-kni/oran-o2ims/blob/25341212e7b943a95711adb8983e5800bbc052a7/docs/dev/config_fsm.md))

TODO:
- ~~Unit tests~~
- ~~add CGO in prow for creating SVG in docs~~
- ~~system test - retesting~~


internal/configfsm/fsm.go
```diff
@@ -0,0 +1,315 @@
...
+func (h *BaseFSMHelper) IsPoliciesMatched() bool {
+	return h.PoliciesMatched
+}
+
+// Resets the NonCompliantAt field to zero
+func (h *BaseFSMHelper) ResetNonCompliantAt() {
+	h.NonCompliantAt = time.Time{}
+}
+
+// Returns true if NonCompliantAt is was resetm to zero, false otherwise
```
alegacy, line 291: Typo/Grammar:  "is was", "resetm"

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,315 @@
...
+	// Transitions to InProgress state
+	ClusterNotReadyToInProgress Trigger = "ClusterNotReady->InProgress"
+	OutOfDateToInProgress       Trigger = "OutOfDate->InProgress"
+	CompletedToInProgress       Trigger = "Completed->InProgress"
+
+	// Transitions to TimedOut state
+	ClusterNotReadyToTimedOut Trigger = "ClusterNotReady->TimedOut"
+	InProgressToTimedOut      Trigger = "InProgress->TimedOut"
+
+	// Transitions to Completed state
```
alegacy, line 38: These transitions to `Completed` seem suspect to me.  Is it really possible to go directly from a `TimedOut` or `OutOfDate` state to `Completed` without doing any other work?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,315 @@
...
+			fmt.Println("Entering InProgress")
+			if fsmHelper.IsResetNonCompliantAt() {
+				fsmHelper.SetResetNonCompliantAtNow()
+			}
+
+			if fsmHelper.IsTimedOut() {
+				return fsm.Fire(InProgressToTimedOut, fsmHelper)
+			}
+			if fsmHelper.IsAllPoliciesCompliant() {
+				return fsm.Fire(InProgressToCompleted, fsmHelper)
```
alegacy, line 155: If `Completed` is the final state then its pre-conditions should be checked last so that if any of the other conditions are true they take precedence.

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,315 @@
...
+func (h *BaseFSMHelper) IsClusterReady() bool {
+	return h.ClusterReady
+}
+
+// Returns true if all policies are enforced and compliant, false otherwise
+func (h *BaseFSMHelper) IsAllPoliciesCompliant() bool {
+	return h.AllPoliciesCompliant
+}
+
+// Returns true if there is at least one non compliant policy in enforce, false otherwise
```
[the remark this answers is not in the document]
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed, removed these transitions.
The following test had a tranisition from Timedout to Completed by mistake I believe, because of a missing timeout set to zero, so were essentially testing this condition:
- "Verify status conditions when configuration change causes ClusterInstance rendering to fail but configuration policy becomes compliant"
- "Verify status conditions when ClusterInstance provision has completed, ManagedCluster becomes ready and configuration policy becomes compliant"
This test tested it directly : Moves from TimedOut to Completed if all the policies are compliant. But that might be because of the previous 2?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,315 @@
...
+			if !fsmHelper.IsPoliciesMatched() {
+				return fsm.Fire(InProgressToMissing, fsmHelper)
+			}
+			if !fsmHelper.IsClusterReady() {
+				return fsm.Fire(InProgressToClusterNotReady, fsmHelper)
+			}
+			if !fsmHelper.IsNonCompliantPolicyInEnforce() && !fsmHelper.IsAllPoliciesCompliant() {
+				return fsm.Fire(InProgressToOutOfDate, fsmHelper)
+			}
+			return nil
```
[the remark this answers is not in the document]
edcdavid: Nil in this context means no error, so there would be nothing to do. If any of the Fire function return an error, The RunFSM function returns an error. This could happen if the state that is fired is producing a non permitted transition.

internal/configfsm/fsm.go
edcdavid: Fixed.

internal/configfsm/fsm.go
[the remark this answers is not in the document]
alegacy: right so the FSM will remain in the current state until it fires again and will repeat that until one of the preconditions matches to move it to another state?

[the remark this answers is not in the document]
edcdavid: yes, correct. The RunFSM runs the state transitions to completion from the current state until it is called again. When it reaches "return nil", the RunFSM exits.

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// Self transitions
+	MissingToMissing                 Trigger = "Missing->Missing"
+	ClusterNotReadyToClusterNotReady Trigger = "ClusterNotReady->ClusterNotReady"
+	InProgressToInProgress           Trigger = "InProgress->InProgress"
+	TimedOutToTimedOut               Trigger = "TimedOut->TimedOut"
+	CompletedToCompleted             Trigger = "Completed->Completed"
+	OutOfDateToOutOfDate             Trigger = "OutOfDate->OutOfDate"
+
+	// States
```
irinamihai, line 54: It would be nice here to use the Conditions from the [utils package](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/utils/conditions.go#L52-L72).

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	InProgressToMissing      Trigger = "InProgress->Missing"
+	CompletedToMissing       Trigger = "Completed->Missing"
+	TimedOutToMissing        Trigger = "TimedOut->Missing"
+	OutOfDateToMissing       Trigger = "OutOfDate->Missing"
+
+	// Transitions to ClusterNotReady state
+	MissingToClusterNotReady    Trigger = "Missing->ClusterNotReady"
+	InProgressToClusterNotReady Trigger = "InProgress->ClusterNotReady"
+	CompletedToClusterNotReady  Trigger = "Completed->ClusterNotReady"
+	OutOfDateToClusterNotReady  Trigger = "OutOfDate->ClusterNotReady"
```
irinamihai, line 26: Could we add a small comment explaining why `TimedOut -> ClusterNotReady `is not a viable transition, please?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// Transitions to InProgress state
+	ClusterNotReadyToInProgress Trigger = "ClusterNotReady->InProgress"
+	OutOfDateToInProgress       Trigger = "OutOfDate->InProgress"
+	CompletedToInProgress       Trigger = "Completed->InProgress"
```
irinamihai, line 31: We could have `Missing -> InProgress` if there are no policies at all during the initial config, but we add one as Day2 config, directly in `enforce`.

internal/configfsm/fsm.go
irinamihai: We can actually go from `TimedOut` to `Completed` if the policies become `Compliant` eventually and no other action is being taken as a consequence of the time out (which is the case right now).
It's also possible to go from `OutOfDate` to `Completed` if the `inform`, `NonCompliant` policy/policies that's causing the `OutOfDate` gets deleted and all the remaining policies are `Compliant`.

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	InProgressToTimedOut      Trigger = "InProgress->TimedOut"
+
+	// Transitions to Completed state
+	InProgressToCompleted Trigger = "InProgress->Completed"
+
+	// Transitions to OutOfDate state
+	InProgressToOutOfDate      Trigger = "InProgress->OutOfDate"
+	TimedOutToOutOfDate        Trigger = "TimedOut->OutOfDate"
+	ClusterNotReadyToOutOfDate Trigger = "ClusterNotReady->OutOfDate"
+	CompletedToOutOfDate       Trigger = "Completed->OutOfDate"
```
irinamihai, line 44: `Missing` -> `OutOfData` is also a possibility if we have no policies, but then we add policies in `inform`.

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+		Permit(TimedOutToOutOfDate, OutOfDate).
+		Permit(TimedOutToMissing, Missing).
+		PermitReentry(TimedOutToTimedOut)
+
+	return fsm, nil
+}
+
+// FsmHelper Helper Interface to collect all functions needed for state machine decisions
+type FsmHelper interface {
+	IsPoliciesMatched() bool
```
irinamihai, line 254: Rename to `ArePoliciesMatched`?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	ResetNonCompliantAt()
+	IsResetNonCompliantAt() bool
```
irinamihai, line 256: Rename to `IsNonCompliantAtZero`?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	SetResetNonCompliantAtNow()
```
irinamihai, line 257: Maybe just `SetNonCompliantAtNow`?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	IsTimedOut() bool
+	IsClusterReady() bool
+	IsAllPoliciesCompliant() bool
```
irinamihai, line 260: `AreAllPoliciesCompliant`?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	if err != nil {
+		return aState, fmt.Errorf("could not get state, err: %w", err)
+	}
+	return aState, nil
+}
+
+// InitFSM Initialize the state machine algorithm
+func InitFSM(state string) (fsm *stateless.StateMachine, err error) {
+	fsm = stateless.NewStateMachine(state)
+	fsm.Configure(Start).Permit(StartToMissing, Missing)
```
Missxiaoguo, line 118: Does this mean that `missing` is checked first? If there are no policies and Cluster is not ready, will the status condition set to Missing instead of ClusterNotReady?

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,309 @@
...
+	fsm.Configure(ClusterNotReady).
+		OnEntry(func(_ context.Context, args ...any) error {
+			fsmHelper := args[0].(FsmHelper)
+			fmt.Println("Entering ClusterNotReady")
+			if fsmHelper.IsTimedOut() {
+				return fsm.Fire(ClusterNotReadyToTimedOut, fsmHelper)
+			}
+			if fsmHelper.IsClusterReady() {
+				if fsmHelper.IsNonCompliantPolicyInEnforce() || fsmHelper.IsAllPoliciesCompliant() {
+					return fsm.Fire(ClusterNotReadyToInProgress, fsmHelper)
```
irinamihai: Normally, if all the policies are compliant, it should go to `Completed`.
Is the idea here that it will go to `InProgress` first and then to `Completed`?

internal/configfsm/fsm.go
edcdavid: Fixed with the latest updates. Redundant direct states removed. The machine each's state make their own decisions and transition between states until a final state has been reached.

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed with the latest updates. Redundant direct states removed. The machine each's state make their own decisions and transition between states until a final state has been reached.

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
edcdavid: Fixed

internal/configfsm/fsm.go
Missxiaoguo: Thanks for explanation!

internal/controllers/clusterrequest_controller.go
```diff
@@ -876,200 +878,88 @@ func (t *clusterRequestReconcilerTask) handleClusterPolicyConfiguration(ctx cont
...
-					utils.CRconditionTypes.ConfigurationApplied,
-				),
-			)
-		}
-	} else if configurationAppliedCondition.Reason == string(utils.CRconditionReasons.Completed) {
-		t.object.Status.ClusterDetails.NonCompliantAt = metav1.Now()
-	}
-
-	return policyTimedOut
+	return aConfigFSMHelper.NonCompliantPolicyInEnforce, nil
```
Missxiaoguo, line 910: Currently, as long as there are non-compliant enforce policies, we requeue. Do we still need to periodically check after timeout? Since if temporary ACM issue gets resolved and policies's compliant status gets changed, reconciliation will get triggered anyway.

edcdavid: Fixed.

internal/configfsm/fsm.go
edcdavid: Yes, correct.

internal/controllers/utils/utils.go
```diff
@@ -1044,21 +1042,7 @@ func ClusterIsReadyForPolicyConfig(
...
-
-	joinedCondition := meta.FindStatusCondition(
-		managedCluster.Status.Conditions,
-		clusterv1.ManagedClusterConditionJoined)
-	if joinedCondition != nil && joinedCondition.Status == metav1.ConditionTrue {
-		joined = true
-	}
-
-	return available && hubAccepted && joined, nil
+	return available, nil
```
irinamihai, line 1045: Why shouldn't we also check for joined and accepted? :thinking:

edcdavid: I changed it to pass 2 unit test cases, since they only set the available condition:
- "Verify status conditions when ClusterInstance provision has completed, ManagedCluster becomes ready and configuration policy becomes compliant"
- "Verify status conditions when configuration change causes ClusterInstance rendering to fail but configuration policy becomes compliant"

see: https://github.com/openshift-kni/oran-o2ims/blob/a617e9c002be312d3c353b6db861106acf07c6b5/internal/controllers/clusterrequest_controller_test.go#L1026

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,277 @@
...
+		}).
+		Permit(InProgressToClusterNotReady, ClusterNotReady).
+		Permit(InProgressToTimedOut, TimedOut).
+		Permit(InProgressToOutOfDate, OutOfDate).
+		Permit(InProgressToCompleted, Completed).
+		PermitReentry(InProgressToInProgress)
+	fsm.Configure(OutOfDate).
+		OnEntry(func(_ context.Context, args ...any) error {
+			fsmHelper := args[0].(FsmHelper)
+			fmt.Println("Entering InProgress")
```
irinamihai, line 174: Typo: `OutOfDate` instead of `InProgress`

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,277 @@
...
+			fsmHelper.ResetNonCompliantAt()
+			if !fsmHelper.ArePoliciesMatched() ||
```
irinamihai, line 176: If we're in `outOfDate` because of an `Inform` `NonCompliant` policy, but then we delete it, it should go to `Completed`, through `InProgress`, but neither one of those conditions would facilitate the transition. Kindly confirm.

internal/controllers/utils/utils.go
edcdavid: Fixed. I re-added the joined and accepted condition check and modified the existing unit test

internal/configfsm/fsm.go
edcdavid: Fixed. Also added a unit test for it

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,278 @@
...
+	fsm.Configure(InProgress).
+		OnEntry(func(_ context.Context, args ...any) error {
+			fsmHelper := args[0].(FsmHelper)
+			fmt.Println("Entering " + InProgress)
+			if fsmHelper.IsNonCompliantAtZero() ||
+				fsmHelper.IsAllPoliciesCompliant() ||
+				(!fsmHelper.IsNonCompliantPolicyInEnforce() && !fsmHelper.IsAllPoliciesCompliant()) {
+				fsmHelper.SetNonCompliantAtNow()
+			}
+			if !fsmHelper.ArePoliciesMatched() {
```
irinamihai, line 149: It looks like we can group this one with `|| !fsmHelper.IsClusterReady()` since they both transition to `ClusterNotReady`

edcdavid: we can't group these 2 because the "fsmHelper.IsTimedOut()" must be evaluated first. Basically if we are in timeout we should stay in TimedOut even if the cluster is not ready

internal/controllers/utils/conditions.go
```diff
@@ -49,27 +49,17 @@ var CTconditionReasons = struct {
...
-	ClusterNotReady: "ClusterNotReady",
-	Completed:       "Completed",
-	Failed:          "Failed",
-	InProgress:      "InProgress",
-	Missing:         "Missing",
-	OutOfDate:       "OutOfDate",
-	TimedOut:        "TimedOut",
-	Unknown:         "Unknown",
-}
+const (
```
Missxiaoguo, line 52: May I ask the reason for this change?

internal/controllers/clusterrequest_controller.go
Missxiaoguo: Thanks!

internal/controllers/utils/conditions.go
edcdavid: It is to be able to re-use the constants names as states instead of coming up with new ones.

internal/configfsm/fsm.go
```diff
@@ -0,0 +1,278 @@
...
+		}).
+		Permit(OutOfDateToInProgress, InProgress).
+		PermitReentry(OutOfDateToOutOfDate)
+	fsm.Configure(Completed).
+		OnEntry(func(_ context.Context, args ...any) error {
+			fsmHelper := args[0].(FsmHelper)
+			fmt.Println("Entering " + Completed)
+			fsmHelper.ResetNonCompliantAt()
+
+			if !fsmHelper.ArePoliciesMatched() ||
```
irinamihai, line 192: It looks here like we cannot get to OutOfDate from Completed :thinking:

internal/controllers/utils/conditions.go
Missxiaoguo: Thanks for explanation! But I think we can still reuse the constants (i.e, string(utils.CRconditionReasons.Missing), without the needing to change it?

edcdavid: The problem I had is that these are not constants, they are  variables so go would not let me do the assignment.

internal/configfsm/fsm.go
edcdavid: Fixed! also added a unit test for it

internal/controllers/utils/conditions.go
Missxiaoguo: What about keeping the struct while referencing to the constant vars you have added? so we can still benefit from the grouped reasons, and we don't have to change the existing references.
var CRconditionReasons = struct {
	NotApplied      ConditionReason
}{
	NotApplied:      NotApplied,
}

edcdavid: Fixed

docs/dev/config_fsm.md
```diff
@@ -0,0 +1,39 @@
...
+```
+digraph {
```
pixelsoccupied, line 5: Would it be possible generate this using `make generate` by `sm.ToGraph()` as mentioned [here](https://github.com/qmuntal/stateless?tab=readme-ov-file#export-to-dot-graph)

Also since this generated...would it be nicer to move this under say `docs/generated/clusterrequest-fsm` to reflect the content better?

pixelsoccupied (commented):
Although it's a nice use of fsm, i dont know if the concept of finite statement machine is compatible with the notion of k8s controller? Meaning operators are meant to be "level based", i.e look at the state of the cluster at every requeue until we reach desired state vs "edge-based" where we are relying to history stored in status. 

Api convention doc specifically says `.status` should not be used for state machines.

lots of great doc on this: 
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
- https://book.kubebuilder.io

Since this is totally new pattern we are introducing (I couldnt find any reference of fsm lib in operators anywhere), at the very least this should be better documented for anyone coming to develop this in the future?

edcdavid:
Closing this PR for now, does not follow the no state/machine recommendation for operator. Should recompute the state from information available in the cluster. This follow up PR is attempting to do this: https://github.com/openshift-kni/oran-o2ims/pull/246

PR #211, opened 2024-09-19
MGMT-18976: Use kube-rbac-proxy for authentication and authorization

alegacy:
This replaces the current authentication and authorization code by leveraging the kube-rbac-proxy.  This reduces the amount of code we must support and allows us to support the standard Kubernetes RBAC mechanism for authorizing users to use the O2IMS API endpoints.


alegacy:
/hold

...Ok to review, but want to do another pass at testing before I merge

internal/controllers/utils/constants.go
```diff
@@ -153,3 +139,12 @@ const (
...
 )
+
+// POD Container Names
+const (
+	ServerContainerName = "server"
+	RbacContainerName   = "rbac"
+)
+
+// KubeRbacProxyDefaultImage Default image for the kube-rbac-proxy
+const KubeRbacProxyDefaultImage = "gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0"
```
alegacy, line 150: @rauhersu  here is where that image is specified.

rauhersu: Understood, thanks ! We need the same old trick, this needs to be overridden from an env variable defined on a yaml file. I'd need to understand the architecture of your pod better, as I see this kuberbac proxy is executed as an extra container as part of a new pod here: https://github.com/openshift-kni/oran-o2ims/pull/211/files#diff-3cb677c868fecacf955312b938dfa9b44229aee58113a1a07f12fb5a5147631bR777

In any case, if the env we have defined on the CSV is accessible here (I guess you are exec'ing this second pod from the main operator container image) you might just want to grab the environment from the CSV and additionally, should you want to override it as part of the inventory CR - I see this is optional - , define the casuistic here: https://github.com/openshift-kni/oran-o2ims/pull/211/files#diff-3cb677c868fecacf955312b938dfa9b44229aee58113a1a07f12fb5a5147631bR741 which in terms of priorities translates into:

1st image from the CSV (env)
2nd image from the inventory CR
3rd image from the hardcoded golang code.

Note also, that if you want to perform that override choosing your own kuberbac image (2nd), you are entering 'developer mode', bypassing the official customer config.

Cc @donpenney

rauhersu: /hold

Needs upstream-midstream agreeement

alegacy:
@rauhersu @donpenney I've updated my PR with a new commit to manage the kube-rbac-proxy image differently.  Would you mind taking another look?

config/default/kustomization.yaml
```diff
@@ -140,3 +140,17 @@ patches:
...
+    -   source:
+            fieldPath: spec.template.spec.containers.[name=kube-rbac-proxy].image
+            kind: Deployment
+            name: controller-manager
+        targets:
+            -   fieldPaths:
+                    - spec.template.spec.containers.[name=manager].env.[name=^KUBE_RBAC_PROXY_IMAGE$].value
+                select:
+                    kind: Deployment
+                    name: controller-manager
```
donpenney, line 156: nit: Whitespace is inconsistent with the rest of the yaml

donpenney (commented):
/lgtm

config/default/kustomization.yaml
alegacy: Fixed.

rauhersu:
Changes lgtm, thx !

/unhold

donpenney (approved):

PR #213, opened 2024-09-20
CNF-14425: Add timeout handling for hardware provisioning

tliu2021:
This update introduces timeout handling for node pool provisioning status. It tracks the start time of the hardware provisioning process, allowing the system to detect if it exceeds the configured timeout. If a timeout occurs,
it is reflected in the HardwareProvisioned condition of the cluster request.

Additionally, this update improves error handling by implementing requeue on errors and stopping requeue when provisioning has failed. A sample of hardware templates has also been included for reference.


api/v1alpha1/clusterrequest_types.go
```diff
@@ -66,6 +66,8 @@ type NodePoolRef struct {
 	Name string `json:"name,omitempty"`
 	// Contains the namespace of the created NodePool.
 	Namespace string `json:"namespace,omitempty"`
+	// Represents the timestamp of the first status check for hardware provisioning
+	StatusCheckStart metav1.Time `json:"statusCheckStart,omitempty"`
```
edcdavid, line 70: StatusCheckStart seems generic, what about HardwareProvisioningStart?

internal/controllers/clusterrequest_controller.go
```diff
@@ -1892,59 +1903,89 @@ func (t *clusterRequestReconcilerTask) updateNodeStatusWithHostname(ctx context.
 	return true
 }
 
-// updateHardwareProvisioningStatus updates the status for the created ClusterInstance
+// updateHardwareProvisioningStatus updates the status for the ClusterRequest
 func (t *clusterRequestReconcilerTask) updateHardwareProvisioningStatus(
-	ctx context.Context, nodePool *hwv1alpha1.NodePool) error {
+	ctx context.Context, nodePool *hwv1alpha1.NodePool) (bool, bool, error) {
```
edcdavid, line 1908: Optional: would it help using the same method as in https://github.com/openshift-kni/oran-o2ims/pull/210 to expose the transitions?
image

api/v1alpha1/clusterrequest_types.go
Missxiaoguo: I am adding `ClusterProvisionStartedAt metav1.Time` in the `ClusterDetails`

tliu2021: I will rename it.

internal/controllers/clusterrequest_controller.go
tliu2021: Thank you for your suggestions. I believe using an FSM would be overkill for this scenario. The current approach is simple and sufficient for the logic I’m implementing. Introducing an FSM would add unnecessary complexity and is more suited for handling complex transitions between multiple states. The current solution should work well while keeping the code clean.

edcdavid: Ok got it, in this case it might be simple enough to follow. Personally I find it easier to extract the logic from the code (even in this case). In my opinion, it makes it easier to test independently, understand and share. Also, for any similar timeout we could easily reuse the FSM code/testing

api/v1alpha1/clusterrequest_types.go
tliu2021: It has been renamed to HardwareProvisioningCheckStart since it tracks the timestamp of the first status check for hardware provisioning.

internal/controllers/clusterrequest_controller_test.go
```diff
@@ -443,7 +443,13 @@ func removeRequiredFieldFromClusterInstanceCm(
 func removeRequiredFieldFromClusterInstanceInput(
 	ctx context.Context, c client.Client, crName, crNamespace string) {
 	// Remove required field hostname
-	currentCR := &oranv1alpha1.ClusterRequest{}
+	currentCR := &oranv1alpha1.ClusterRequest{
+		Spec: oranv1alpha1.ClusterRequestSpec{
+			Timeout: oranv1alpha1.Timeout{
+				HardwareProvisioning: 1,
+			},
+		},
```
Missxiaoguo, line 451: I think this isn’t needed since it was configured previously. The queried ClusterRequest should already have this set

internal/controllers/clusterrequest_controller_test.go
```diff
@@ -592,6 +598,9 @@ var _ = Describe("ClusterRequestReconcile", func() {
 						Raw: []byte(testPolicyTemplateInput),
 					},
 				},
+				Timeout: oranv1alpha1.Timeout{
```
Missxiaoguo, line 601: Could we also verify that the start timestamp is popluated for hw provisioning in the some of testcases in the ClusterRequestReconcile?

internal/controllers/clusterrequest_controller.go
```diff
@@ -1649,45 +1661,42 @@ func (r *ClusterRequestReconciler) handleFinalizer(
...
 func (t *clusterRequestReconcilerTask) checkNodePoolProvisionStatus(ctx context.Context,
-	nodePool *hwv1alpha1.NodePool) bool {
+	nodePool *hwv1alpha1.NodePool) (bool, bool, error) {
 
 	// Get the generated NodePool and its status.
 	exists, err := utils.DoesK8SResourceExist(ctx, t.client, nodePool.GetName(),
 		nodePool.GetNamespace(), nodePool)
 
-	if err != nil || !exists {
+	if err != nil {
```
Missxiaoguo, line 1670: As err is now returned, I think we don't have to add error logs in each inner function if we are already logging the error at a the higher level.
For example, line1671-1676 and 1691-1696 and error log inside of updateHardwareProvisioningStatus

internal/controllers/clusterrequest_controller.go
```diff
@@ -1725,9 +1734,11 @@ func (t *clusterRequestReconcilerTask) waitForHardwareData(ctx context.Context,
 				nodePool.GetNamespace(),
 			),
 		)
-		return t.updateClusterInstance(ctx, clusterInstance, nodePool)
+		if !t.updateClusterInstance(ctx, clusterInstance, nodePool) {
```
Missxiaoguo, line 1737: Is there a plan to return err for updateClusterInstance func?

internal/controllers/clusterrequest_controller.go
```diff
@@ -213,7 +213,15 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (ctrl.Result, er
...
 
 	// wait for the NodePool to be provisioned and update BMC details in ClusterInstance
-	if !t.waitForHardwareData(ctx, renderedClusterInstance, renderedNodePool) {
+	provisioned, timedOutOrFailed, err := t.waitForHardwareData(ctx, renderedClusterInstance, renderedNodePool)
+	if err != nil {
+		return requeueWithError(err)
+	}
+	if timedOutOrFailed {
+		// Timeout occurred or failed, stop requeuing
+		return ctrl.Result{}, nil
```
Missxiaoguo, line 222:
```suggestion
		return doNotRequeue(), nil
```

internal/controllers/clusterrequest_controller_test.go
tliu2021: Removed.

internal/controllers/clusterrequest_controller_test.go
tliu2021: Yes, added a few Expect() for the the start timestamp.

internal/controllers/clusterrequest_controller.go
tliu2021: I removed the extra error logs.

internal/controllers/clusterrequest_controller.go
tliu2021: Currently, errors are consolidated into the message 'failed to update the rendered cluster instance' at the caller level, while the specific error details are logged for further troubleshooting. Since these are unlikely error conditions, I don't see a need to return the inner errors at this time.

internal/controllers/clusterrequest_controller.go
tliu2021: Changed.

internal/controllers/clusterrequest_controller.go
```diff
@@ -1902,59 +1901,83 @@ func (t *clusterRequestReconcilerTask) updateNodeStatusWithHostname(ctx context.
...
+		// No provisioning condition found, set the status to unknown.
+		status = metav1.ConditionUnknown
+		reason = string(utils.CRconditionReasons.Unknown)
+		message = "Unknown state of hardware provisioning"
+	}
+
+	// Check for timeout if not already failed or provisioned
+	if status != metav1.ConditionTrue && reason != string(hwv1alpha1.Failed) {
+		elapsedTime := time.Since(t.object.Status.NodePoolRef.HardwareProvisioningCheckStart.Time)
+		if elapsedTime >= time.Duration(t.object.Spec.Timeout.HardwareProvisioning)*time.Second {
```
irinamihai, line 1953: Should this be `time.Minute`?

tliu2021: Yes, just realize that it is in minutes, I saw the default 90 and assumed it was in seconds.

internal/controllers/clusterrequest_controller_test.go
```diff
@@ -1562,9 +1586,58 @@ var _ = Describe("waitForNodePoolProvision", func() {
...
+		Expect(c.Create(ctx, np)).To(Succeed())
+
+		// First call to checkNodePoolProvisionStatus (before timeout)
+		provisioned, timedOutOrFailed, err := task.checkNodePoolProvisionStatus(ctx, np)
+		Expect(provisioned).To(Equal(false))
+		Expect(timedOutOrFailed).To(Equal(false))
+		Expect(err).ToNot(HaveOccurred())
+
+		// Simulate a delay to represent timeout (greater than 1 seconds)
+		time.Sleep(2 * time.Second) // This ensures the timeout occurs
```
irinamihai, line 1629: I think it's preferred to edit `reconciledCR.Status.NodePoolRef.HardwareProvisioningCheckStart` than wait here.

internal/controllers/clusterrequest_controller.go
tliu2021: Fixed.

internal/controllers/clusterrequest_controller.go
```diff
@@ -273,7 +282,10 @@ func (t *clusterRequestReconcilerTask) checkClusterDeployConfigState(ctx context
 	nodePool := &hwv1alpha1.NodePool{}
 	nodePool.SetName(t.object.Status.NodePoolRef.Name)
 	nodePool.SetNamespace(t.object.Status.NodePoolRef.Namespace)
-	t.checkNodePoolProvisionStatus(ctx, nodePool)
+	_, _, err := t.checkNodePoolProvisionStatus(ctx, nodePool)
+	if err != nil {
+		return requeueWithError(err)
+	}
```
Missxiaoguo, line 288: Should this be applied here as well? Skip checking the rest of other resources status if it times out or fails and periodically check if hw provision is not done yet
```
	if timedOutOrFailed {
		// Timeout occurred or failed, stop requeuing
		return doNotRequeue(), nil
	}
	if !provisioned {
	   requeue
	}
```

internal/controllers/clusterrequest_controller.go
```diff
@@ -1723,10 +1720,10 @@ func (t *clusterRequestReconcilerTask) updateClusterInstance(ctx context.Context
 // waitForHardwareData waits for the NodePool to be provisioned and update BMC details
 // and bootMacAddress in ClusterInstance.
 func (t *clusterRequestReconcilerTask) waitForHardwareData(ctx context.Context,
-	clusterInstance *siteconfig.ClusterInstance, nodePool *hwv1alpha1.NodePool) bool {
+	clusterInstance *siteconfig.ClusterInstance, nodePool *hwv1alpha1.NodePool) (bool, bool, error) {
 
-	provisioned := t.checkNodePoolProvisionStatus(ctx, nodePool)
-	if provisioned {
+	provisioned, timedOutOrFailed, err := t.checkNodePoolProvisionStatus(ctx, nodePool)
```
Missxiaoguo, line 1725: Should we log the err here since the inner error logs are removed?

internal/controllers/clusterrequest_controller.go
tliu2021: This check will enforce the use of the test plugin for everyone. Are we sure we want to do that now?                                                                   ```
if !provisioned {
	   requeue
	}
```

internal/controllers/clusterrequest_controller.go
tliu2021: We have this  error log "Failed to update the NodePool status for ClusterRequest" in checkNodePoolProvisionStatus().

internal/controllers/clusterrequest_controller.go
Missxiaoguo: I think the workaround check has to be added as well. 
Even though we’re watching for NodePool status updates, since a timeout has been added, we’ll need to periodically check.

internal/controllers/clusterrequest_controller.go
Missxiaoguo: Okay, I was considering also logging the error returned from utils.DoesK8SResourceExist, but since we requeue on error, we should get the trackback in log for any temporary API issues. 
However, I think err var is missing in the log "Failed to update the NodePool status for ClusterRequest"

internal/controllers/clusterrequest_controller_test.go
tliu2021: It has been chanaged to move the start time back.

internal/controllers/clusterrequest_controller.go
tliu2021: Updated.

internal/controllers/clusterrequest_controller.go
tliu2021: The err var is added to the log.

internal/controllers/clusterrequest_controller.go
tliu2021: Yes.

PR #214, opened 2024-09-20
CNF-13594: Use IMAGE env variable as default image

alegacy:
This allows the --image parameter to default to whatever image reference has been specified in the IMAGE environment variable.  This is to facilitate the midstream build overlay which will set the image to be used to the official image reference.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -69,18 +69,12 @@ spec:
       - name: manager
         image: controller:latest
         env:
-        # Note that we have a kustomization replacement that copies the above `image` into this
-        # environment variable.
+          # Note that we have a kustomization replacement that copies the above `image` into this
+          # environment variable.  It is used as a default value for the --image command line argument
+          # since it and all other arguments are implemented as CLI arguments.
         - name: IMAGE
           value: controller:latest
         command:
-        # We use a shell script here in order to be able to pass the value of the `IMAGE`
-        # environment variable in the `--image` flag. We don't wan to read environment variables
-        # directly from the binary because all the other command use flags as their main
-        # configuration mechanism, and we don't want to deviate from that.
-        #
-        # Please keep the `exec` before the command, it removes the shell from the picture once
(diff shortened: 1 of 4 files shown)


config/manager/manager.yaml
```diff
@@ -69,18 +69,12 @@ spec:
       - name: manager
         image: controller:latest
         env:
-        # Note that we have a kustomization replacement that copies the above `image` into this
-        # environment variable.
+          # Note that we have a kustomization replacement that copies the above `image` into this
+          # environment variable.  It is used as a default value for the --image command line argument
+          # since it and all other arguments are implemented as CLI arguments.
         - name: IMAGE
           value: controller:latest
```
rauhersu, line 76: I see IMAGE has the value 'controller:latest' by default: is this customized somehow to end up pointing to an image to quay.io ? Just wondering if the current image '"quay.io/openshift-kni/oran-o2ims-operator:latest" passed as an argument prior to these changes is still valid for you after this.

rauhersu: Update : I believe you will get the new image from the CSV : https://github.com/openshift-kni/oran-o2ims/blob/main/bundle/manifests/oran-o2ims.clusterserviceversion.yaml#L1045 according the the kustomization.yaml comment, but just to double check.

alegacy: Yes, if the IMAGE variable is set then it will default to that; otherwise it will fail to launch the operator.

donpenney: When the bundle manifests are generated, the controller:latest gets replaced by the images.controller value in config/manager/kustomization.yaml, which is set to the IMG variable in the Makefile with:
```
cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
```
This is then used to generate the manifests for deploying the operator, or building the bundle.

donpenney (commented):
/lgtm

rauhersu (approved):

PR #215, opened 2024-09-24
CNF-13185: Enhancement for alarms

pixelsoccupied:
This enhancement talks about re-architecting the Alarm server as specified in `InfrastructureMonitoring Service API` o-ran spec. 

Notable changes include: 
- Data returned from API calls follow closely to what's defined by o-ran 
- Dynamically checking and gathering cluster resources during init such as PrometheuseRule
- Combining servers into the same code base 
- Introducing persistence storage via Postgres

This enhancement includes all the code tested during spike, the k8s resources needed to deploy through operator and other libraries/tool that can be used to quickly develop this.

co-authored with @browsell and @Jennifer-chen-rh

### DIFF
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -699,7 +699,8 @@ We will need few K8s resources that will be eventually applied by the Operator.
 #### Alarm server
 This is essentially a typical CRUD app and we need the following
 
-- Deployment: It should have one initContainers that runs to completion to perform DB migration and one main container which starts the main server. 
+- Deployment: It should have one initiContainer to create a DB, one initContainers that perform DB migration (create/alter table etc) 
+  and one main container which starts the main server. 
   No HA, so set replica to 1. It should also contain all the ENV variables needed to talk to postgres deployment DB_HOST, DB_PORT, DB_NAME, DB_USER etc.
   Suitable resources should be provided but not much memory and CPU is need for CRUD apps. 
 - Secrets: DB creds and configs should be read from postgres deployment. 
@@ -718,11 +719,13 @@ This deployment can be leveraged by many microservices by creating their own Dat
 - Secrets: default creds needed to spin postgres, with type: Opaque. This secret will then be ready by 
    - POSTGRES_USER: `o-ran`
    - POSTGRES_PASSWORD: `o-ran`
+   - POSTGRES_DB: `o-ran` # Note: this is simply there to be explicit.  
+                                   If not provided `POSTGRES_USER` is used to create default DB. But ultimately this DB not used as each service will create their own.
 - ConfigMap: For others to know the hostname and port
    - POSTGRES_HOST: "postgres.o-ran-namespace.svc.cluster.local"
    - POSTGRES_PORT: "5432"
 
-With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming migration already and `o-ran-infrastructure-monitoring-alarms` already created):
+With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming initContainer of app already created DB `o-ran-infrastructure-monitoring-alarms`):
 ```go
 connStr := "postgres://o-ran:o-ran@postgres.o-ran-namespace.svc.cluster.local:5432/o-ran-infrastructure-monitoring-alarms?sslmode=disable"
 ```
@@ -736,7 +739,7 @@ Note:
   It will expose all the external apis (o-ran) and internal apis (alertmanager + future h/w) and 
   handle them concurrently to drive all the features.
 - The HTTP server should be built with latest Go 1.22 `net/http` std lib. The latest update in the package brings in 
-  many requested features including mapping URI pattern. This allows to drop third party lib `gorilla/mux`.
+  many requested features including mapping URI pattern. This allows us to not rely on third party lib such as `gorilla/mux`.
 - Prefer creating structs to hold HTTP data for idiomatic Go code.
 - OpenAPI spec should be the source of truth. Other than standardization, free validation and documentation,
   with it, we can leverage a code generator such [this](https://github.com/oapi-codegen/oapi-codegen), allowing us to avoid writing boilerplate code. 
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -502,7 +502,7 @@ VALUES
     ('4.16'),
     ('4.17');
 
--- Most of the data in this table will come from k8s calls for now
+-- Most of the data in this table will come client-go calls
 INSERT INTO alarm_dictionary (alarm_dictionary_version, resource_type_id, entity_type, vendor)
 VALUES
     ('4.16', 'b3e7149e-d471-4d0f-aaa6-d5e9aa9e713a', 'telco-model-OpenShift-4.16.2', 'Red Hat'),
@@ -512,12 +512,11 @@ VALUES
 -- Insert into alarm_definitions
 INSERT INTO alarm_definitions (alarm_name, alarm_last_change, alarm_description, proposed_repair_actions , alarm_additional_fields)
 VALUES
-    ('NodeClockNotSynchronising','4.16', 'Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP is configured on this host.','{"CustomKey": "CustomValue"}'),
-    ('LowMemory','4.16','Low memory.','Low memory, with a longer description to help user fix the issue','{"CustomKey2": "CustomValue"}'),
-    ('NodeClockNotSynchronising','4.17','Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP2 is configured on this host.', '{"CustomKey3": "CustomValue"}');
+    ('NodeClockNotSynchronising','4.16', 'Clock not synchronising.','https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/NodeClockNotSynchronising.md','{"customKey": "customValue"}'),
+    ('NodeClockNotSynchronising','4.17','Clock not synchronising.','https://github.com/openshift/runbooks/blob/master/alerts/cluster-monitoring-operator/NodeClockNotSynchronising.md', '{"customKey": "customValue"}');
 -- The data here mapped from PrometheusRule. See ##PrometheusRule to AlarmDefinition mapping section for more.
 
--- probable_cause will be auto populated
+-- probable_cause will be auto populated. see the section "##Get Probable cause ID, name and description" for more details
 ```
 ## Detailed Server Instantiation
 
@@ -567,6 +566,7 @@ For a given OCP release, the alarmDefinitions and probableCauses are fixed, so t
               for: 10m
               labels: 
                 severity: critical #### (alarm_definitions.alarm_additional_fields)
+                customKey: customValue
  ```
 
 Notes on Init phase 
@@ -673,6 +673,9 @@ WHERE alarm_cleared_time < NOW() - INTERVAL '24 hour' and status = 'resolved';
 We can apply the CR before server starts and remove it during shutdown as part of teardown e.g inside `server.RegisterOnShutdown`
 
 ### Get Probable cause ID, name and description
+As soon as a new AlarmDefinition row is added, we generate a new UUID and add a row to probable_cause table automatically (leveraging postgres trigger function) .
+When a user queries for all or a specific probableCause, we will simply join the tables to get the data.
+
 ```sql
 SELECT
     probable_causes.probable_cause_id,
@@ -723,14 +726,7 @@ Note:
   This will allow for replicated persistent storage
 
 ## Tooling and general dev guidelines
-- Note on concurrency: The new alarm servers provide alarm query, alarm subscription, and alarm notification functional modules and 
-  will be running under the single daemon and single instance of the alarm daemon. 
-  The alarm data entries and alarm subscription entries and other related alarm management 
-  data will be stored in the postgres db tables described in this PR. 
-  1st phase alarm delivery has low scale of alarm subscriptions, and alarm counts to be supported. 
-  The new daemon will rely on the DB and the concurrency features offered by postgres db, and the usage of semaphore 
-  is tried to be avoided. The approach will be re-evaluated during performance and scaling optimization (preferably running as part of our CI)
-  as well as when more alarm features such as HW manager alarms is supported and when HA support is enabled.
+- All the features stated above will be developed within in the same controller driven by APIs that are published.
 - The HTTP server should be built with latest Go 1.22 `net/http` std lib. The latest update in the package brings in 
   many requested features including mapping URI pattern. This allows to drop third party lib `gorilla/mux`.
 - Prefer creating structs to hold HTTP data for idiomatic Go code.
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -325,8 +325,6 @@ alarmEventRecord := AlarmEventRecord{
 }
 ```
 
-
-
 ## Schema
 All O-RAN services will use the same O-CLOUD DB service. More on DB deployment [here](#postgres).
 
@@ -693,7 +691,7 @@ This is essentially a typical CRUD app and we need the following
 - Deployment: It should have one initContainers that runs to completion to perform DB migration and one main container which starts the main server. 
   No HA, so set replica to 1. It should also contain all the ENV variables needed to talk to postgres deployment DB_HOST, DB_PORT, DB_NAME, DB_USER etc.
   Suitable resources should be provided but not much memory and CPU is need for CRUD apps. 
-- Secrets and ConfigMap: DB creds and configs should be read from here. We can probably use the configs from Postgres (seen below) for user and password. 
+- Secrets: DB creds and configs should be read from postgres deployment. 
 - Service: Expose and balance using `ClusterIP` (though to start with we will set replica to 1)
 - Ingress: Expose service so that it can be called by users from outside the cluster
 
@@ -706,7 +704,21 @@ This deployment can be leveraged by many microservices by creating their own Dat
   Deployment should also mount to the right path `/var/lib/postgresql/data` (please check the latest docs).
 - PersistentVolumeClaim: At least 20G PVC should be used. Using test, we used about 60k rows DB which took about ~2Gb
 - Service: ClusterIP should be good as it's only used within the cluster. 
-- Secrets and Config: default creds needed to spin postgres 
+- Secrets: default creds needed to spin postgres, with type: Opaque. This secret will then be ready by 
+   - POSTGRES_USER: `o-ran`
+   - POSTGRES_PASSWORD: `o-ran`
+- ConfigMap: For others to know the hostname and port
+   - POSTGRES_HOST: "postgres.o-ran-namespace.svc.cluster.local"
+   - POSTGRES_PORT: "5432"
+
+With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming migration already and `o-ran-infrastructure-monitoring-alarms` already created):
+```go
+connStr := "postgres://o-ran:o-ran@postgres.o-ran-namespace.svc.cluster.local:5432/o-ran-infrastructure-monitoring-alarms?sslmode=disable"
+```
+
+Note:
+- ODF (ceph operator) must be deployed on the hub cluster that database requests a pvc from (same storageclass).
+  This will allow for replicated persistent storage
 
 ## Tooling and general dev guidelines
 - Note on concurrency: The new alarm servers provide alarm query, alarm subscription, and alarm notification functional modules and 
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -8,6 +8,7 @@ reviewers: # Include a comment about what domain expertise a reviewer is expecte
   - @alegacy
   - @bartwensley
   - @browsell
+  - @edcdavid
   - @Jennifer-chen-rh
 approvers: # A single approver is preferred, the role of the approver is to raise important questions, help ensure the enhancement receives reviews from all applicable areas/SMEs, and determine when consensus is achieved such that the EP can move forward to implementation.  Having multiple approvers makes it difficult to determine who is responsible for the actual approval.
   - @browsell
@@ -337,10 +338,10 @@ alarmEventRecord := AlarmEventRecord{
 ## Schema
 All O-RAN services will use the same O-CLOUD DB service. More on DB deployment [here](#postgres).
 
-Each table is modeled after O-RAN data structures. DB in our case wil be called `o-ran-infrastructure-monitoring-alarms`.
+Each table is modeled after O-RAN data structures. DB in our case wil be called `alarms`.
 Init SQL may look like the following:
 ```sql
-CREATE DATABASE o-ran-infrastructure-monitoring-alarms;
+CREATE DATABASE alarms;
 
 -- ENUM for ManagementInterfaceID
 DROP TYPE IF EXISTS ManagementInterfaceID CASCADE;
@@ -425,7 +426,7 @@ CREATE TABLE alarm_subscription_info (
              consumer_subscription_id UUID,
              filter TEXT,
              callback TEXT NOT NULL,
-             largest_number_alarm_event_seen_so_far BIGINT NOT NULL DEFAULT 0,
+             event_cursor BIGINT NOT NULL DEFAULT 0,
              created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP
 );
 
@@ -560,7 +561,7 @@ For a given OCP release, the alarmDefinitions and probableCauses are fixed, so t
 5. Take all the PromRules and build the AlarmDefinition. See [here](#prometheusrule-to-alarmdefinition-mapping) for mapping. 
    - A corresponding probablyCause row will be autometically added.
 
-6. Apply the required CR to activate the internal endpoint for alertmanager notification. See [here](#steps-internalv1caas-alertsalertmanager) for more.
+6. Apply the required CR to activate the internal endpoint for alertmanager notification. See [here](#steps-for-internalv1caas-alertsalertmanager) for more.
 
 7. The server should now be ready to take requests. 
 
@@ -658,7 +659,7 @@ Notes on Init phase
     SELECT aer.*
     FROM alarm_event_record aer
              JOIN alarm_subscription_info asi ON asi.alarm_subscription_id = 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11'
-    WHERE aer.alarm_sequence_number > asi.largest_number_alarm_event_seen_so_far
+    WHERE aer.alarm_sequence_number > asi.event_cursor
       and aer.perceived_severity = 'CRITICAL'
     ORDER BY aer.alarm_sequence_number;
     ```
@@ -673,7 +674,7 @@ Notes on Init phase
   And finally update 
   ```sql
   UPDATE alarm_subscription_info
-  SET largest_number_alarm_event_seen_so_far = $largestProcessedSequenceNumber
+  SET event_cursor = $largestProcessedSequenceNumber
   WHERE alarm_subscription_id = 'a2eebc99-9c0b-4ef8-bb6d-6bb9bd380a13';
   ```
 - NOTE: `alarm_sequence_number` is automatically handled from inside the DB. When the sequence increments, subscriber is notified. 
@@ -717,6 +718,9 @@ We will need few K8s resources that will be eventually applied by the Operator.
 We need two Jobs that can help with DB
 
 - One job that creates a Database using `CREATE DATBASE` cmd
+  - ALARMS_DB: `alarms`
+  - ALARMS_USER: `alarms`
+  - ALARM_PASSWORD: `alarms`
 - One job that creates all the tables as part of migration (create new tables, updates, etc)
   See [postgres](#postgres) for all the required credentials
 
@@ -724,7 +728,7 @@ We need two Jobs that can help with DB
 This is essentially a typical CRUD app and we need the following
 
 - Deployment: one main container which starts the main server. 
-  No HA, so set replica to 1. It should also contain all the ENV variables needed to talk to postgres deployment DB_HOST, DB_PORT, DB_USER etc (consult with POSTGRES deployment for latest).
+  No HA, so set replica to 1. It should also contain all the ENV variables needed to talk to postgres deployment ALARMS_DB, POSTGRES_PORT, ALARMS_USER etc (consult with POSTGRES deployment + Init JOB for alarms for latest).
   DB_NAME will be set to whatever is used in [here](#jobs-to-initialize-alarm-server-db).
   Suitable resources should be provided but not much memory and CPU is need for CRUD apps (TBD, need to experiment for exact values). 
 - Secrets: DB creds and configs should be read from postgres deployment. 
@@ -739,17 +743,18 @@ This deployment can be leveraged by many microservices by creating their own Dat
   Deployment should also mount to the right path `/var/lib/postgresql/data` (please check the latest docs).
 - PersistentVolumeClaim: At least 20G PVC should be used. Using test, we used about 60k rows DB which took about ~2Gb
 - Service: ClusterIP should be good as it's only used within the cluster. 
-- Secrets: default creds needed to spin postgres, with type: Opaque. This secret will then be ready by 
-   - POSTGRES_USER: `o-ran`
-   - POSTGRES_PASSWORD: `o-ran`
-   - POSTGRES_DB: `o-ran` # Note: this is simply there to be explicit. If not provided `POSTGRES_USER` is used to create default DB. But ultimately this DB will not be used as each service will create their own.
-- ConfigMap: For others to know the hostname and port
+- Secrets: default creds needed to spin postgres 
+   - POSTGRES_ADMIN_USER: `o-ran`
+   - POSTGRES_ADMIN_PASSWORD: `o-ran`
+- ConfigMap:
    - POSTGRES_HOST: "postgres.o-ran-namespace.svc.cluster.local"
    - POSTGRES_PORT: "5432"
+   - POSTGRES_DEFAULT_DB: `o-ran` # Note: this is simply there to be explicit. If not provided `POSTGRES_USER` is used to create default DB. But ultimately this DB will not be used as each service will create their own.
 
-With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming initContainer of app already created DB `o-ran-infrastructure-monitoring-alarms`):
+
+With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming initContainer of app already created DB `alarms`):
 ```go
-connStr := "postgres://o-ran:o-ran@postgres.o-ran-namespace.svc.cluster.local:5432/o-ran-infrastructure-monitoring-alarms?sslmode=disable"
+connStr := "postgres://alarms:alarms@postgres.o-ran-namespace.svc.cluster.local:5432/alarms?sslmode=disable"
 ```
 
 Note:
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -1,4 +1,4 @@
----
+```yaml
 title: lifecycle-of-infrastructure-monitoring-alarms-api
 authors:
   - @browsell
@@ -20,7 +20,7 @@ replaces:
   - "None"
 superseded-by:
   - "None"
----
+```
 
 # Lifecycle of infrastructure monitoring alarms api
 
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -316,7 +316,7 @@ alarmEventRecord := AlarmEventRecord{
 		ResourceTypeID:     "c9d3f9c5-8429-4484-8179-2a7977071bbf", // See section "Notes on Init phase" on how labels.managed_cluster mapped to resourceTypeID
 		ResourceID:        "d43bf16b-c9c6-432b-934e-7b670cc6a2cc", // labels.managed_cluster
 		AlarmRaisedTime:   "12-31-1920 12:32:14Z",                 // comes from alertmanager `startsAt`
-		AlarmAcknowledged: false,                                                        
+        AlarmAcknowledged: false,                                                        
 		PerceivedSeverity: SeverityCritical,                      // comes from labels.severity
 		Extensions:        []KeyValue{{Key: "namespace", Value: "openshift-cluster-version"}, ....}, // any labels that are not processed already (e.g skip labels.severity)
 		
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -35,6 +35,7 @@ superseded-by:
   - [Find AlarmDefinitionID and ProbableCauseID from current Alerts](#for-a-given-resourcetypeid-and-alarmname-coming-from-am-alert-find-the-alarmdefinitionid-and-probablecauseid)
   - [Notification tracking](#notification-tracking)
   - [Cleaning historical data](#daily-archive-cleanup)
+  - [Get ProbableCause ID, name and description](#get-probable-cause-id-name-and-description)
 - [Kubernetes](#k8s-resources)
 - [Tooling](#tooling-and-general-dev-guidelines)
 - [Future Updates](#future-updates)
@@ -50,8 +51,6 @@ OCP cluster resources to data structures defined by `O-RAN` spec. Among other th
 the service exposes APIs, configures Alertmanager deployment, read PrometheusRules from managedclusters and finally 
 store data in a persistent storage. 
 
-See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` for more (download from [here](https://specifications.o-ran.org/download?id=674)).
-
 ### Goals
 - Define steps to initialize and for when ready serve API calls
 - Define database schema
@@ -60,9 +59,10 @@ See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` for mor
 
 ## Key O-RAN data structures
 `InfrastructureMonitoring Service API Alarms`, primarily deals with the following O-RAN data structures during initialization. 
-Comments for each attribute is taken from O-RAN spec doc. 
+Comments for each attribute is taken from O-RAN spec doc.
 
 Please note that this is not an exhaustive list but are here to help the reader get a feel for the Alarm specific data we are dealing with.
+See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` (download from [here](https://specifications.o-ran.org/download?id=674)) for more.
 
 - AlarmDictionary
     This is primarily the link between Alarms and Inventory. A ResourceType (currently we are mostly dealing with type "cluster") can have exactly one AlarmDictionary.
@@ -147,6 +147,8 @@ Please note that this is not an exhaustive list but are here to help the reader 
     ```
 
 ## Infrastructure Monitoring Service Alarms API
+See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` (download from [here](https://specifications.o-ran.org/download?id=674)) to check APIs that we need to expose.
+
 | **Endpoint**                                                                            | **HTTP Method** | **Description**                                                     | **Input Payload**                                                            | **Returned Data**                   |
 |-----------------------------------------------------------------------------------------|-----------------|---------------------------------------------------------------------|------------------------------------------------------------------------------|-------------------------------------|
 | `/O2ims_infrastructureMonitoring/{apiVersion}/alarms`                                   | GET             | Retrieve the list of alarms.                                        | Optional query parameters `filter`                                           | A list of `AlarmEventRecord`        |
@@ -167,10 +169,10 @@ Please note that this is not an exhaustive list but are here to help the reader 
 
 
 
-| **Internal Endpoint**                           | **HTTP Method** | **Description**                              | **Input Payload**                                                        | **Returned Data** |
-|-------------------------------------------------|-----------------|----------------------------------------------|--------------------------------------------------------------------------|-------------------|
-| `/internal/v1/caas-alerts/alertmanager`         | POST            | Alertmanager notifications come through here | https://prometheus.io/docs/alerting/latest/configuration/#webhook_config | None              |
-| `/internal/v1/hardware-alerts/{hw-vendor-name}` | POST            | TBD                                          | TBD                                                                      | TBD               |
+| **Internal Endpoint**                           | **HTTP Method** | **Description**                              | **Input Payload**                                                                                | **Returned Data** |
+|-------------------------------------------------|-----------------|----------------------------------------------|--------------------------------------------------------------------------------------------------|-------------------|
+| `/internal/v1/caas-alerts/alertmanager`         | POST            | Alertmanager notifications come through here | Payload defined [here](https://prometheus.io/docs/alerting/latest/configuration/#webhook_config) | None              |
+| `/internal/v1/hardware-alerts/{hw-vendor-name}` | POST            | TBD                                          | TBD                                                                                              | TBD               |
 
 
 ### `alarms` family
@@ -618,6 +620,18 @@ DELETE FROM alarm_event_record_archive
 WHERE alarm_cleared_time < NOW() - INTERVAL '24 hour' and status = 'resolved';
 ```
 We can apply the CR before server starts and remove it during shutdown as part of teardown e.g inside `server.RegisterOnShutdown`
+
+### Get Probable cause ID, name and description
+```sql
+SELECT
+    probable_causes.probable_cause_id,
+    ad.alarm_name,
+    ad.alarm_description
+FROM probable_causes
+         JOIN alarm_definitions ad
+              on ad.alarm_definition_id = probable_causes.alarm_definition_id
+WHERE probable_causes.probable_cause_id = 'f5ac4ac7-0ff3-40a2-b305-77313c28136a';
+```
 
 ## K8s resources
 We will need few K8s resources that will be eventually applied by the Operator. 
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -41,15 +41,14 @@ superseded-by:
 
 ## Summary
 
-`o-ran` requires `InfrastructureMonitoring Service API` which is a collection of APIs that can be queried by client to 
-monitor the health of the `o-cloud`. This enhancement describes initialization steps and ready steps for `InfrastructureMonitoring Service API`.
-
-More specifically we will describe everything needed for `Alarms`. 
+`o-ran` requires `InfrastructureMonitoring Service Alarms API` which is a collection of APIs that can be queried by client to 
+monitor the health of the `o-cloud`. This enhancement describes initialization steps and ready steps for `InfrastructureMonitoring Service Alarms API` 
+as well everything that's needed to fully develop the app.
 
 At a high level, this service can be viewed as a thin wrapper of ACM observability stack which translates 
 OCP cluster resources to data structures understood and defined by `o-ran` spec. Among other things
 the service exposes APIs, configures Alertmanager deployment, read PrometheusRules from managedclusters and finally 
-store data in a persistence storage. 
+store data in a persistent storage. 
 
 ### Goals
 - Define steps to initialize and for when ready serve API calls
--- a/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
+++ b/docs/enhancements/infrastructure-monitoring-service-api/alarms.md
@@ -153,23 +153,23 @@ See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` (downlo
 ## Infrastructure Monitoring Service Alarms API
 See the official doc `O-RAN.WG6.O2IMS-INTERFACE-R003-v06.00 (June 2024)` (download from [here](https://specifications.o-ran.org/download?id=674)) to check APIs that we need to expose.
 
-| **Endpoint**                                                                       | **HTTP Method** | **Description**                                                     | **Input Payload**                                                            | **Returned Data**                   |
-|------------------------------------------------------------------------------------|-----------------|---------------------------------------------------------------------|------------------------------------------------------------------------------|-------------------------------------|
-| `/O2ims_infrastructureMonitoring/v1alpha/alarms`                                   | GET             | Retrieve the list of alarms.                                        | Optional query parameters `filter`                                           | A list of `AlarmEventRecord`        |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarms/{alarmEventRecordId}`              | GET             | Retrieve exactly one alarm identified by `alarmEventRecordId`.      | None                                                                         | Exactly one `AlarmEventRecord`      |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarms/{alarmEventRecordId}`              | PATCH           | Modify exactly one alarm identified by `alarmEventRecordId` to ack. | `AlarmEventRecordModifications`(no perceivedSeverity only alarmAcknowledged) | `AlarmEventRecordModifications`     |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarmSubscriptions`                       | GET             | Retrieve the list of alarm subscriptions.                           | Optional query parameters `filter`                                           | A list of `AlarmSubscriptionInfo`   |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarmSubscriptions`                       | POST            | Create a new alarm subscriptions.                                   | `AlarmSubscriptionInfo`                                                      | Exactly one `AlarmSubscriptionInfo` |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarmSubscriptions/{alarmSubscriptionId}` | GET             | Retrieve exactly one subscription using `alarmSubscriptionId`.      | None                                                                         | Exactly one `AlarmSubscriptionInfo` |
-| `/O2ims_infrastructureMonitoring/v1alpha/alarmSubscriptions/{alarmSubscriptionId}` | DELETE          | Delete exactly one subscription using `alarmSubscriptionId`.        | None                                                                         | None                                |
-| `/O2ims_infrastructureMonitoring/v1alpha/probableCause`                            | GET             | Retrieve all probable causes                                        | None                                                                         | A list of `ProbableCause`           |
-| `/O2ims_infrastructureMonitoring/v1alpha/probableCause/{probableCauseId}`          | GET             | Retrieve exactly one probable cause using `probableCauseId`.        | None                                                                         | Exactly one `ProbableCause`         |
+| **Endpoint**                                                                  | **HTTP Method** | **Description**                                                     | **Input Payload**                                                            | **Returned Data**                   |
+|-------------------------------------------------------------------------------|-----------------|---------------------------------------------------------------------|------------------------------------------------------------------------------|-------------------------------------|
+| `/O2ims_infrastructureMonitoring/v1/alarms`                                   | GET             | Retrieve the list of alarms.                                        | Optional query parameters `filter`                                           | A list of `AlarmEventRecord`        |
+| `/O2ims_infrastructureMonitoring/v1/alarms/{alarmEventRecordId}`              | GET             | Retrieve exactly one alarm identified by `alarmEventRecordId`.      | None                                                                         | Exactly one `AlarmEventRecord`      |
+| `/O2ims_infrastructureMonitoring/v1/alarms/{alarmEventRecordId}`              | PATCH           | Modify exactly one alarm identified by `alarmEventRecordId` to ack. | `AlarmEventRecordModifications`(no perceivedSeverity only alarmAcknowledged) | `AlarmEventRecordModifications`     |
+| `/O2ims_infrastructureMonitoring/v1/alarmSubscriptions`                       | GET             | Retrieve the list of alarm subscriptions.                           | Optional query parameters `filter`                                           | A list of `AlarmSubscriptionInfo`   |
+| `/O2ims_infrastructureMonitoring/v1/alarmSubscriptions`                       | POST            | Create a new alarm subscriptions.                                   | `AlarmSubscriptionInfo`                                                      | Exactly one `AlarmSubscriptionInfo` |
+| `/O2ims_infrastructureMonitoring/v1/alarmSubscriptions/{alarmSubscriptionId}` | GET             | Retrieve exactly one subscription using `alarmSubscriptionId`.      | None                                                                         | Exactly one `AlarmSubscriptionInfo` |
+| `/O2ims_infrastructureMonitoring/v1/alarmSubscriptions/{alarmSubscriptionId}` | DELETE          | Delete exactly one subscription using `alarmSubscriptionId`.        | None                                                                         | None                                |
+| `/O2ims_infrastructureMonitoring/v1/probableCause`                            | GET             | Retrieve all probable causes                                        | None                                                                         | A list of `ProbableCause`           |
+| `/O2ims_infrastructureMonitoring/v1/probableCause/{probableCauseId}`          | GET             | Retrieve exactly one probable cause using `probableCauseId`.        | None                                                                         | Exactly one `ProbableCause`         |
 
 
-| **Endpoint for clients but not currently in spec**                         | **HTTP Method** | **Description**                                              | **Input Payload** | **Returned Data**           |
-|----------------------------------------------------------------------------|-----------------|--------------------------------------------------------------|-------------------|-----------------------------|
-| `/O2ims_infrastructureMonitoring/v1alpha/probableCauses`                   | GET             | Retrieve all probable causes                                 | None              | A list of `ProbableCause`   |
-| `/O2ims_infrastructureMonitoring/v1alpha/probableCauses/{probableCauseId}` | GET             | Retrieve exactly one probable cause using `probableCauseId`. | None              | Exactly one `ProbableCause` |
+| **Endpoint for clients but not currently in spec**                    | **HTTP Method** | **Description**                                              | **Input Payload** | **Returned Data**           |
+|-----------------------------------------------------------------------|-----------------|--------------------------------------------------------------|-------------------|-----------------------------|
+| `/O2ims_infrastructureMonitoring/v1/probableCauses`                   | GET             | Retrieve all probable causes                                 | None              | A list of `ProbableCause`   |
+| `/O2ims_infrastructureMonitoring/v1/probableCauses/{probableCauseId}` | GET             | Retrieve exactly one probable cause using `probableCauseId`. | None              | Exactly one `ProbableCause` |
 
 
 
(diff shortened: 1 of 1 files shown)


docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+	AlarmChangeType       AlarmLastChangeType     `json:"alarmChangeType"`                 // M, 1, Indicates the type of change that occurred during the alarm last change; added, deleted, modified.
+	AlarmDescription      string                  `json:"alarmDescription"`                // M, 1, Provides a longer descriptive meaning of the alarm condition and a description of the consequences of the alarm condition. This is intended to be read by an operator to give an idea of what happened and a sense of the effects, consequences, and other impacted areas of the system
+	ProposedRepairActions string                  `json:"proposedRepairActions"`           // M, 1, Provides guidance for proposed repair actions.
+	ClearingType          ClearingType            `json:"clearingType"`                    // M, 1, Whether alarm is cleared automatically or manually
+	ManagementInterfaceID []ManagementInterfaceID `json:"managementInterfaceId,omitempty"` // M, 0..N, List of management interface over which alarms are transmitted for this Entity Type. RESTRICTION: For the O-Cloud IMS Services this value is limited to O2IMS.
```
Jennifer-chen-rh, line 76: Slice? only O2IMS is applied to our extension.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+- AlarmEventRecord
+
+```go
+type AlarmEventRecord struct {
+	AlarmEventRecordID    uuid.UUID         `json:"alarmEventRecordId"`              // M, Identifier of an entry in the AlarmEventRecord. Locally unique within the scope of an O-Cloud instance.
+	ResourceID            uuid.UUID         `json:"resourceId"`                      // M, A reference to the resource instance which caused the alarm.
+    AlarmDefinitionID     uuid.UUID         `json:"alarmDefinitionId"`               // M, A reference to the Alarm Definition record in the Alarm Dictionary associated with the referenced Resource Type.
+    ProbableCauseID       uuid.UUID         `json:"probableCauseId"`                 // M, A reference to the ProbableCause of the Alarm.
+    AlarmRaisedTime       time.Time         `json:"alarmRaisedTime"`                 // M, This field is populated with a Date/Time stamp value when the AlarmEventRecord is created.
+    AlarmChangedTime      *time.Time        `json:"alarmChangedTime,omitempty"`      // M, This field is populated with a Date/Time stamp value when any value of the AlarmEventRecord is modified.
```
Jennifer-chen-rh, line 100: Why need pointers here?

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+    PerceivedSeverity     PerceivedSeverity `json:"perceivedSeverity"`               // M, This is an enumerated set of values which identify the perceived severity of the alarm.
+    Extensions            []KeyValue        `json:"extensions"`                      // M, These are unspecified (not standardized) properties (keys) which are tailored by the vendor or operator to extend the information provided about the O-Cloud Alarm.
+}
+```
+
+- Subscriber 
+
+```go
+type AlarmSubscriptionInfo struct {
+    AlarmSubscriptionID         uuid.UUID `json:"alarmSubscriptionID"`
```
Jennifer-chen-rh, line 113: New standard has subscriptionID

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+3. Response with AlarmSubscriptionInfo
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarmSubscriptions/{alarmSubscriptionId}` with GET
+1. Client calls with an `alarmSubscriptionId`
+2. Query the storage `alarm_subscription_info`
+3. Response with retrieved instance of AlarmSubscriptionInfo
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarmSubscriptions/{alarmSubscriptionId}` with GET
+1. Client calls with an `alarmSubscriptionId`
+2. Query the storage `alarm_subscription_info` with delete
```
Jennifer-chen-rh, line 174: ?

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+|-------------------------------------------------|-----------------|-----------------------------------------------|--------------------------------------------------------------------------|-------------------|
+| `internal/v1/caas-alerts/alertmanager`          | POST            | Alertmanager notifications comes through here | https://prometheus.io/docs/alerting/latest/configuration/#webhook_config | None              |
+| `internal/v1/hardware-alerts/{hw-vendor-name}}` | POST            | TBD                                           | TBD                                                                      | TBD               |
+
+
+
+### `alarms` family
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarms` with GET
+1. Get active alarms from `alarm_event_record` table
```
Jennifer-chen-rh, line 143: 2. we need apply the filter via DB select schemas

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+3. Response with retrieved list of AlarmEventRecord
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarms/{alarmEventRecordId}` with GET
```
Jennifer-chen-rh, line 146: We should also query the history database if it is not active table

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarms/{alarmEventRecordId}` with PATCH
+1. Client calls with an AlarmEventRecordID and `AlarmEventRecordModifications` as patch payload
+2. If AlarmEventRecordModifications.alarmAcknowledged is True, update `alarm_event_record` table
+3. Response with `AlarmEventRecordModifications`
+
+### `alarmSubscriptions` family
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarmSubscriptions` with GET
+1. Query the storage `alarm_subscription_info`
```
Jennifer-chen-rh, line 159: also need filter the records if filter is present.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+3. Response with a list of AlarmSubscriptions
```
Jennifer-chen-rh, line 160: Also there is possibility that empty payload is returned.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarmSubscriptions` with POST
+1. Client calls with `AlarmSubscriptionInfo` payload
+2. Query the storage `alarm_subscription_info` 
+3. Response with AlarmSubscriptionInfo
+
+### Steps for `/O2ims_infrastructureMonitoring/{apiVersion}/alarmSubscriptions/{alarmSubscriptionId}` with GET
+1. Client calls with an `alarmSubscriptionId`
+2. Query the storage `alarm_subscription_info`
+3. Response with retrieved instance of AlarmSubscriptionInfo
```
Jennifer-chen-rh, line 170: according to standard, the corresponding code or error code should be returned accordingly.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+
+A minimal configuration
+```yaml
+route:
+  receiver: webhook_receiver
+receivers:
+  - name: webhook_receiver
+    webhook_configs:
+      - url: "o-ran-inventory-api-alarms.kubernetes.svc/internal/v1/caasAlerts/alertmanager"
+        send_resolved: true
```
Jennifer-chen-rh, line 192: We need to discuss with Brent about if we should filter out the OCP management alert we talked earlier today.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Maybe Im missing something here....this is simply a get query on the DB with no query parameters. Where is the filter coming from? 

Essentially get all AlarmEventRecord right?

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+          ON ad.alarm_definition_id = pc.alarm_definition_id
+    WHERE
+        adict.resource_type_id = '481688c8-2782-4534-a9de-88ca5154411d'
+        AND ad.alarm_name = 'LowMemory';
+    ```
+
+
+### Notification tracking
+
+- Collect all subscriber info including ID, callback and filter
```
Jennifer-chen-rh, line 506: subscription

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+- Collect all the AlarmEventRecord based on sequence number and optionally the filter. Here we are collecting everything that's "CRITICAL"
+    ```sql
+    SELECT aer.*
+    FROM alarm_event_record aer
+             JOIN alarm_subscription_info asi ON asi.alarm_subscription_id = 'a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11'
```
Jennifer-chen-rh, line 511: We should loop though the alarm_subscription_info table?

Jennifer-chen-rh (changes requested):
General question about DB serial number. It will increase with the entry of DB rows or increase with DB row update?

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
Jennifer-chen-rh: I will check the recording by Daniel

pixelsoccupied:
@Jennifer-chen-rh on entry. More on SERIAL datatypes https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-SERIAL

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Good point. We can do one "full" query instead of looping and also using `*` is a bad idea. This is query is just show that it's possible to grab only the latest events for each subscriber...and definitely not optimized.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Pointer simply there to indicate that the value can be null e.g AlarmChangedTime. AlarmRaisedTime on the other hand must be set (not pointer).

Jennifer-chen-rh (commented):
@pixelsoccupied aha, I felt something missing in PR. Now I figured out that he section we discussed about history table entry age out not here.

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Just following the spec as spoke about it internally

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Added doc for `?filter`

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: Like we spoke about. By default we will only look at active alerts. But user may access resolved if provided with `?filter` query

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: fixed

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: updated doc to include filter

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: updated to indicate returning a suitable error code

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: like we spoke about it internally...not doing this

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+            "generatorURL": "https://console-openshift-console.apps.cnfdf27.sno.telco5gran.eng.rdu2.redhat.com/monitoring/graph?g0.expr=sum+by+%28channel%2C+namespace%2C+upstream%29+%28cluster_version_available_updates%29+%3E+0&g0.tab=1",
+            "fingerprint": "91406cd113ad87e5"
+        },
+   ]
+    ```
+2. Sync `alarm_event_record` Table
+   - Update rows to "resolved" that are missing in the current payload (i.e previously seen but somehow missed the "resolved" notification)
+   - Upsert all the "firing" and "Resolved" alerts as indicated by `alerts[].status`. 
+   - The table will auto archive the resolved alerts by moving the row to another table. See Schema section for more.
+3. Grab the Subscribers and send notification.
```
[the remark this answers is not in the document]
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+            },
+            "startsAt": "2024-08-28T11:39:17.958Z",
+            "endsAt": "0001-01-01T00:00:00Z",
+            "generatorURL": "https://console-openshift-console.apps.cnfdf27.sno.telco5gran.eng.rdu2.redhat.com/monitoring/graph?g0.expr=sum+by+%28channel%2C+namespace%2C+upstream%29+%28cluster_version_available_updates%29+%3E+0&g0.tab=1",
+            "fingerprint": "91406cd113ad87e5"
+        },
+   ]
+    ```
+2. Sync `alarm_event_record` Table
+   - Update rows to "resolved" that are missing in the current payload (i.e previously seen but somehow missed the "resolved" notification)
```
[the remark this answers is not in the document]
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/lifecycle.md
```diff
@@ -0,0 +1,530 @@
...
+            "annotations": {
+                "description": "For more information refer to 'oc adm upgrade' or https://console-openshift-console.apps.cnfdf27.sno.telco5gran.eng.rdu2.redhat.com/settings/cluster/.",
+                "summary": "Your upstream update recommendation service recommends you update your cluster."
+            },
+            "startsAt": "2024-08-28T11:39:17.958Z",
+            "endsAt": "0001-01-01T00:00:00Z",
+            "generatorURL": "https://console-openshift-console.apps.cnfdf27.sno.telco5gran.eng.rdu2.redhat.com/monitoring/graph?g0.expr=sum+by+%28channel%2C+namespace%2C+upstream%29+%28cluster_version_available_updates%29+%3E+0&g0.tab=1",
+            "fingerprint": "91406cd113ad87e5"
+        },
+   ]
```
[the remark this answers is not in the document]
pixelsoccupied: updated

Jennifer-chen-rh (commented):
@pixelsoccupied 
1. the datastructures are still public
2. Missing the the alarm notification event object structure

pixelsoccupied:
@Jennifer-chen-rh 
updated 1) but will not include `AlarmEventNotification` as this is effectively `AlarmEventRecord`. As mentioned in the intro, the list of DS is not exhaustive but rather get a better understanding of o-ran data.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
bartwensley, line 1: Looks like there is a formatting error in the file - get this error when displaying the rich diff:
Error in user YAML: (<unknown>): found character that cannot start any token while scanning for the next token at line 3 column 5

Possibly missing quotes around the users?

pixelsoccupied: Should be good now! That top level block with title/author was being buggy

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+- AlarmSubscriptionInfo
+   This what stores info about subscription who needs to be notified when an Alarm is raised
+
+   3.3.6.2.3
+    ```go
+    type AlarmSubscriptionInfo struct {
+        SubscriptionID              uuid.UUID `json:"subscriptionID"`         // M, Identifier of the subscription. Locally unique within the scope of an O-Cloud instance.
+        ConsumerSubscriptionID      uuid.UUID `json:"consumerSubscriptionId"` // O, The consumer may provide its identifier for tracking, routing, or identifying the subscription used to report the event.
+        Filter                      string    `json:"filter"`                 // O, Criteria for events which do not need to be reported or will be filtered by the subscription notification service. Therefore, if a filter is not provided then all events are reported.
+        Callback                    string    `json:"callback"`               // M, The fully qualified URI to a consumer procedure which can process a Post of the AlarmEventNotification.
```
bartwensley, line 143: Just curious if any kind of authentication is used for this URI?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+2. Sync `alarm_event_record` Table
+   - Update rows to "resolved" that are missing in the current payload (i.e previously seen but somehow missed the "resolved" notification)
+     ```sql
+      UPDATE alarm_event_record
+      SET status = 'resolved', alarm_cleared_time = CURRENT_TIMESTAMP
+      WHERE (finger_print, alarm_raised_time) NOT IN (
+             ('Something-Thats-Only-Now-Available', '2023-09-01 10:02:00+00')
+      );
+     ```
+   - Create new AlarmEventRecord. Alert entry payload will require Alarm Definition ID and Probable Cause ID which can retrieved as shown [here](#for-a-given-resourcetypeid-and-alarmname-coming-from-am-alert-find-the-alarmdefinitionid-and-probablecauseid)
```
bartwensley, line 277: nit: the "here" link doesn't work for me - maybe just an issue with the preview in github?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+   - Upsert all AlarmEventRecord all the "firing" and "resolved" alerts as indicated by `alerts[].status` i.e a bulk insert + update operation
```
bartwensley, line 278: Upsert?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+  And finally update 
+  ```sql
+  UPDATE alarm_subscription_info
+  SET largest_number_alarm_event_seen_so_far = $largestProcessedSequenceNumber
+  WHERE alarm_subscription_id = 'a2eebc99-9c0b-4ef8-bb6d-6bb9bd380a13';
+  ```
+- NOTE: `alarm_sequence_number` is automatically handled from inside the DB. When the sequence increments, subscriber is notified. 
+  See notification conditions [here](#conditions-for-notifying-subscriber)
+  
+### Conditions for Notifying subscriber
```
bartwensley, line 598: Is this something O-RAN specifies? If so a reference would be good.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+      alarm_changed_time = EXCLUDED.alarm_changed_time;
+     ```
+3. Grab the Subscriptions and send notification. 
+   - Database interaction is further explained [here](#notification-tracking)
+4. Move all the `status: resolved` rows from `alarm_event_record` to `alarm_event_record_archive`
+
+Eventually data in `alarm_event_record_archive` will be cleared (hardcoded to 24hr) as seen [here](#daily-archive-cleanup-)
+
+## Schema
+We only take care of Alarms* data contained within a specific DB, this approach allows for
```
bartwensley, line 293: I don't understand what that means - can you elaborate or provide an example?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+    ('4.17', 'f8b9e100-fd9f-4923-b96f-89418e9c2560', 'telco-model-OpenShift-4.17.2', 'Red Hat');
+
+-- Insert into alarm_definitions
+INSERT INTO alarm_definitions (alarm_name, alarm_last_change, alarm_description, proposed_repair_actions , alarm_additional_fields)
+VALUES
+    ('NodeClockNotSynchronising','4.16', 'Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP is configured on this host.','{"CustomKey": "CustomValue"}'),
+    ('LowMemory','4.16','Low memory.','Low memory, with a longer description to help user fix the issue','{"CustomKey2": "CustomValue"}'),
+    ('NodeClockNotSynchronising','4.17','Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP2 is configured on this host.', '{"CustomKey3": "CustomValue"}');
+
+-- probable_cause will be auto populated
```
bartwensley, line 484: How will it be auto populated? Where will the data come from?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+    ('4.16', 'b3e7149e-d471-4d0f-aaa6-d5e9aa9e713a', 'telco-model-OpenShift-4.16.2', 'Red Hat'),
+    ('4.16', '481688c8-2782-4534-a9de-88ca5154411d', 'telco-model-OpenShift-4.16.8', 'Red Hat' ),
+    ('4.17', 'f8b9e100-fd9f-4923-b96f-89418e9c2560', 'telco-model-OpenShift-4.17.2', 'Red Hat');
+
+-- Insert into alarm_definitions
+INSERT INTO alarm_definitions (alarm_name, alarm_last_change, alarm_description, proposed_repair_actions , alarm_additional_fields)
+VALUES
+    ('NodeClockNotSynchronising','4.16', 'Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP is configured on this host.','{"CustomKey": "CustomValue"}'),
+    ('LowMemory','4.16','Low memory.','Low memory, with a longer description to help user fix the issue','{"CustomKey2": "CustomValue"}'),
+    ('NodeClockNotSynchronising','4.17','Clock not synchronising.','Clock at {{ $labels.instance }} is not synchronising. Ensure NTP2 is configured on this host.', '{"CustomKey3": "CustomValue"}');
```
bartwensley, line 482: Where does this data come from? How will we decide which alarms to define? Where does the proposed_repair_action come from?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+    apiVersion: cluster.open-cluster-management.io/v1
+    kind: ManagedCluster
+    metadata:
+      labels:
+        openshiftVersion-major-minor: "4.16"
+    ```
+
+- `alarm_dictionary` essentially links `ResourceTypeID` and `Version`. The conversion can be seen below.
+
+    | managed_cluster<br/>`from alerts`<br/> | resourceID  <br/>`same as managed_cluster`<br/> | resourceTypeID for caas<br/>`derived from (resourceID + ResourceKindLogical + ResourceClassCompute)`<br/> |
```
bartwensley, line 503: So initially we are only supporting the ManagedCluster resource type? Does this account for additional resource types in the future? What are some examples of resource types we might want to support in the future?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+                summary: Clock not synchronising. #### (alarm_definitions.alarm_description)
+              expr: |
+                min_over_time(node_timex_sync_status{job="node-exporter"}[5m]) == 0
+                and
+                node_timex_maxerror_seconds{job="node-exporter"} >= 16
+              for: 10m
+              labels: 
+                severity: critical #### (alarm_definitions.extensions)
+    ```
+    - Use ACM to get credentials of the unique major.minor clusters and retrieve all the PrometheusRules from them to parse.
```
bartwensley, line 534: Do PrometheusRules ever change or new ones get added (e.g. on a z-stream upgrade)? Or if a cluster gets installed with a release that wasn't being managed yet? What will be the strategy to handle that?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+This deployment can be leveraged by many microservices by creating their own Database. 
+
+- Deployment: One replica should be good enough for our case. 
+  Default username, password and dbname (note this is simply to allow postgres pod initialize) should be provided. 
+  Deployment should also mount to the right path `/var/lib/postgresql/data` (please check the latest docs).
+- PersistentVolumeClaim: At least 20G PVC should be used. Using test, we used about 60k rows DB which took about ~2Gb
+- Service: ClusterIP should be good as it's only used within the cluster. 
+- Secrets and Config: default creds needed to spin postgres 
+
+## Tooling and general dev guidelines
```
bartwensley, line 636: Can you add a brief note about how concurrency will be handled? I assume there can be multiple requests coming in parallel - GET/POST/PATCH/DELETE on the external API along with notifications from the AlertManager (and then potentially multiple callbacks to subscribers). Which of these events are going to be handled in parallel and will serialization be required for any of them? If so, what mechanism(s) will be used?

bartwensley (commented):
Looks good - I took a look and added some comments and questions.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,668 @@
...
+- Deployment: It should have one initContainers that runs to completion to perform DB migration and one main container which starts the main server. 
+  No HA, so set replica to 1. It should also contain all the ENV variables needed to talk to postgres deployment DB_HOST, DB_PORT, DB_NAME, DB_USER etc.
+  Suitable resources should be provided but not much memory and CPU is need for CRUD apps. 
+- Secrets and ConfigMap: DB creds and configs should be read from here. We can probably use the configs from Postgres (seen below) for user and password. 
+- Service: Expose and balance using `ClusterIP` (though to start with we will set replica to 1)
+- Ingress: Expose service so that it can be called by users from outside the cluster
+
+
+#### Postgres
+This deployment can be leveraged by many microservices by creating their own Database.
```
browsell, line 627: This needs more detail.

1. How are the credentials being exposed and consumed for the various servers in the o-cloud operator that need a DB
2. There is a requirement that you have replicated persistent storage 
3. How do the clients discover the ClusterIP ?
4. How are you configuring the DB, where does this configuration come from ?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,668 @@
...
+`o-ran` requires `InfrastructureMonitoring Service Alarms API` which is a collection of APIs that can be queried by client to 
+monitor the health of the `o-cloud`. This enhancement describes initialization steps and ready steps for `InfrastructureMonitoring Service Alarms API` 
+as well everything that's needed to fully develop the app.
+
+At a high level, this service can be viewed as a thin wrapper of ACM observability stack which translates 
+OCP cluster resources to data structures understood and defined by `o-ran` spec. Among other things
+the service exposes APIs, configures Alertmanager deployment, read PrometheusRules from managedclusters and finally 
+store data in a persistent storage. 
+
+### Goals
```
browsell, line 53: Another goal is to allow future integration of alarms from additional sources (i.e. H/W)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,668 @@
...
+`InfrastructureMonitoring Service API Alarms`, primarily deals with the following o-ran data structures during initialization. 
+Comments for each attribute is taken from o-ran spec doc. 
+
+Please note that this is not an exhaustive list but are here to help the reader get a feel for the Alarm specific data we are dealing with.
+
+- AlarmDictionary
+    This is primarily the link between Alarms and Inventory. A ResourceType (currently we are mostly dealing with type "cluster") can have exactly one AlarmDictionary.
+    
+    ```go
+    // 3.2.6.2.8-1
```
browsell, line 69: Expand

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,668 @@
...
+- Define developer tools
+
+## Key o-ran data structures
+`InfrastructureMonitoring Service API Alarms`, primarily deals with the following o-ran data structures during initialization. 
+Comments for each attribute is taken from o-ran spec doc. 
+
+Please note that this is not an exhaustive list but are here to help the reader get a feel for the Alarm specific data we are dealing with.
+
+- AlarmDictionary
+    This is primarily the link between Alarms and Inventory. A ResourceType (currently we are mostly dealing with type "cluster") can have exactly one AlarmDictionary.
```
browsell, line 66: NodeCluster

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell: Needs to be the same as the other services (i.e. inventory)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell, line 1: Please add walkthroughs, in particular the initialization sequence. 
Please describe how alerts are translated, and the mapping of alert fields (including the extensions) to alarm fields.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,667 @@
...
+
+`o-ran` requires `InfrastructureMonitoring Service API` which is a collection of APIs that can be queried by client to 
+monitor the health of the `o-cloud`. This enhancement describes initialization steps and ready steps for `InfrastructureMonitoring Service API`.
+
+More specifically we will describe everything needed for `Alarms`. 
+
+At a high level, this service can be viewed as a thin wrapper of ACM observability stack which translates 
+OCP cluster resources to data structures understood and defined by `o-ran` spec. Among other things
+the service exposes APIs, configures Alertmanager deployment, read PrometheusRules from managedclusters and finally 
+store data in a persistence storage.
```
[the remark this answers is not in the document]
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,667 @@
...
+
+### Goals
+- Define steps to initialize and for when ready serve API calls
+- Define database schema
+- Define K8s CRs
+- Define developer tools
+
+## Key o-ran data structures
+`InfrastructureMonitoring Service API Alarms`, primarily deals with the following o-ran data structures during initialization. 
+Comments for each attribute is taken from o-ran spec doc.
```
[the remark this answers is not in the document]
pixelsoccupied: Included the name and download link

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+        ClearingType          ClearingType            `json:"clearingType"`                    // M, 1, Whether alarm is cleared automatically or manually
+        ManagementInterfaceID []ManagementInterfaceID `json:"managementInterfaceId,omitempty"` // M, 0..N, List of management interface over which alarms are transmitted for this Entity Type. RESTRICTION: For the O-Cloud IMS Services this value is limited to O2IMS.
+        PKNotificationField   []string                `json:"pkNotificationField,omitempty"`   // M, 0..N, Identifies which field or list of fields in the alarm notification contains the primary key (PK) into the Alarm Dictionary for this interface; i.e. which field contains the Alarm Definition ID.
+        AlarmAdditionalFields []AttributeValuePair    `json:"alarmAdditionalFields,omitempty"` // M, 0..N, List of metadata key-value pairs used to associate meaningful metadata to the related resource type.
+    }
+    ```
+
+- ProbableCause
+
+    ProbableCause is a subset of data present in AlarmDefinition
```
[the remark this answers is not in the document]
pixelsoccupied: Yeah `ProbableCause` is somewhat vague in the interface doc. But basically when the user wants a get (or pushed) a new alerts event...the spec says to pass on the `ProbableCauseID` with it. And with the ID user may see Alert name and Alert description, and the name are description are copy of its corresponding AlarmDefinition. 

We are not exposing any API to get AlarmDefinition but we do have `/O2ims_infrastructureMonitoring/{apiVersion}/probableCause/{probableCauseId}` to get probablecause data. Note that this API is not in spec currently and we are the one introducing it (will update the doc reflect this)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Upsert is op that inserts rows into a database table if they do not already exist, or updates them if they do. In postgres its essentially `INSERT ... ON CONFLICT UPDATE`

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: No actually we dont have anything specific...but in the doc it says 
```
The Alarm Change Notification is used by the O-Cloud Infrastructure Monitoring Service to report changes to the O2ims alarm List to an O2ims consumer that has subscribed to such notifications.
``` 
So a change it our case it basically firing for the first time, resolved and when/if the alarmChangedTime updates.  I'll update the doc to include this info

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Yeah it's a bit confusing. Basically it's just to each service that needs a DB will create their own. 

e.g alarm server will run something 
```
CREATE DATABASE o-ran-infrastructure-monitoring-alarms;
``` 
a "myservice" will run 
```
CREATE DATABASE myservice;
``` 

From their own migration/init scripts

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell: Refer to 3.7.2 Alarm Notification Use Case in O-RAN-WG6.ORCH-USE-CASES-R003-v10.00

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell: Please clarify in the doc that you are creating the alarm service DB in the O-Cloud DB service

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: updated and also included the download link

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: updated to clarify

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: > Where does this data come from
Each definition is a Rule defined in PrometheusRule CR 

> How will we decide which alarms to define
We are retrieving all the  PrometheusRule CRs from unique major.minor clusters. I mentioned it in the doc but e.g if we are managing 3 clusters 4.16.2, 4.17.2 and 4.16.8, Pick 4.16.8 and 4.17.2 which effectively represents all the rules in 4.16.z and 4.17.z clusters

> Where does the proposed_repair_action come from
I changed the doc a bit more to understand the mapping (##PrometheusRule to AlarmDefinition mapping) but basically this is `rules.annotations.runbook_url` if available

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,669 @@
...
+    ```go
+    type AlarmSubscriptionInfo struct {
+        SubscriptionID              uuid.UUID `json:"subscriptionID"`         // M, Identifier of the subscription. Locally unique within the scope of an O-Cloud instance.
+        ConsumerSubscriptionID      uuid.UUID `json:"consumerSubscriptionId"` // O, The consumer may provide its identifier for tracking, routing, or identifying the subscription used to report the event.
+        Filter                      string    `json:"filter"`                 // O, Criteria for events which do not need to be reported or will be filtered by the subscription notification service. Therefore, if a filter is not provided then all events are reported.
+        Callback                    string    `json:"callback"`               // M, The fully qualified URI to a consumer procedure which can process a Post of the AlarmEventNotification.
+    }
+    ```
+
+## Infrastructure Monitoring Service Alarms API
```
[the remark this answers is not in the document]
pixelsoccupied: Added the link to interface doc

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: So yeah building on the previous comment (probableCause being a subset of alarmDefinition), as soon as a new AlarmDefinition row is added, we are generate a new UUID and add a row to `probable_cause` table automatically (leveraging postgres trigger function) .

When a user queries for all or a specific probableCause, we will simply join the tables to get the data. 
```
probableCauseID = newID
proableCasueName = AlarmDefinitionName
proableCasueDescription = AlarmDefinitionDescription
```

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Yeah exactly. So for now we are assuming ResourceType that's "Logical"(kind) and "Compute"(class) is effectively a cluster. Given that it's a cluster, we are building the alarmsDefinition based PromRules. And all the alerts (which we calling CaaS alerts) will be delivered to us from Alertmanager using one of the internal endpoints.  

With that in mind, we can potentially support any arbitrary ResourceTypes...as long as we know how to get the related alarmsDefinition (linked using alarmDictionary) and then potentially open up another endpoint which receives the alerts which then can converted to o-ran specific data structure for events. 

We briefly looked into distinguishing between cluster and node (so Physical (kind) + Compute (class)) but unfortunately the alerts that are coming from alertmanager dont do a good job telling if an alert is node specific of cluster specific (everything is cluster)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: > ever change or new ones get added

Yeah potentially? But at every restart we will audit the managedCluster (added this bit in the doc) and if we pick the latest one (i.e between 4.14.4 and 4.14.6 we pick 4.14.6) it should be mitigate some of concerns.

> release that wasn't being managed yet
We can't do that. If a cluster is not managed, hub cant send us the alerts.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Added a bit on it. But although we wont be doing HA for now (replica set to 1) we can assume the server can handle concurrency safely.  Postgres (but i think it's a SQL spec) comes with features of the box (e.g MVCC configured by default) for concurrent read and write which would allow us to write the code without too much headaches. 

Other than that we would add some tests using a simple http load generator and hopefully catch anything that we missed out during development. 

But overall I'm hoping there wont be too many gotchas along the way since such CRUD apps are quite common but will dig into this further as we develop.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,668 @@
...
+  - [Find AlarmDefinitionID and ProbableCauseID from current Alerts](#for-a-given-resourcetypeid-and-alarmname-coming-from-am-alert-find-the-alarmdefinitionid-and-probablecauseid)
+  - [Notification tracking](#notification-tracking)
+  - [Cleaning historical data](#daily-archive-cleanup)
+- [Kubernetes](#k8s-resources)
+- [Tooling](#tooling-and-general-dev-guidelines)
+- [Future Updates](#future-updates)
+
+## Summary
+
+`o-ran` requires `InfrastructureMonitoring Service Alarms API` which is a collection of APIs that can be queried by client to
```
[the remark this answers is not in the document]
pixelsoccupied: Updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: fixed

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Added back the init walk through. Also included a "## Alertmanager Example payload" section which shows the mapping in Go

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Added all the info, including default username and password, and where it can be stored. Also an example connection string that an can use which includes how client can discover the postgres service

Also specified that ODF is required in the hub

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell: Please update the section above to reflect your comments

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
browsell: add to the doc

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,763 @@
...
+Note:
+- ODF (ceph operator) must be deployed on the hub cluster that database requests a pvc from (same storageclass).
+  This will allow for replicated persistent storage
+
+## Tooling and general dev guidelines
+- Note on concurrency: The new alarm servers provide alarm query, alarm subscription, and alarm notification functional modules and 
+  will be running under the single daemon and single instance of the alarm daemon. 
+  The alarm data entries and alarm subscription entries and other related alarm management 
+  data will be stored in the postgres db tables described in this PR. 
+  1st phase alarm delivery has low scale of alarm subscriptions, and alarm counts to be supported.
```
browsell, line 730: Not relevant

browsell: I think we need more detail here.  How is the functionality being partitioned - Query, notifications, subscription handling,  alert manager interface. 

This all be done within a single controller ?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Added a new section "Get Probable cause ID, name and description" and example query

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Updated the example with real data

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -303,6 +304,27 @@ Eventually data in `alarm_event_record_archive` will be cleared (hardcoded to 24
...
+
+```go
+alarmEventRecord := AlarmEventRecord{
+		AlarmDefinitionID: "82139b9c-3683-427b-a1f5-4bf9d6dbb0d8", // See section "For a given ResourceTypeID and AlarmName (coming from AM alert), find the AlarmDefinitionID and ProbableCauseID"
+		ProbableCauseID:   "7fe31219-c2a2-48c7-a2f5-f4bcfe0486e7", // See section "For a given ResourceTypeID and AlarmName (coming from AM alert), find the AlarmDefinitionID and ProbableCauseID"
+        AlarmEventRecordID: uuid.New(), // Auto generated with DB                 
+		ResourceTypeID:     "c9d3f9c5-8429-4484-8179-2a7977071bbf", // See section "Notes on Init phase" on how labels.managed_cluster mapped to resourceTypeID
+		ResourceID:        "d43bf16b-c9c6-432b-934e-7b670cc6a2cc", // labels.managed_cluster
+		AlarmRaisedTime:   "12-31-1920 12:32:14Z",                              // comes from alertmanager `startsAt`
+		AlarmAcknowledged: false,
```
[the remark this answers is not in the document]
pixelsoccupied: removed

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Removed the irrelevant text. 

Added the info on functionality. But yeah i think only one server can do the job. It will expose all the external apis (o-ran) and internal apis (alertmanager + future h/w) and handle them concurrently to drive all the features.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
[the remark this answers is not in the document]
bartwensley: OK - but I still don't see how "ProbableCause is a subset of data present in AlarmDefinition". The ProbablCause.Name is the name of the ProbableCause - not the name of the Alert and same for the Description - right?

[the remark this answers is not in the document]
bartwensley: Ahh... I see you explain this later - we've decided to use the AlarmName and a reference to the Alarm Description in the runbook repo and not create something new.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
bartwensley: Thanks Nahian - might be good to add a sentence or two about that in the summary.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
bartwensley: I was thinking of the case where a hub had only 4.16 clusters installed (for example) and the the first 4.18 cluster is installed. What would trigger the alarm service to read the 4.18 alarm definitions in that case?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
bartwensley: When you say "handle them concurrently" do you mean through goroutines? I assume this will be necessary, since it won't be OK to block the server while (for example) a notification is being sent to an external subscriber.

pixelsoccupied: > through goroutines?

Yeah exactly! The handler starts a new goroutine on every request automatically. So nothing will be blocking. 

As an example, if user calls `/alarms` api and at the alartmanager calls `/caas-alerts` api, it will effectively be two routines. So we can return the `/alarms` call and notify subscribers all at the same time.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: > What would trigger the alarm service to read the 4.18 alarm definitions 

Currently nothing tbh. So the only way to get the latest data (e.g new 4.18 cluster) we would be need to restart. But to get this info without a full restart we can need to setup some kind of "watch" over the managedcluster CRs. It could be done with the operator (so operator would requeue when it detects a change) and simply delete the alarms pod OR alarms server would directly watch the CRs and update DB as needed. But ultimately though...this could be something Inventory/resource server should alert us with("btw new cluster/resourceType added...update your DB")?

Maybe @browsell can weigh in on this.

bartwensley: Could you just read the alarm definitions for all supported versions?

pixelsoccupied: i dont think that's possible since the available clusters tells us which definitions to store. 

We are retrieving PromRules and then generating alarm definitions (and finally storing it) based on the currently set of major.minor cluster as indicated by hub. In other words...we need first know that there's a new 4.18 cluster before actually retrieving and storing the definitions. And currently the only way to know this by auditing at startup

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Added this bit under "Steps for `internal/v1/caas-alerts/alertmanager`" . Also explains how we can potentially serve any ResourceType

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,770 @@
...
+endpoint if alert aggregator is not Alertmanager. 
+
+A minimal configuration
+```yaml
+route:
+  receiver: webhook_receiver
+receivers:
+  - name: webhook_receiver
+    webhook_configs:
+      - url: "o-ran-inventory-api-alarms.kubernetes.svc/internal/v1/caasAlerts/alertmanager" # this will be derived from deployment
```
alegacy, line 251: wouldn't the address/hostname part here be something like:  alarm-server.oran-o2ims.svc.cluster.local using the typical convention of <service>.<namespace>.svc.cluster.local.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,770 @@
...
+All O-RAN services will use the same O-CLOUD DB service. More on DB deployment [here](#postgres).
+
+Each table is modeled after O-RAN data structures. DB in our case wil be called `o-ran-infrastructure-monitoring-alarms`.
+Init SQL may look like the following:
+```sql
+CREATE DATABASE o-ran-infrastructure-monitoring-alarms;
+
+-- ENUM for ManagementInterfaceID
+DROP TYPE IF EXISTS ManagementInterfaceID CASCADE;
+CREATE TYPE ManagementInterfaceID AS ENUM ('O1', 'O2DMS', 'O2IMS', 'OpenFH');
```
alegacy, line 346: DB enums tend to be a bit difficult to deal with over schema upgrades if values need to be added.   Projects often use just regular integer columns and let the ORM map from enums in code to integers in DB.  This also has the side effect of consuming less space in the DB which is especially beneficial on large tables.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,770 @@
...
+            alarm_raised_time TIMESTAMPTZ NOT NULL,
+            alarm_changed_time TIMESTAMPTZ,
+            alarm_cleared_time TIMESTAMPTZ,
+            alarm_acknowledged_time TIMESTAMPTZ,
+            alarm_acknowledged BOOLEAN NOT NULL default FALSE,
+            perceived_severity PerceivedSeverity NOT NULL,
+            extensions JSONB,
+            created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+            finger_print TEXT not null,
+            alarm_sequence_number BIGINT DEFAULT nextval('alarm_sequence_seq'),
```
alegacy, line 459: I presume you intend on using this sequence number to control the order that these records are presented to a subscription, but I believe this is going to be problematic for events that may change states multiple times before we get to send it up to a subscriber.   If a record changes multiple times it may be possible that the subscriber will only see the last update.  To address this it may be necessary to maintain a separate table just to track the individual transitions of each alarm_event_record so that subscribers see the full picture.  

Maybe I've misread this, but this came to mind when I saw the trigger function below that updates the sequence value on an update.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,770 @@
...
+EXECUTE FUNCTION set_alarm_sequence_on_update();
+
+
+-- Table: alarm_event_record_archive, is identical to alarm_event_record. Use to this to eventually store events that are considered historical (i.e status: resolved)
+DROP TABLE IF EXISTS alarm_event_record_archive CASCADE;
+CREATE TABLE alarm_event_record_archive
+(LIKE alarm_event_record INCLUDING ALL);
+```
+
+Please note that script above was used test and need some updates for production e.g use `CREATE TABLE IF NOT EXISTS` and remove `DROP TABLE IF EXISTS`.
```
alegacy, line 496: "was used test" -> "was used to test"?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,770 @@
...
+   - POSTGRES_PASSWORD: `o-ran`
+   - POSTGRES_DB: `o-ran` # Note: this is simply there to be explicit.  
+                                   If not provided `POSTGRES_USER` is used to create default DB. But ultimately this DB not used as each service will create their own.
+- ConfigMap: For others to know the hostname and port
+   - POSTGRES_HOST: "postgres.o-ran-namespace.svc.cluster.local"
+   - POSTGRES_PORT: "5432"
+
+With the Secrets and ConfigMap applied, an app may connect to the DB with the following given the service (assuming initContainer of app already created DB `o-ran-infrastructure-monitoring-alarms`):
+```go
+connStr := "postgres://o-ran:o-ran@postgres.o-ran-namespace.svc.cluster.local:5432/o-ran-infrastructure-monitoring-alarms?sslmode=disable"
```
alegacy, line 730: We should try to enable TLS on all inter-pod communications.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: yep! updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Yeah good point about having issues with future migration. I'll go with the lookup table method instead of enum. Will update the doc. 

e.g 

```
CREATE TABLE management_interfaces (
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) UNIQUE NOT NULL
);

INSERT INTO management_interfaces (name) VALUES ('O1'), ('O2DMS'), ('O2IMS'), ('OpenFH');
```

```
CREATE TABLE alarms_table (
    ...
    management_interface_id INT REFERENCES management_interfaces(id) NOT NULL DEFAULT 3
);
```

alegacy: Yes, that would work, but is it really necessary?  ORM's are usually able to do that mapping in code without requiring a separate table... but either way works.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Yeah good point. I have deliberately left out the TLS stuff for now. Like alertmanager webhook also has a `tls_config` that we may configure as well? Added a note under "## Future Updates" to make sure we fix this after the init dev work

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
alegacy: ok, but could probably still do this in code simply enough with the extra table.  Anyway... details.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,794 @@
...
+3. Take the rules and build the corresponding alarmDefinition dictionary
+   - Link alarm dictionary to the resourceType
+   - alarmDictionary version will be that major.minor of cluster
+
+4. Assuming DB table already created, persist each alarmDictionary as a DB table
+
+5. Take all the PromRules and build the AlarmDefinition. See [here](#prometheusrule-to-alarmdefinition-mapping) for mapping. 
+   - A corresponding probablyCause row will be autometically added.
+
+6. Apply the required CR to activate the internal endpoint for alertmanager notification. See [here](#steps-internalv1caas-alertsalertmanager) for more.
```
edcdavid, line 563: #steps-internalv1caas-alertsalertmanager -> #steps-for-internalv1caas-alertsalertmanager)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,794 @@
...
+
+
+-- Table: alarm_subscription_info
+DROP TABLE IF EXISTS alarm_subscription_info CASCADE;
+CREATE TABLE alarm_subscription_info (
+             subscription_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+             consumer_subscription_id UUID,
+             filter TEXT,
+             callback TEXT NOT NULL,
+             largest_number_alarm_event_seen_so_far BIGINT NOT NULL DEFAULT 0,
```
alegacy, line 428: `largest_number_alarm_event_seen_so_far` is a bit wordy.  Perhaps `event_cursor`?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,794 @@
...
+}
+```
+
+## Schema
+All O-RAN services will use the same O-CLOUD DB service. More on DB deployment [here](#postgres).
+
+Each table is modeled after O-RAN data structures. DB in our case wil be called `o-ran-infrastructure-monitoring-alarms`.
+Init SQL may look like the following:
+```sql
+CREATE DATABASE o-ran-infrastructure-monitoring-alarms;
```
alegacy, line 343: Since we are not sharing this postgres instance there is no need for the database name to be so specific.  Perhaps "alarm-server"? or just "alarms".

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Updated!

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: Yep too wordy! Updated to `event_cursor`. (will add the comments for each col during dev to explain them better)

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
pixelsoccupied: yep simply `alarms` works. Updated

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+
+    | managed_cluster<br/>`from alerts`<br/> | resourceID  <br/>`same as managed_cluster`<br/> | resourceTypeID for caas<br/>`derived from (resourceID + ResourceKindLogical + ResourceClassCompute)`<br/> |
+    |----------------------------------------|-------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
+    | f90561e2-6420-4924-b081-f4f8eaf50618   | f90561e2-6420-4924-b081-f4f8eaf50618            | 4586f964-6c6f-407b-9b18-cb3c9a712ec4                                                                      |
+   
+    `entity_type` data should be coming from Inventory API but for now we can hard-code it. `telco-model-OpenShift-<Full Version>`
+- `alarm_definitions` reflects Rules in PromRule CR. We only grab the full set based on unique entries in `Versions` table
+    - Use ACM to get credentials of the unique major.minor clusters and retrieve all the PrometheusRules from them to parse. 
+      E.g if we are managing 3 clusters 4.16.2, 4.17.2 and 4.16.8, Pick 4.16.8 and 4.17.2 which effectively represents all the rules in 4.16.z and 4.17.z clusters.
```
mlguerrero12, line 618: There exist the possibility that the cluster 4.16.2 has some custom rules (added manually). These won't be added to the dictionary but alertmanager will still receive them. We could still create an alarm record and send a notification for these, there will be some empty fields. What do you think of this?

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+-- The data here mapped from PrometheusRule. See ##PrometheusRule to AlarmDefinition mapping section for more.
+
+-- probable_cause will be auto populated. see the section "##Get Probable cause ID, name and description" for more details
+```
+## Detailed Server Instantiation
+
+For a given OCP release, the alarmDefinitions and probableCauses are fixed, so these can be built up front. For CaaS alarms only one resource type, “NodeCluster”, all alarms map to it.
+
+1. Query all managed clusters to get list of unique major.minor versions
+   - Need to monitor for major.minor new versions
```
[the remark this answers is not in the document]
pixelsoccupied: Yeah that's a great point! Touched on this briefly few weeks ago but for now we can start with this approach for now and maybe move this code resource server eventually? Basically someone has to get the managedCluster CR and read the content....and in this case it's us for now.

But yeah I agree....ultimately Resource server should watch/get cluster resources and we (alarms) should simply be notified about the change and update our db as needed. 

And on the other hand resource server exposes endpoints that needs to return resourceType with its alarm dictionary right? Does that mean we need expose an additional internal GET endpoint for it to retrieve that (assuming we are not sharing the table)? probably a question for https://github.com/openshift-kni/oran-o2ims/pull/262 

cc @bartwensley @browsell @alegacy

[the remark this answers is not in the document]
mlguerrero12: No, I'm assuming we share the same table. What I meant was to have internal endpoints in the Alarm server so that the Resource server can notify operations on ResourceTypes. For instance, if a new ResourceType is added, then they could send us a Post (with all needed information) for us to determine whether we need to add a new dictionary for this new resource type. If so, then we inspect the managed cluster and get the rules to add a new dictionary. Same for Delete and Update. The idea is to modify the DB accordingly. They still need to go to the DB whenever a query on a resource type is performed. 

What I'm looking with this is to have a more dynamic way to update the dictionaries instead on relying only on restarts.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -690,17 +712,27 @@ WHERE probable_causes.probable_cause_id = 'f5ac4ac7-0ff3-40a2-b305-77313c28136a'
 ## K8s resources
 We will need few K8s resources that will be eventually applied by the Operator. 
 
+### Jobs to Initialize Alarm Server DB
+We need two Jobs that can help with DB
+
+- One job that creates a Database using `CREATE DATBASE` cmd
```
bartwensley, line 718: nit: spelling "DATBASE"

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+-- Table: versions, link alarm_dictionary and alarm_definitions using version.
+DROP TABLE IF EXISTS versions CASCADE;
+CREATE TABLE versions (
+         version_number VARCHAR(50) PRIMARY KEY
+);
+
+-- Table: alarm_dictionary
+DROP TABLE IF EXISTS alarm_dictionary CASCADE;
+CREATE TABLE alarm_dictionary (
+          resource_type_id UUID PRIMARY KEY,
```
alegacy, line 364: i realize that there should be only 1 tuple per resource type but we should still maintain a separate UUID for the "alarm_dictionary_id" to decouple those two tables.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+           alarm_description TEXT NOT NULL,
+           proposed_repair_actions TEXT NOT NULL,
+           alarm_dictionary_version VARCHAR(50) NOT NULL, -- Links alarm_dictionary and alarm_definitions
+           alarm_additional_fields JSONB,
+           alarm_change_type AlarmLastChangeType DEFAULT 'added' NOT NULL,
+           clearing_type ClearingType DEFAULT 'automatic' NOT NULL,
+           management_interface_id ManagementInterfaceID[] DEFAULT ARRAY['O2IMS']::ManagementInterfaceID[],
+           pk_notification_field TEXT[] DEFAULT ARRAY['alarm_definition_id']::TEXT[],
+           created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+           CONSTRAINT fk_alarm_definitions_version FOREIGN KEY (alarm_dictionary_version) REFERENCES versions(version_number),
```
alegacy, line 390: This method of linking these two tables seems a bit more complicated than necessary.  I would prefer to see a column in this table for `alarm_dictionary_id` and a FK constraint pointing directly to the other table since these are 1:N rather than M:N.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+           alarm_definition_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+           alarm_name VARCHAR(255) NOT NULL,
+           alarm_last_change VARCHAR(50) NOT NULL,
+           alarm_description TEXT NOT NULL,
+           proposed_repair_actions TEXT NOT NULL,
+           alarm_dictionary_version VARCHAR(50) NOT NULL, -- Links alarm_dictionary and alarm_definitions
+           alarm_additional_fields JSONB,
+           alarm_change_type AlarmLastChangeType DEFAULT 'added' NOT NULL,
+           clearing_type ClearingType DEFAULT 'automatic' NOT NULL,
+           management_interface_id ManagementInterfaceID[] DEFAULT ARRAY['O2IMS']::ManagementInterfaceID[],
```
alegacy, line 387: similar comment to above.  This will only ever be "O2IMS" so no need to store as array.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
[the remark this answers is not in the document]
alegacy: I've updated my [PR](https://github.com/openshift-kni/oran-o2ims/pull/262) with info related to our discussion with respect to the interactions between the two components.

docs/enhancements/infrastructure-monitoring-service-api/alarms.md
```diff
@@ -0,0 +1,797 @@
...
+
+4. Assuming DB table already created, persist each alarmDictionary as a DB table
+
+5. Take all the PromRules and build the AlarmDefinition. See [here](#prometheusrule-to-alarmdefinition-mapping) for mapping. 
+   - A corresponding probablyCause row will be autometically added.
+
+6. Apply the required CR to activate the internal endpoint for alertmanager notification. See [here](#steps-for-internalv1caas-alertsalertmanager) for more.
+
+7. The server should now be ready to take requests. 
+
```
mlguerrero12, line 565: I know it is implicit above, but I think we should clearly say somewhere that we will assume rules of all resources of a specific resource type are the same. This means we won't be able to see added rules to a specific resource, unless this one is selected for inspection (not deterministic).

Also, inspired from https://github.com/openshift-kni/oran-o2ims/pull/262, we could/should implement and sync mechanism for the alarm dictionary. This way we could overcome the limitation of not seeing newly added (or deleted) rules. In the documentation, we could say that once the system is up, rule modifications should be applied to all resources of a specific resource type and that the system will pick up these changes every X (i.e. 5 minutes) time.

PR #216, opened 2024-09-24
Fix make-run and the controller debug config

irinamihai:


.vscode/launch.json
```diff
@@ -147,7 +147,8 @@
                         "mode": "auto",
                         "program": "${workspaceFolder}",
                         "env": {
-                                "KUBECONFIG": "${env:KUBECONFIG}"
+                                "KUBECONFIG": "${env:KUBECONFIG}",
+                                "IMAGE": "quay.io/openshift-kni/oran-o2ims-operator:latest"
```
irinamihai, line 151: Not sure if I should have this as "{env:IMAGE}" or just leave it as this :thinking:

alegacy: Sort of depends on how people are using it.  If people are setting an IMAGE in their global environment and always using that then I suppose ${env:IMAGE} would be ok, but I would think that, unlike KUBECONFIG, overriding the IMAGE value more frequently would be likely so it should be left to specify in the IDE's config.   Leave it as-is...just my $0.02... as I always do a deploy and never run standalone (the further along we get in development the better it would be if everyone tested in-cluster rather than standalone).

PR #217, opened 2024-09-25
CNF-14539: Logic update for ztp-done on the ClusterRequest

irinamihai:
Description:
- instead of looking at the ztp-done label on the managedCluster (which is not accurate for the ORAN case due to the usage of enforce policies), make sure the Provisioned condition is Completed and that all the policies are compliant.
- the ZTP Not Done status is set only once the ClusterInstance provisioning starts


internal/controllers/clusterrequest_controller_test.go
```diff
@@ -3634,3 +3629,290 @@ var _ = Describe("hasPolicyConfigurationTimedOut", func() {
...
+		ciDefaultsCm     = "clusterinstance-defaults-v1"
+		hwTemplateCm     = "hwTemplate-v1"
+		policy           *policiesv1.Policy
+		ptDefaultsCm     = "policytemplate-defaults-v1"
+		clusterRequestCR *oranv1alpha1.ClusterRequest
+		req              reconcile.Request
+	)
+
+	BeforeEach(func() {
+		// Define the needed resources.
```
Missxiaoguo, line 3650: I am thinking could these tests be intergrated/injected into the e2e cases in the ClusterRequestReconcile? So we don't have to re-declare all the resources.

Missxiaoguo (commented):
> the ZTP Not Done status is set only once the ClusterInstance provisioning starts
Is this description wrong? From the code, ztp not done is set only once ClusterInstance provision has completed

irinamihai:
> > the ZTP Not Done status is set only once the ClusterInstance provisioning starts
> 
> Is this description wrong? From the code, ztp not done is set only once ClusterInstance provision has completed

I went too many times back and forth. I think I'll update the code to have `ZTP Not done` once the provisioning starts. Thanks for pointing it out!

donpenney (approved):

PR #218, opened 2024-09-25
CNF-14357: Add timeout for Cluster provisioning

Missxiaoguo:
- Add ClusterProvisionStartedAt to record the start timestamp of cluster provisioning.
- If cluster provisioning takes longer than the time configured in spec.timeout.clusterProvisioning, the ClusterProvisioned condition will be set to false with the reason TimedOut.
- Only check policy configuration when cluster provisioning has started and has neither failed nor timed out.
- Requeue every 5 minutes if cluster provisioning is still in progress or unknown.
- Change the requeue interval to 5 minutes for policy configuration as well.


internal/controllers/utils/conditions.go
```diff
@@ -84,3 +85,43 @@ func SetStatusCondition(existingConditions *[]metav1.Condition, conditionType Co
 		},
 	)
 }
+
+// IsClusterProvisionStarted checks if the cluster provision condition is present
```
bartwensley, line 89: Shouldn't this check that ClusterProvisionStartedAt is set?

internal/controllers/clusterrequest_controller.go
```diff
@@ -1135,6 +1140,9 @@ func (t *clusterRequestReconcilerTask) updateClusterProvisionStatus(ci *siteconf
 				metav1.ConditionUnknown,
 				"Waiting for cluster provisioning to start",
 			)
+			if t.object.Status.ClusterDetails.ClusterProvisionStartedAt.IsZero() {
+				t.object.Status.ClusterDetails.ClusterProvisionStartedAt = metav1.Now()
+			}
```
tliu2021, line 1145: Since this code is common to both cases, consider moving it outside the if-else clause.

internal/controllers/clusterrequest_controller.go
```diff
@@ -1143,6 +1151,26 @@ func (t *clusterRequestReconcilerTask) updateClusterProvisionStatus(ci *siteconf
 			ciProvisionedCondition.Status,
 			ciProvisionedCondition.Message,
 		)
+
+		if t.object.Status.ClusterDetails.ClusterProvisionStartedAt.IsZero() {
+			t.object.Status.ClusterDetails.ClusterProvisionStartedAt = metav1.Now()
+		}
```
tliu2021, line 1157: See comment above.

internal/controllers/clusterrequest_controller.go
Missxiaoguo: There is another nested if condition, where it's only be set when clusterInstance provision is not present but clusterInstanceProcessed is completed.

internal/controllers/clusterrequest_controller.go
```diff
@@ -236,25 +236,28 @@ func (t *clusterRequestReconcilerTask) run(ctx context.Context) (ctrl.Result, er
...
 	// Handle policy configuration only after the cluster provisioning
-	// has started.
-	crProvisionedCond := meta.FindStatusCondition(
-		t.object.Status.Conditions,
-		string(utils.CRconditionTypes.ClusterProvisioned))
-	if crProvisionedCond != nil {
+	// has started, and not failed or timedout (completed, in-progress or unknown)
+	if utils.IsClusterProvisionStarted(t.object) &&
+		!utils.IsClusterProvisionFailed(t.object) &&
+		!utils.IsClusterProvisionTimedOut(t.object) {
```
irinamihai, line 249: It might be better to have these checks in their own function, `isClusterReadyForConfig`, since they're used exactly the same below. We would avoid issues in case just one place is mistakenly updated.

internal/controllers/utils/conditions.go
```diff
@@ -84,3 +85,46 @@ func SetStatusCondition(existingConditions *[]metav1.Condition, conditionType Co
 		},
 	)
 }
+
+// IsStatusConditionPresent checks if the cluster provision condition is present
```
irinamihai, line 89: Comment doesn't match function name

internal/controllers/utils/conditions.go
```diff
@@ -84,3 +85,46 @@ func SetStatusCondition(existingConditions *[]metav1.Condition, conditionType Co
...
+		}
+	}
+	return false
+}
+
+// IsClusterProvisionTimedOutOrFailed checks if the cluster provision condition status is timedout or failed
+func IsClusterProvisionTimedOutOrFailed(cr *oranv1alpha1.ClusterRequest) bool {
+	condition := meta.FindStatusCondition(cr.Status.Conditions, (string(CRconditionTypes.ClusterProvisioned)))
+	if condition != nil {
+		if condition.Status == metav1.ConditionFalse &&
```
irinamihai, line 110: Here it should be enough to only check the reasons.

Missxiaoguo: Yeah, but I prefer to also add the status false check to ensure that's expected behaviour - timeout/failed is false

PR #220, opened 2024-09-26
CNF-14679: Adapt to single H/W manager plugin

tliu2021:
This update introduces the following changes:
- Added an environment variable HWMGR_PLUGIN_NAMESPACE to the deployment to specify the namespace for the single plugin.
- Introduced a new field hwMgrId in the NodePool spec to identify the plugin adapter.
- Added ObservedGeneration fields for both the cloud manager and the hardware manager plugin in the NodePool status.

Note: To continue using the test plugin, use the following commands:
make run HWMGR_PLUGIN_NAMESPACE=oran-hwmgr-plugin-test
make deploy HWMGR_PLUGIN_NAMESPACE=oran-hwmgr-plugin-test


Makefile
```diff
@@ -170,7 +173,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	$(KUSTOMIZE) build config/default | sed 's|\plugin-namespace-placeholder|$(HWMGR_PLUGIN_NAMESPACE)|g' | $(KUBECTL) apply -f -
```
donpenney, line 176: Is this a common method for doing such customizations? TALM sets a number of env variables by creating a patch.yaml that gets applied by the kustomization.yaml:
https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/Makefile#L268

internal/controllers/utils/utils.go
```diff
@@ -1018,3 +1023,38 @@ func GetEnvOrDefault(name, defaultValue string) string {
 	}
 	return value
 }
+
+// GetHwMgrPluginNS returns the value of environment variable HWMGR_PLUGIN_NAMESPACE
+func GetHwMgrPluginNS() string {
+	// Ensure that this code only runs once
+	once.Do(func() {
```
donpenney, line 1030: This `once.Do` is new to me... Very cool!

api/hardwaremanagement/v1alpha1/node_pools.go
```diff
@@ -42,6 +42,9 @@ type NodePoolSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Location Spec",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	LocationSpec `json:",inline"`
 
+	// HwMgrId is the identifier for the hardware manager plugin adaptor.
+	HwMgrId string `json:"hwMgrId" yaml:"hwMgrId"`
```
donpenney, line 46: This breaks the current plugin code, as it is adding a new *required* field. I think add `omitempty` here may avoid that. We can likely drop the `omitempty` later once the plugin is updated to know about the new field.

donpenney: Verified that adding `omitempty` allows the current plugin code to work, which is referencing the previous version of the CRD, which it causes a failure to update the CR without it, as it would be missing this new field.
```suggestion
	HwMgrId string `json:"hwMgrId,omitempty"`
```

Makefile
tliu2021: I am not sure if the TALM approach is the best option. I'll investigate further to see if there's a better way to handle this.

api/hardwaremanagement/v1alpha1/node_pools.go
tliu2021: Right, need to support the current plugin code.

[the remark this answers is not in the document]
tliu2021: The yaml field  is not needed and it will be removed.

Makefile
tliu2021: Kustomize in FluxCD (kustomize.toolkit.fluxcd.io) supports post-build variable substitution, but I don't think it's necessary to introduce that feature just for this. Instead, I'm using a ConfigMap to pass the environment variable to the deployment.

donpenney: It just seemed like TALM's method of creating a patch with env vars is a little more formal than doing a sed to modify the output of the kustomize build while piping to the apply

tliu2021: I agree, but I believe using a ConfigMap with Kustomize Replacements is also a solid solution for this use case.

api/hardwaremanagement/v1alpha1/node_pools.go
tliu2021: Updated.

Makefile
```diff
@@ -193,6 +198,30 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
...
+$(KUBECTL): $(LOCALBIN)
+	@if [ -n "$(SYSTEM_KUBECTL)" ]; then \
+		echo "Using system kubectl: $(SYSTEM_KUBECTL)"; \
+		ln -sf $(SYSTEM_KUBECTL) $(LOCALBIN)/kubectl; \
+	else \
+		echo "System kubectl not found, downloading to $(LOCALBIN)..."; \
+		if [ "$(OS)" = "darwin" ]; then \
+			curl -LO "https://dl.k8s.io/release/$$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/darwin/$(ARCH)/kubectl"; \
+		else \
+			curl -LO "https://dl.k8s.io/release/$$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/$(ARCH)/kubectl"; \
```
donpenney, line 218: We should probably use a fixed version here and just always use kubectl from LOCALBIN. The Makefile has the ENVTEST_K8S_VERSION version from the original operator-sdk setup, I believe, but the envtest bits have been removed. It might be a bit nicer codewise to use it, actually, as it will handle the arch for us. You can see an example in the plugin Makefile:
https://github.com/openshift-kni/oran-hwmgr-plugin/blob/main/Makefile#L120

It downloads the latest setup-envtest binary with the envtest target, which is then used to download kubectl and return its path.

For example:
```
dpenney ~/testing/envtest $ mkdir bin
dpenney ~/testing/envtest $ LOCALBIN=$PWD/bin
dpenney ~/testing/envtest $ GOBIN=$LOCALBIN go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240927101401-4381fa0aeee4
dpenney ~/testing/envtest $ find bin
bin
bin/setup-envtest
dpenney ~/testing/envtest $ bin/setup-envtest use 1.28.3 --bin-dir $LOCALBIN -p path
/home/dpenney/testing/envtest/bin/k8s/1.28.3-linux-amd64

dpenney ~/testing/envtest $ find bin
bin
bin/setup-envtest
bin/k8s
bin/k8s/1.28.3-linux-amd64
bin/k8s/1.28.3-linux-amd64/etcd
bin/k8s/1.28.3-linux-amd64/kube-apiserver
bin/k8s/1.28.3-linux-amd64/kubectl
```

Makefile
```diff
@@ -329,6 +360,7 @@ ci-job: deps-update generate fmt vet lint fmt test bundle-check
 
 .PHONY: clean
 clean:
+	@rm config/manager/env-config.yaml  # Clean up the temporary file
```
donpenney, line 363: nit: In TALM, the generated file is added to a .gitignore to avoid accidental inclusion in a commit. If generation were to fail or be interrupted, for example, the file would not get deleted and could inadvertently get committed. Mind you, the ci-job should catch such an issue:
https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/config/manager/related-images/.gitignore

tliu2021: This file is added to .gitignore.  The rm command is an unnecessary step and can be removed.

donpenney: Thanks Tao... I didn't notice the ,gitignore

Makefile
tliu2021: The Makefile is updated.

Makefile
```diff
@@ -193,6 +197,12 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.15.0
 
+.PHONY: kubectl
+kubectl: $(KUBECTL) ## Use envtest to download kubectl
+$(KUBECTL): $(LOCALBIN) $(ENVTEST)
+	KUBEBUILDER_ASSETS=$$($(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path); \
```
donpenney, line 203: This can be on one cmd, where KUBEBUILDER_ASSETS is an env var just for the command being executed, rather than two separate commands. But also, we could add a check here if KUBECTL exists, to only do this one. Although it can be enhanced to allow for updating the version. Either way, should also probably be a symlink rather than a cp

tliu2021: The KUBEBUILDER_ASSETS environment variable is scoped to the single shell session where it is set, so there is no risk of leakage between commands. A check to see if KUBECTL exists is added and cp has been changed to a symlink.

donpenney (approved):

PR #222, opened 2024-09-27
adding a new target to deploy controller and Metadata server and crea…

mmorency2021:
This is part of  ORAN - Project infrastructure - Automatically start services 
  https://issues.redhat.com/browse/MGMT-18997

-using Target deploy to install  operator and enable services 
-  Using clusterID to deploy meta  data server , user can edit it to match their cloudId


Makefile
```diff
@@ -21,7 +21,6 @@ ginkgo_flags:=
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= 4.16.0
-
```
donpenney, line 24: This blank line should not be removed

Makefile
```diff
@@ -82,7 +81,13 @@ CONTAINER_TOOL ?= docker
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# generating the backend token 
+TOKEN_BACKEND := $(shell oc create token -n open-cluster-management multicluster-operators --duration=24h)
+BACKEND_TOKEN_RS := $(shell oc create token oauth-apiserver-sa -n openshift-oauth-apiserver --duration=48h)
+BACKEND_URL := $(shell oc get route -n open-cluster-management search-api -o json | jq -r '"https://" + .spec.host')
+
 
+#CURRENTUSER    := $$(id -u)
```
donpenney, line 90: Should this be deleted?

Makefile
```diff
@@ -123,7 +128,7 @@ build: manifests generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: manifests generate fmt vet binary ## Run a controller from your host.
-	IMAGE=$(IMAGE_TAG_BASE):$(VERSION) ./oran-o2ims start controller-manager
+	./oran-o2ims start controller-manager
```
donpenney, line 131: Why is the IMAGE env var removed?

Makefile
```diff
@@ -171,11 +176,23 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
-
+	
+
```
donpenney, line 180: Remove the duplicate blank line

Makefile
```diff
@@ -171,11 +176,23 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
...
 
+.PHONY: deploy-service
+deploy-service: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	@sed -i '' -e 's/ingressHost:.*/ingressHost: $(INGRESS_HOST)/' config/services/metadata-server.yaml 
+	@sed -i '' -e 's/cloudId: .*/cloudId: $(CLOUD_ID)/' config/services/metadata-server.yaml
+	$(KUSTOMIZE) build config/deployment | $(KUBECTL) apply -f -
+
+.PHONY: undeploy-service
+undeploy-service: undeploy controller and deleting services from K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
```
donpenney, line 193: This looks like it's missing a #

Makefile
```diff
@@ -171,11 +176,23 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
...
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
-
+	
+         
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
+.PHONY: deploy-service
```
donpenney, line 185: Why a separate deployment? Shouldn't this be all part of the one operator deployment configuration, but presumably with additional containers?

mmorency2021: @donpenney  thanks for the comments. i was playing a little bit with the original files and some changes has been done  by mistake. fixing them all and also put one target to deploy the oeprator and the services.

Makefile
```diff
@@ -82,6 +82,10 @@ CONTAINER_TOOL ?= docker
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# setting  the backend token and the backend URL
```
alegacy, line 85: None of these settings should be required.  These can all be auto-detected by the operator at runtime.

Makefile
```diff
@@ -170,11 +174,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	@sed -i '' -e 's/ingressHost:.*/ingressHost: $(INGRESS_HOST)/' config/services/metadata-server.yaml
```
alegacy, line 177: It should be possible to auto-detect this at runtime rather than specifying it.

Makefile
```diff
@@ -170,11 +174,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
...
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	@sed -i '' -e 's/ingressHost:.*/ingressHost: $(INGRESS_HOST)/' config/services/metadata-server.yaml 
+	@sed -i '' -e 's/cloudId: .*/cloudId: $(CLOUD_ID)/' config/services/metadata-server.yaml
+	$(KUSTOMIZE) build config/services | $(KUBECTL) apply -f -
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/services | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
```
alegacy, line 183: IMHO since this is going to be our default way of deploying we should keep this as it was.

Makefile
```diff
@@ -170,11 +174,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+	@sed -i '' -e 's/ingressHost:.*/ingressHost: $(INGRESS_HOST)/' config/services/metadata-server.yaml 
+	@sed -i '' -e 's/cloudId: .*/cloudId: $(CLOUD_ID)/' config/services/metadata-server.yaml
```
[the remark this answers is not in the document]
mmorency2021: actually i consider that the User should provide the cloud_id. 
to deploy operator and service user should use below format 

make deploy  CLOUD_ID=c0332915-6bbb-4d8d-8628-0ab3cc2c7e5e INGRESS_HOST=o2ims.apps.hubcluster-1.hubcluster-1.lab.eng.cert.redhat.com

config/services/kustomization.yaml
alegacy, line 1: As per my comment in the makefile... IMHO, this should all be included in the existing default kustomization

config/services/ressource.yaml
alegacy, line 1: typo in the filename `ressource` -> `resource`

Makefile
alegacy, line 1: Unless I am misunderstanding what you are trying to do I would have thought that this services directory would only include a kustomization.yaml file and an inventory.yaml file.   The kustomatization.yaml in the default directory should then include this directory... all of the other files that are included here now don't appear to be necessary.

mmorency2021: make sense if we are deploying all services with the operator by default. i will adjust it. thanks

Makefile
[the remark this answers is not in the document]
alegacy: I see, but in a real production environment this would not be deployed in single operation.  First the operator would be deployed with a default InventoryCR to simply enable the services, and then later a user would have to edit that CR to add the global o-cloud id as well as other attributes (that are yet to be defined).

Makefile
mmorency2021: i will put  service manifest under /manager directory.

Makefile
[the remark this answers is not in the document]
mmorency2021: true. in production  ofcourse Operator get deployed then user enabled services. this is why i n the first place i created two targets. one to deploy the operator and another one to enable the services.
waiting on @bartwensley comments since he created the Jira. thanks @alegacy.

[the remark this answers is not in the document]
alegacy: Sorry, I guess I wasn't clear... the deploying of the inventory CR should be in two steps... first the default version of it that enables the services without any extra config (e.g., o-cloud id)... and then later the user will come along and edit it to add configuration details to it (i.e., oauth credentials, SMO address, etc...)

bartwensley:
> This is part of ORAN - Project infrastructure - Automatically start services https://issues.redhat.com/browse/MGMT-18997
> 
> * Adding new target called: deploy-service to deploy  IMS operator controller and metadata server. other services like ressource and deployment server will be added later.

I would like this to be done through the existing "make deploy" - deploying the operator should also start the services. We don't need a target to deploy the operator without starting the services.

Makefile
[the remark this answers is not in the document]
bartwensley: I agree with @alegacy here - we only want a single deploy target that will start the services without any additional configuration. The CLOUD_ID needs to be specified by editing the inventory CR - it should not be specified as part of the "make deploy".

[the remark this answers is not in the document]
mmorency2021: what about the ingressHost ? 
user will update it later also?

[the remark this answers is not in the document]
alegacy: Theoretically we should be able to derive that from the running system.   We could explore options to query it from the default ingresscontroller.  Doing something (in go code) that would be the equivalent to this:

```
oc get ingresscontrollers.operator.openshift.io -n openshift-ingress-operator default -o json |jq -j .spec.domain
```

Unless there is a better way @irinamihai

config/services/manager_config_patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
```
pixelsoccupied, line 10: I'm curious why we are not generating this at runtime from the operator and then applying it with client-go/controller-runtime? See an example https://github.com/operator-framework/operator-sdk/blob/master/testdata/go/v4/memcached-operator/internal/controller/memcached_controller.go#L306 

This we can capture all the tokens/hosts/etc and construct the required CRs at runtime and once deployed we can monitoring this at every requeue

From a dev/deploy we simply `make deploy` and everything is handled automatically. In other words, we deploy the operator and the operator deploys and manages everything else for us

Makefile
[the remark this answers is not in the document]
mmorency2021: using this , you can get the domain , but not the ingresshost. 
oc get ingresscontrollers.operator.openshift.io -n openshift-ingress-operator default -o json | jq .status.domain
"apps.hubcluster-1.hubcluster-1.lab.eng.cert.redhat.com"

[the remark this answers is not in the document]
mmorency2021: for the metadata server are you expecting the domain ? IngressHost is no tpart of the ingresscontrollers operator.

see below :

oc explain ingresscontrollers.operator.openshift.io.spec
GROUP:      operator.openshift.io
KIND:       IngressController
VERSION:    v1

FIELD: spec <Object>

DESCRIPTION:
    spec is the specification of the desired behavior of the IngressController.
    
FIELDS:
  clientTLS	<Object>
    clientTLS specifies settings for requesting and verifying client
    certificates, which can be used to enable mutual TLS for edge-terminated and
    reencrypt routes.

  defaultCertificate	<Object>
    defaultCertificate is a reference to a secret containing the default
    certificate served by the ingress controller. When Routes don't specify
    their own certificate, defaultCertificate is used. 
     The secret must contain the following keys and data: 
     tls.crt: certificate file contents tls.key: key file contents 
     If unset, a wildcard certificate is automatically generated and used. The
    certificate is valid for the ingress controller domain (and subdomains) and
    the generated certificate's CA will be automatically integrated with the
    cluster's trust store. 
     If a wildcard certificate is used and shared by multiple HTTP/2 enabled
    routes (which implies ALPN) then clients (i.e., notably browsers) are at
    liberty to reuse open connections. This means a client can reuse a
    connection to another route and that is likely to fail. This behaviour is
    generally known as connection coalescing. 
     The in-use certificate (whether generated or user-specified) will be
    automatically integrated with OpenShift's built-in OAuth server.

  domain	<string>
    domain is a DNS name serviced by the ingress controller and is used to
    configure multiple features: 
     * For the LoadBalancerService endpoint publishing strategy, domain is used
    to configure DNS records. See endpointPublishingStrategy. 
     * When using a generated default certificate, the certificate will be valid
    for domain and its subdomains. See defaultCertificate. 
     * The value is published to individual Route statuses so that end-users
    know where to target external DNS records. 
     domain must be unique among all IngressControllers, and cannot be updated. 
     If empty, defaults to ingress.config.openshift.io/cluster .spec.domain.

  endpointPublishingStrategy	<Object>
    endpointPublishingStrategy is used to publish the ingress controller
    endpoints to other networks, enable load balancer integrations, etc. 
     If unset, the default is based on
    infrastructure.config.openshift.io/cluster .status.platform: 
     AWS:          LoadBalancerService (with External scope) Azure:
    LoadBalancerService (with External scope) GCP:          LoadBalancerService
    (with External scope) IBMCloud:     LoadBalancerService (with External
    scope) AlibabaCloud: LoadBalancerService (with External scope) Libvirt:
    HostNetwork 
     Any other platform types (including None) default to HostNetwork. 
     endpointPublishingStrategy cannot be updated.

  httpCompression	<Object>
    httpCompression defines a policy for HTTP traffic compression. By default,
    there is no HTTP compression.

  httpEmptyRequestsPolicy	<string>
    httpEmptyRequestsPolicy describes how HTTP connections should be handled if
    the connection times out before a request is received. Allowed values for
    this field are "Respond" and "Ignore".  If the field is set to "Respond",
    the ingress controller sends an HTTP 400 or 408 response, logs the
    connection (if access logging is enabled), and counts the connection in the
    appropriate metrics.  If the field is set to "Ignore", the ingress
    controller closes the connection without sending a response, logging the
    connection, or incrementing metrics.  The default value is "Respond". 
     Typically, these connections come from load balancers' health probes or Web
    browsers' speculative connections ("preconnect") and can be safely ignored.
    However, these requests may also be caused by network errors, and so setting
    this field to "Ignore" may impede detection and diagnosis of problems.  In
    addition, these requests may be caused by port scans, in which case logging
    empty requests may aid in detecting intrusion attempts.

  httpErrorCodePages	<Object>
    httpErrorCodePages specifies a configmap with custom error pages. The
    administrator must create this configmap in the openshift-config namespace.
    This configmap should have keys in the format "error-page-<error
    code>.http", where <error code> is an HTTP error code. For example,
    "error-page-503.http" defines an error page for HTTP 503 responses.
    Currently only error pages for 503 and 404 responses can be customized. Each
    value in the configmap should be the full response, including HTTP headers.
    Eg-
    https://raw.githubusercontent.com/openshift/router/fadab45747a9b30cc3f0a4b41ad2871f95827a93/images/router/haproxy/conf/error-page-503.http
    If this field is empty, the ingress controller uses the default error pages.

  httpHeaders	<Object>
    httpHeaders defines policy for HTTP headers. 
     If this field is empty, the default values are used.

  logging	<Object>
    logging defines parameters for what should be logged where.  If this field
    is empty, operational logs are enabled but access logs are disabled.

  namespaceSelector	<Object>
    namespaceSelector is used to filter the set of namespaces serviced by the
    ingress controller. This is useful for implementing shards. 
     If unset, the default is no filtering.

  nodePlacement	<Object>
    nodePlacement enables explicit control over the scheduling of the ingress
    controller. 
     If unset, defaults are used. See NodePlacement for more details.

  replicas	<integer>
    replicas is the desired number of ingress controller replicas. If unset, the
    default depends on the value of the defaultPlacement field in the cluster
    config.openshift.io/v1/ingresses status. 
     The value of replicas is set based on the value of a chosen field in the
    Infrastructure CR. If defaultPlacement is set to ControlPlane, the chosen
    field will be controlPlaneTopology. If it is set to Workers the chosen field
    will be infrastructureTopology. Replicas will then be set to 1 or 2 based
    whether the chosen field's value is SingleReplica or HighlyAvailable,
    respectively. 
     These defaults are subject to change.

  routeAdmission	<Object>
    routeAdmission defines a policy for handling new route claims (for example,
    to allow or deny claims across namespaces). 
     If empty, defaults will be applied. See specific routeAdmission fields for
    details about their defaults.

  routeSelector	<Object>
    routeSelector is used to filter the set of Routes serviced by the ingress
    controller. This is useful for implementing shards. 
     If unset, the default is no filtering.

  tlsSecurityProfile	<Object>
    tlsSecurityProfile specifies settings for TLS connections for
    ingresscontrollers. 
     If unset, the default is based on the
    apiservers.config.openshift.io/cluster resource. 
     Note that when using the Old, Intermediate, and Modern profile types, the
    effective profile configuration is subject to change between releases. For
    example, given a specification to use the Intermediate profile deployed on
    release X.Y.Z, an upgrade to release X.Y.Z+1 may cause a new profile
    configuration to be applied to the ingress controller, resulting in a
    rollout.

  tuningOptions	<Object>
    tuningOptions defines parameters for adjusting the performance of ingress
    controller pods. All fields are optional and will use their respective
    defaults if not set. See specific tuningOptions fields for more details. 
     Setting fields within tuningOptions is generally not recommended. The
    default values are suitable for most configurations.

  unsupportedConfigOverrides	<Object>
    unsupportedConfigOverrides allows specifying unsupported configuration
    options.  Its use is unsupported.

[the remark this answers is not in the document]
alegacy: The ingress host is just the domain prepended with "apps.".  This is the ingress controller domain on my system:

`apps.hubcluster.hub.dev.vz.bos2.lab`

and this is my ingress host:

`o2ims.apps.hubcluster.hub.dev.vz.bos2.lab"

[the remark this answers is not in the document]
irinamihai: Here's on the cloudIds from the O-RAN standard:
```
oCloudId: Identifier for the containing O-Cloud. This identifier is copied from the O-Cloud Id assigned by the SMO during the O-Cloud deployment.

globalcloudId: Identifier of the O-Cloud instance. Globally unique across O-Cloud instances. This value was provided by the SMO with the callback URI at the beginning of O-Cloud genesis and used in registration at the end of genesis.
```

On the ingressHost: I agree with @alegacy that it can be derived, as he pointed out, from the ingresscontroller or from one of the existing routes. The user could then go ahead and edit it if desired.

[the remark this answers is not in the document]
mmorency2021: on our side we can deploy the operator with the service using the domain , then user can edit it after?  in my lab i just add a prefix to the domain to set the ingressHost. 
i will do the changes according to your comments.  thanks all.

donpenney:
Please note that this update impacts the generated manifests, which must also be included in the PR

config/manager/metadata-server.yaml
```diff
@@ -0,0 +1,24 @@
...
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/instance: orano2ims-sample
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: orano2ims
+    app.kubernetes.io/part-of: oran-o2ims
+  name: sample
+  namespace: oran-o2ims
+spec:
+  alarmSubscriptionServerConfig:
+    enabled: false
```
browsell, line 15: What does enabled mean in this context ?

mmorency2021: Enabled indicates if the alarm server   should be started or not.

mmorency2021:
i had to close this PR since they have changed some logic upstream. i created a target called deploy but  a target called deploy has been pushed to main branch.

PR #223, opened 2024-10-01
Add initial upgrade reconciler

sudomakeinstall2:


api/v1alpha1/clusterrequest_types.go
```diff
@@ -41,6 +42,8 @@ type ClusterRequestSpec struct {
 
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Timeout Timeout `json:"timeout,omitempty"`
+
+	SeedImageRef ibuv1.SeedImageRef `json:"seedImageRef,omitempty"`
```
browsell, line 46: This cannot be a top level parameter. It would have to be provided as a template parm, which raises the question do we even expose this of just make a fixed parm of the template version

bartwensley: Agreed - the seedImageRef will be specified as part of the cluster template and will be fixed for a particular cluster template version - we still need a bit of investigation to figure out the best way to do that.

config/crd/bases/ims.oran.openshift.io_clusterrequests.yaml
bartwensley, line 1: Why is this (and the other CRDs) showing up as a new file?

sudomakeinstall2: these are because of the recent group name change. I will remove them.

jc-rh:
Replaced by #264

PR #225, opened 2024-10-01
Requires permission to contact the /o2ims-infrastructureInventory/v1

alosadagrande:
Missing permission to contact the  /o2ims-infrastructureInventory/v1 API.

### DIFF
--- a/config/samples/testing/client-service-account-rbac.yaml
+++ b/config/samples/testing/client-service-account-rbac.yaml
@@ -10,6 +10,7 @@ metadata:
   name: oran-o2ims-test-client-role
 rules:
   - nonResourceURLs:
+      - /o2ims-infrastructureInventory/v1
       - /o2ims-infrastructureInventory/v1/*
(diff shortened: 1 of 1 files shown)


config/samples/testing/client-service-account-rbac.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
   name: oran-o2ims-test-client-role
 rules:
   - nonResourceURLs:
+      - /o2ims-infrastructureInventory/v1
```
alegacy, line 13: Did you try just removing the `/` and making it `- /o2ims-infrastructureInventory/v1`

alosadagrande: That would work, but then we allow any call starting with v1, and there should be nothing listening on that, at least nothing we want to interact with. see 2) 

- /o2ims-infrastructureInventory/v1
**- /o2ims-infrastructureInventory/v1whatever**
- /o2ims-infrastructureInventory/v1/whatever

PR #226, opened 2024-10-01
Upversion golang to 1.22

sudomakeinstall2:
In order to use IBGU for upgrade, TALM should be added as a dependency. Since TALM is using golang1.22, it is required to upversion golang to 1.22.


donpenney:
You have to also update https://github.com/openshift-kni/oran-o2ims/blob/main/Dockerfile.tools

donpenney:
I think https://github.com/openshift-kni/oran-o2ims/blob/main/api/hardwaremanagement/go.mod should probably be updated in this PR as well. In the near future, we should look at creating a check we can include in the ci-job that makes sure this subfolder go.mod aligns with the main one - although that may make it tricky when we add dependbot config

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift-kni/oran-o2ims
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.8
```
donpenney, line 6: Do we still need the k8s.io/api and k8s.io/client-go replace directives? These are go-1.20 modules:
https://github.com/kubernetes/api/blob/v0.28.2/go.mod
https://github.com/kubernetes/client-go/blob/v0.28.2/go.mod

I guess we need to check the siteconfig operator dependencies here. Maybe we can update or remove the related replace directive completely?

The latest siteconfig operator is go-1.21, but still uses the 0.28.2 api and client-go.

sudomakeinstall2: https://github.com/stolostron/siteconfig/pull/112

sudomakeinstall2: I created a PR to remove the directive from siteconfig. However TALM has one of these directives, and it needs quite a bit of work to remove. It needs complicated code changes.
Maybe if move our APIs to a separate repo we could avoid these issues.

donpenney: You could do as we've done in this repo, creating go submodules for the exported APIs. The oran-hwmgr-plugin is able to import just the api/hardwaremanagement folder, for example, without then pulling in all the dependencies from the main oran-o2ims go.mod

donpenney:
In addition to `api/hardwaremanagement`, we now have two additional exported API modules that need to be updated:
`api/inventory`
`api/provisioning`

donpenney (commented):
/lgtm

PR #227, opened 2024-10-01
CNF-14652: Update API group for Cluster Request&Template

irinamihai:
Description:
- update the API group to o2ims.provisioning.oran.org for ClusterRequest and ClusterTemplate
- move ClusterRequest and ClusterTemplate APIs to their own directory and go module
- move the Inventory API to its own directory and go module


irinamihai:
/hold
Still going through some end to end tests

go.mod
```diff
@@ -22,7 +22,9 @@ require (
 	github.com/json-iterator/go v1.1.12
 	github.com/onsi/ginkgo/v2 v2.14.0
 	github.com/onsi/gomega v1.30.0
-	github.com/openshift-kni/oran-o2ims/api/hardwaremanagement v0.0.0-00010101000000-000000000000
+	github.com/openshift-kni/oran-o2ims/api/hardwaremanagement v0.0.0-20241001130125-a052f08603f7
```
donpenney, line 25: Why is this version being changed?

internal/controllers/clustertemplate_controller.go
```diff
@@ -41,7 +41,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	hwv1alpha1 "github.com/openshift-kni/oran-o2ims/api/hardwaremanagement/v1alpha1"
-	oranv1alpha1 "github.com/openshift-kni/oran-o2ims/api/v1alpha1"
+	oranv1alpha1prov "github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1"
```
donpenney, line 44: `provv1alpha1` or `provisioningv1alpha1` would be more consistent with other imports, with the version as the suffix

internal/controllers/inventory_controller.go
```diff
@@ -30,7 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sptr "k8s.io/utils/ptr"
 
-	oranv1alpha1 "github.com/openshift-kni/oran-o2ims/api/v1alpha1"
+	oranv1alpha1inv "github.com/openshift-kni/oran-o2ims/api/inventory/v1alpha1"
```
donpenney, line 33: Same comment as above for import name format, with version as a suffix

go.mod
donpenney: I don't think it should have been. Plus, we have that performed by ci-job on each PR. The update includes upversioning multiple dependencies, in addition to the restructuring. Maybe you inadvertently also upgraded this module?

irinamihai: I noticed the hardware manager gets included as a required module with a version containing a string with today's date when I do `go mod tidy` for the provisioning package after the initial `go mod init` for provisioning. :thinking: 
Then after I do `go mod tidy` for the main oran-o2ims package, the main `go.mod` gets updated to the same hardware manager version.

donpenney: I think the only way this would have happened would be if you had run `go get github.com/openshift-kni/oran-o2ims/api/hardwaremanagement`. Ultimately, it doesn't really matter, since there's a `replace` directive. It's just an unnecessary change that sticks out and made me curious

internal/controllers/inventory_controller_test.go
```diff
@@ -33,7 +33,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	oranv1alpha1 "github.com/openshift-kni/oran-o2ims/api/v1alpha1"
+	oranv1alpha1inv "github.com/openshift-kni/oran-o2ims/api/inventory/v1alpha1"
```
donpenney, line 36: Missed a rename here

irinamihai: I will run another end to end install today to validate the renaming. Thank you!

irinamihai:
/hold
End to end test with latest changes.

donpenney (commented):
/lgtm

irinamihai:
/unhold
Performed end to end SNO install + inventory run of the metadata and deployment manager server.

bartwensley:
I missed this before it was merged, but it is quite possible that the ClusterTemplate will not become part of the O-RAN spec, so it may not belong under the o2ims.provisioning.oran.org API. I guess we can change that later if it turns out to be the case?

irinamihai:
> I missed this before it was merged, but it is quite possible that the ClusterTemplate will not become part of the O-RAN spec, so it may not belong under the o2ims.provisioning.oran.org API. I guess we can change that later if it turns out to be the case?

@bartwensley , I put them together for now because they are both part of provisioning. We can definitely update it at any time. As long as it's not part of the standard though, it can belong to any group we choose, isn't it?

PR #229, opened 2024-10-02
CNF-13703: Ignore ClusterInstance immutable fields for vars from NodePool

Missxiaoguo:
For the fields (bmcAddress, bmcCredentialsName, bootMACAddress, and nodeNetwork.interfaces[*].macAddress) that come from NodePool, we should just ignore changes to those fields.

### DIFF
--- a/internal/controllers/utils/constants.go
+++ b/internal/controllers/utils/constants.go
@@ -78,15 +78,25 @@ const (
 )
 
 var (
-	// Non-immutable ClusterInstance fields at the cluster-level
-	AllowedClusterInstanceClusterFields = []string{
-		"extraAnnotations",
-		"extraLabels",
+	// AllowedClusterInstanceFields contains path patterns for fields that are allowed to be updated.
+	// The wildcard "*" is used to match any index in a list.
+	AllowedClusterInstanceFields = [][]string{
+		// Cluster-level non-immutable fields
+		{"extraAnnotations"},
+		{"extraLabels"},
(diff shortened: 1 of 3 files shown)


internal/controllers/utils/constants.go
```diff
@@ -78,15 +78,23 @@ const (
...
-		"extraAnnotations",
-		"extraLabels",
+
+	// IgnoredClusterInstanceFields contains path patterns for fields that should be ignored.
+	IgnoredClusterInstanceFields = [][]string{
+		// Node-level ignored fields
+		{"nodes", "*", "bmcAddress"},
+		{"nodes", "*", "bmcCredentialsName"},
+		{"nodes", "*", "bootMACAddress"},
+		{"nodes", "*", "nodeNetwork", "interfaces", "*", "macAddress"},
```
bartwensley, line 97: Why is this listing specific fields but also including a wildcard?

donpenney: I think the wildcard here is an index into a list of X ... ie. to match node[0].bmcAddress, for example. Probably would be good to comment that here?

Missxiaoguo: Updated with a comment. The wildcard in that specific position represents the index in a list.

Missxiaoguo: > I think the wildcard here is an index into a list of X ... ie. to match node[0].bmcAddress, for example. Probably would be good to comment that here?

Yep, added a comment

PR #231, opened 2024-10-03
NO-ISSUE: Introduce script to sync submodule dependencies

donpenney:
This update adds a script to sync the dependencies of exported submodules against the main go.mod.

Additionally, shellcheck and bashate are added to ci-job.


donpenney:
This is being dropped in favour of using go workspaces after the upgrade to 1.22

PR #234, opened 2024-10-04
NO-ISSUE: Add shellcheck and bashate linters

donpenney:


hack/shellcheck.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+VERSION="0.7.2"
```
irinamihai, line 3: Should we hardcode this here or have it as a variable in the Makefile and pass it to the script?

donpenney: It just needs to be consistent, locking the version. Having it here means a user can just run the script directly, as well. But also, this is just a clone of the same script we've got in our other repos.

PR #236, opened 2024-10-04
CNF-14648: Fix errors during TemplateID update

edcdavid:
This PR fixes the following:
- race condition after updating the templateId field, the version is not updated in time for the next update to succeed. updating the templateId field first if needed and requeue to continue with validation. Wait a bit for the cache to updated with the new object version. Thanks @Missxiaoguo for the solution
- uuid passed by the user was not properly updated and always regenerated.


internal/controllers/clustertemplate_controller.go
```diff
@@ -75,6 +75,10 @@ func requeueWithMediumInterval() ctrl.Result {
 	return requeueWithCustomInterval(1 * time.Minute)
 }
 
+func requeueImmediatly() ctrl.Result {
+	return requeueWithCustomInterval(0)
+}
```
Missxiaoguo, line 80:
```suggestion
func requeueImmediately() ctrl.Result {
	return ctrl.Result{Requeue: true}
}
```

internal/controllers/clustertemplate_controller.go
```diff
@@ -98,6 +102,7 @@ func (r *ClusterTemplateReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	_ = log.FromContext(ctx)
 	result = doNotRequeue()
 
+	time.Sleep(100 * time.Millisecond)
```
Missxiaoguo, line 105: Could we add a brief description here for this pause?

internal/controllers/clustertemplate_controller.go
```diff
@@ -284,15 +297,22 @@ func validateName(c client.Client, name, version, metadataName, namespace string
...
 			return utils.NewInputError("failed to validate templateID, invalid UUID:%s", object.Spec.TemplateID)
 		}
-		return nil
 	}
+	return nil
+}
 
+// generateTemplateID generates a new templateId if it was not present
+// If the templateID does not exist, it is generated by this function using uuid
+func generateTemplateID(ctx context.Context, c client.Client, object *provisioningv1alpha1.ClusterTemplate) error {
```
Missxiaoguo, line 312: I think c.Get(line324-327) in this function could be removed

Missxiaoguo (commented):
Perhaps we can still use CNF-14648 for this PR?

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
```diff
@@ -75,6 +75,10 @@ func requeueWithMediumInterval() ctrl.Result {
 	return requeueWithCustomInterval(1 * time.Minute)
 }
 
+func requeueImmediatly() ctrl.Result {
```
Missxiaoguo, line 78: nit: spelling `immediately`

edcdavid: Fixed

config/samples/v1alpha1_clustertemplate.yaml
```diff
@@ -1,17 +1,20 @@
...
-    app.kubernetes.io/instance: clustertemplate-sample
-    app.kubernetes.io/part-of: oran-o2ims
-    app.kubernetes.io/managed-by: kustomize
-    app.kubernetes.io/created-by: oran-o2ims
   name: clustertemplate-sample.v1.0.0
+  namespace: clustertemplate-a-v4-16
 spec:
+  name: clustertemplate-a 
+  description:   nice description
+  templateId: 57b39bda-ac56-4143-9b10-d1a71517d04f
```
Missxiaoguo, line 9: Probably we should remove templateId since it's auto-generated

edcdavid: Fixed

PR #239, opened 2024-10-04
Proposal: Add DLV debugging capabilities

alegacy:
This adds the changes required so that a developer can build an image capable of using the DLV debugger to debug any of the deployed servers or controller.

Some functionality needed to be disabled in order to make suspending the program being debugged possible (i.e., health checks,
leader election, etc).  This is only true if DEBUG=yes when the build command is executed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
 # Build the manager binary
-FROM registry.hub.docker.com/library/golang:1.22 as builder
+FROM registry.hub.docker.com/library/golang:1.22 as dlvbuilder
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+FROM dlvbuilder as builder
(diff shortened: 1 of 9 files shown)


config/manager/kustomization.yaml
```diff
@@ -10,8 +10,8 @@ generatorOptions:
 
 images:
 - name: controller
-  newName: quay.io/openshift-kni/oran-o2ims-operator
-  newTag: 4.16.0
+  newName: quay.io/alegacy/oran-o2ims
```
irinamihai, line 13: I don't think this needs to change.

alegacy: nope!  forgot to remove it before committing.

alegacy:
@donpenney  could you take another look here.  No changes, just a rebase.

PR #240, opened 2024-10-04
CNF-14779: Add support for NodePool configuration changes

tliu2021:
[CNF-14779](https://issues.redhat.com//browse/CNF-14779): Add support for NodePool configuration changes

This update introduces support for applying NodePool configuration changes triggered by the hardware template updates. These configuration changes are only processed after the cluster has been fully provisioned. The new template is validated against the existing NodePool configuration, and any changes to initial configuration parameters—such as node groups, hwMgrId, or bootInterfaceLabel—will result in a template rendering failure.

Additionally, a new status condition has been added to the NodePool to reflect the updated configuration, and the provisioning request status is updated accordingly.


api/hardwaremanagement/v1alpha1/node_pools.go
```diff
@@ -62,7 +62,8 @@ type Properties struct {
 
 // GenerationStatus represents the observed generation for an operator.
 type GenerationStatus struct {
-	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	ObservedGeneration int64  `json:"observedGeneration,omitempty"`
+	SpecHash           string `json:"specHash,omitempty"`
```
alegacy, line 66: Why do we need both a generation and a hash?  Doesn't the generation change every time an update is applied?

tliu2021: The SpecHash and ObservedGeneration each serve a different purpose. The SpecHash field captures the hash of the spec changes triggered by the hardware template and enables the operator to quickly detect when updates to the NodePool are required. It also differentiates the changes made by the plugin.  The ObservedGeneration field allows each operator to confirm that it has processed the most recent version of the spec. This is essential when multiple operators are involved, as each can independently confirm it has reconciled the latest changes. The ObservedGeneration is updated after a successful spec update.

tliu2021: Based on my discussion with Don, the plugin will not update the NodePool spec, and I will modify this PR accordingly.

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -5,15 +5,72 @@ import (
...
+			}
+			t.logger.InfoContext(
+				ctx,
+				fmt.Sprintf(
+					"NodePool %s in the namespace %s configuration changes have been detected",
+					nodePool.GetName(),
+					nodePool.GetNamespace(),
+				),
+			)
+		} else if utils.IsClusterProvisionPresent(t.object) && !utils.IsClusterProvisionTimedOutOrFailed(t.object) { // in-progress or unknown
```
Missxiaoguo, line 59: I wonder could we just simply reject the hw profile changes if process has started and set `HardwareTemplateRendered` to failed with an error message?
Currently,  once clusterinstance starts installing, any changes to its clusterinstance immutable fields are disallowed and will cause `ClusterInstanceRendered` to fail.

Missxiaoguo: I feel this is a bit of complicated. Do we have a use case where the profile would need updating on Day 0 before provisioning is complete? Why couldn’t they just delete the existing ProvisioningRequest and start with a new profile if the current one isn’t correct?

tliu2021: It is definitely an unlikely scenario, so I'll update the code to reject it.

internal/controllers/utils/hardware_utils.go
```diff
@@ -324,3 +344,106 @@ OuterLoop:
...
+
+	// Update the Kubernetes Custom Resource status and handle any errors
+	if err := UpdateK8sCRStatus(ctx, client, nodePool); err != nil {
+		return fmt.Errorf("failed to update status for NodePool %s in namespace %s: %w", nodePool.GetName(), nodePool.GetNamespace(), err)
+	}
+	return nil
+}
+
+// IsNodePoolProvisioned checks if the cluster provision condition status is completed
+func IsNodePoolProvisioned(nodePool hwv1alpha1.NodePool) bool {
```
Missxiaoguo, line 439: Maybe move this one to conditions.go for better consistency?

internal/controllers/provisioningrequest_hwprovision.go
tliu2021: Changed.

api/hardwaremanagement/v1alpha1/node_pools.go
tliu2021: PR is updated.

internal/controllers/utils/hardware_utils.go
```diff
@@ -324,3 +348,147 @@ OuterLoop:
...
+		}
+		if !found {
+			return true, fmt.Errorf("node group %s found in NodePool spec but not in ConfigMap", specNodeGroup.Name)
+		}
+	}
+	return changesDetected, nil
+}
+
+// UdateNodePoolStatus updates the NodePool status fields
+func UdateNodePoolStatus(ctx context.Context, client client.Client, nodePool *hwv1alpha1.NodePool,
```
Missxiaoguo, line 466: typo: updateNodePoolStatus

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -289,6 +383,28 @@ func (t *provisioningRequestReconcilerTask) handleRenderHardwareTemplate(ctx con
...
+		// Verify the changes
+		changed, err := utils.CompareConfigMapWithNodePool(hwTemplateCm, nodePool, nodeGroup)
+		if err != nil {
+			return nodePool, utils.NewInputError("%w", err)
+		}
+		if changed && !utils.IsProvisioningStateFulfilled(t.object) {
+			return nodePool, utils.NewInputError("hardware template changes are not allowed until the cluster provisioning is fulfilled")
+		}
+	} else {
+		if err = utils.ValidateConfigMapFields(hwTemplateCm); err != nil {
```
Missxiaoguo, line 403: This validates the required fields in the hardware template ConfigMap prepared by CSP. It makes sense to include this as part of the cluster template validation, allowing CSP to address issues earlier

internal/controllers/utils/hardware_utils.go
```diff
@@ -324,3 +348,147 @@ OuterLoop:
...
+		}
+	}
+	return nil
+}
+
+// CompareConfigMapWithNodePool checks if there are any changes in the hardware template config map
+func CompareConfigMapWithNodePool(configMap *corev1.ConfigMap, nodePool *hwv1alpha1.NodePool, nodeGroup []hwv1alpha1.NodeGroup) (bool, error) {
+
+	// First, validate required fields in the ConfigMap
+	if err := ValidateConfigMapFields(configMap); err != nil {
```
Missxiaoguo, line 424: Can be moved to clustertemplate validation

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -5,15 +5,64 @@ import (
...
+	}
+
+	if !exist {
+		return t.createNodePoolResources(ctx, nodePool)
+	}
+
+	// The template validate is already completed; compare NodeGroup and update them if necessary
+	if !equality.Semantic.DeepEqual(existingNodePool.Spec.NodeGroup, nodePool.Spec.NodeGroup) {
+		// Only process the configuration changes if the cluster provisioning is fulfilled
+		if utils.IsProvisioningStateFulfilled(t.object) {
```
Missxiaoguo, line 35: These checks seem redundant since disallowed updates before fulfilled have been validated and rejected in the `CompareConfigMapWithNodePool` within `handleRenderHardwareTemplate`, so renderedHardwareTemplate condition will be failed and this function should not be reached.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -280,6 +289,31 @@ func (t *provisioningRequestReconcilerTask) run(ctx context.Context) (ctrl.Resul
 
 	}
 
+	// Check if NodePool configured status is set after the cluster provisioning status is fulfilled
```
Missxiaoguo, line 292: I just realized since upgrade is supported now(line 277-286), we also need to take that into account. It seems like checkNodePoolConfigStatus should be moved above which prevent further updates if nodePool config is still in progress? 🤔

tliu2021: It has been moved up.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -199,12 +199,21 @@ func (t *provisioningRequestReconcilerTask) run(ctx context.Context) (ctrl.Resul
...
 	if err != nil {
+		if utils.IsInputError(err) {
+			_, err = t.checkClusterDeployConfigState(ctx)
+			if err != nil {
+				return requeueWithError(err)
+			}
+			// Requeue since we are not watching for updates to required resources
+			// if they are missing
+			return requeueWithMediumInterval(), nil
+		}
```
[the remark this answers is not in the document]
tliu2021: Removed.

internal/controllers/provisioningrequest_hwprovision.go
tliu2021: Removed.

internal/controllers/utils/hardware_utils.go
tliu2021: It has been relocated to be part of the cluster template validation.

internal/controllers/utils/hardware_utils.go
tliu2021: Changed.

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -77,23 +122,20 @@ func (t *provisioningRequestReconcilerTask) createNodePoolResources(ctx context.
...
+	if err != nil {
+		return provisioned, nil, timedOutOrFailed, err
 	}
-	return provisioned, timedOutOrFailed, err
+	if provisioned {
+		configured, timedOutOrFailed, err = t.checkNodePoolConfigStatus(ctx, nodePool)
+	} else {
+		// If the node pool is not provisioned, set configured to nil
+		configured = nil
+	}
```
Missxiaoguo, line 137: nit: not needed since `configured` is a pointer, its value is nil if not set in the checkNodePoolConfigStatus.

Missxiaoguo (commented):
/lgtm

internal/controllers/provisioningrequest_hwprovision.go
tliu2021: Yeah, will remove it in the next update.

Missxiaoguo (commented):
/lgtm

edcdavid:
Tested IBI deployment to deploy SNO successfully. Then updated provisioning request with new template including new hwtemplate. Changed only hwprofile and manually moved the nodepool request to "ConfigApplied" triggering a provisioning request to move to fulfilled state. Also tried to apply the updated hw template when the provisioning request was not fulfilled and the HardwareTemplateRendered showed "hardware template changes are not allowed until the cluster provisioning is fulfilled"

PR #242, opened 2024-10-08
Adding validation step for missing required ClusterTemplate's cluster…

edcdavid:
Adding validation step for missing required ClusterTemplate's clusterInstanceParameters fields:
- nodeClusterName
- oCloudSiteId
- policyTemplateParameters
- clusterInstanceParameters
Testing the for the presence of the required schema keys in the properties and in the required sections

### DIFF
--- a/internal/controllers/clustertemplate_controller.go
+++ b/internal/controllers/clustertemplate_controller.go
@@ -350,22 +350,38 @@ func generateTemplateID(ctx context.Context, c client.Client, object *provisioni
 // - policyTemplateParameters
 // - clusterInstanceParameters
 func validateTemplateParameterSchema(object *provisioningv1alpha1.ClusterTemplate) error {
-	mandatoryParams := []string{utils.TemplateParamNodeClusterName,
-		utils.TemplateParamOCloudSiteId,
-		utils.TemplateParamPolicyConfig,
-		utils.TemplateParamClusterInstance}
+	const (
+		typeString   = "type"
+		stringString = "string"
+		objectString = "object"
+	)
+	mandatoryParams := [][]string{{utils.TemplateParamNodeClusterName, stringString},
+		{utils.TemplateParamOCloudSiteId, stringString},
+		{utils.TemplateParamPolicyConfig, objectString},
+		{utils.TemplateParamClusterInstance, objectString}}
 	if object.Spec.TemplateParameterSchema.Size() == 0 {
 		return utils.NewInputError("templateParameterSchema is present but empty:")
 	}
 	var missingParameter []string
+	var badType []string
 	for _, param := range mandatoryParams {
-		_, err := utils.ExtractSubSchema(object.Spec.TemplateParameterSchema.Raw, param)
+		expectedName := param[0]
+		expectedType := param[1]
+		aSubschema, err := utils.ExtractSubSchema(object.Spec.TemplateParameterSchema.Raw, expectedName)
 		if err != nil {
-			if strings.HasPrefix(err.Error(), fmt.Sprintf("subSchema %s does not exist:", param)) {
-				missingParameter = append(missingParameter, param)
+			if strings.HasPrefix(err.Error(), fmt.Sprintf("subSchema %s does not exist:", expectedName)) {
+				missingParameter = append(missingParameter, expectedName)
+				continue
 			} else {
-				return fmt.Errorf("error extracting subschema at key %s: %w", param, err)
+				return fmt.Errorf("error extracting subschema at key %s: %w", expectedName, err)
 			}
+		}
+		if aType, ok := aSubschema[typeString]; ok {
+			if aType != expectedType {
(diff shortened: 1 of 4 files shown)


internal/controllers/clustertemplate_controller.go
```diff
@@ -327,6 +334,54 @@ func generateTemplateID(ctx context.Context, c client.Client, object *provisioni
...
 }
 
+// validateTemplateParameterSchema return true if the schema contained in the templateParameterSchema
+// field contains the required mandatory parameters
+// - nodeClusterName
+// - oCloudSiteId
+// - policyTemplateParameters
+// - clusterInstanceParameters
+func validateTemplateParameterSchema(object *provisioningv1alpha1.ClusterTemplate) error {
+	mandatoryParams := []string{"nodeClusterName",
```
irinamihai, line 344: All these values will probably be standardized, so they can be made constants.

internal/controllers/clustertemplate_controller.go
```diff
@@ -327,6 +334,54 @@ func generateTemplateID(ctx context.Context, c client.Client, object *provisioni
...
+		"oCloudSiteId",
+		"policyTemplateParameters",
+		"clusterInstanceParameters"}
+	if object.Spec.TemplateParameterSchema.Size() == 0 {
+		return utils.NewInputError("templateParameterSchema is present but empty:")
+	}
+	var missingParameter []string
+	for _, param := range mandatoryParams {
+		_, err := utils.ExtractSubSchema(object.Spec.TemplateParameterSchema.Raw, param)
+		if err != nil && strings.HasPrefix(err.Error(), fmt.Sprintf("subSchema %s does not exist:", param)) {
```
irinamihai, line 354: Maybe refactor these lines to smth like:
```
if err != nil {
    if strings.HasPrefix(err.Error(), fmt.Sprintf("subSchema %s does not exist:", param)) {
        missingParameter = append(missingParameter, param)
    } else {
        return fmt.Errorf("error extracting subschema at key %s: %w", param, err)
    }
}
```

internal/controllers/clustertemplate_controller.go
```diff
@@ -327,6 +334,54 @@ func generateTemplateID(ctx context.Context, c client.Client, object *provisioni
...
+	requiredList, err := utils.ExtractSchemaRequired(object.Spec.TemplateParameterSchema.Raw)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling required subschema: %w", err)
+	}
+	for _, param := range mandatoryParams {
+		if !slices.Contains(requiredList, param) {
+			missingRequired = append(missingRequired, param)
+		}
+	}
+	validationFailureReason := fmt.Sprintf("failed to validate ClusterTemplate name:%s. ", object.Name)
```
irinamihai, line 370: I think we can remove `name` from this

internal/controllers/clustertemplate_controller.go
```diff
@@ -327,6 +334,54 @@ func generateTemplateID(ctx context.Context, c client.Client, object *provisioni
...
+	if len(missingParameter) != 0 {
+		validationFailureReason = fmt.Sprintf(" The following mandatory fields are missing: %s", strings.Join(missingParameter, ","))
+	}
+	if len(missingRequired) != 0 {
+		if len(missingParameter) != 0 {
```
irinamihai, line 375: If you end the reason above with a `.` ()line 372 and start the one below with a `The`, you don't need this check.

internal/controllers/clustertemplate_controller.go
Missxiaoguo: There are existing constants for these values.

Missxiaoguo (commented):
Could we also add the validation for the types? for example, 
`nodeClusterName` and `oCloudSiteId` should be of "type: string" and `policyTemplateParameters` and `clusterInstanceParameters` should be of "type: object". That's the key for validating the input types against the schema.

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed

internal/controllers/clustertemplate_controller.go
edcdavid: Fixed.

edcdavid:
> Could we also add the validation for the types? for example, `nodeClusterName` and `oCloudSiteId` should be of "type: string" and `policyTemplateParameters` and `clusterInstanceParameters` should be of "type: object". That's the key for validating the input types against the schema.

Fixed. Also added unit tests.

Missxiaoguo (commented):
/lgtm

PR #238, opened 2024-10-04
CNF-14294: Add registration to SMO

alegacy:
This adds a registration step to the reconcile loop whenever the Inventory CR contains the "smo" configuration attribute.  If that same attribute contains the "oauth" attribute then the registration will be done using an OAuth2 authentication token in each of the HTTP requests sent to the SMO.

PR #245, opened 2024-10-08
CNF-14660: Remove timeout from ProvisioningRequest spec

Missxiaoguo:
- Removed the timeout field from the ProvisioningRequest spec.
- Timeouts for hardware provisioning, cluster provisioning, and configuration are now configured in their respective ConfigMaps, referenced in ClusterTemplate.spec.templates. These timeouts are optional.
       - Config keys `hardwareProvisioningTimeout`, `clusterProvisioningTimeout`, `clusterConfigurationTimeout`
- If a timeout is not specified in the ConfigMap, the default timeout values will be used, consistent with previous behavior:
       - Default for hardware provisioning: 90 minutes
       - Default for cluster provisioning: 90 minutes
       - Default for cluster configuration: 30 minutes
- Added sample ConfigMaps for clusterInstance and policyTemplate.


internal/controllers/utils/utils.go
```diff
@@ -602,6 +601,20 @@ func ExtractTemplateDataFromConfigMap[T any](ctx context.Context, c client.Clien
 	return validData, nil
 }
 
+// ExtractTimeoutConfigFromConfigMap extracts the timeout config from the ConfigMap by key if exits,
+// converting it from integer to time.Duration. Returns an error if the value is not a valid integer.
+func ExtractTimeoutConfigFromConfigMap(cm *corev1.ConfigMap, key string) (time.Duration, error) {
```
irinamihai, line 606: I think this function can be removed and instead use GetConfigMapField (https://github.com/openshift-kni/oran-o2ims/pull/238/files#diff-ff9c80a7f6ade29a53eec0fe48e61578a5c73fac596e79b467b386fa3cc8873bR629) from Allain's PR or just keep this function, but call `GetConfigMapField` and convert to int.

It should still be renamed to something more similar, like `GetConfigMapTimeout` / `GetTimeoutFromConfigMap`

internal/controllers/utils/utils.go
```diff
@@ -1112,10 +1125,8 @@ func ClusterIsReadyForPolicyConfig(
 }
 
 // TimeoutExceeded returns true if it's been more time than the timeout configuration.
-func TimeoutExceeded(provisioningRequest *provisioningv1alpha1.ProvisioningRequest) bool {
-	timeSince := time.Since(provisioningRequest.Status.ClusterDetails.NonCompliantAt.Time)
-	timeout := time.Duration(provisioningRequest.Spec.Timeout.Configuration) * time.Minute
-	return timeSince > timeout
+func TimeoutExceeded(startTime time.Time, timeout time.Duration) bool {
```
irinamihai, line 1128: I think it would be more clear to leave this function as it is and pass a timeout in minutes.
If not, every time this function is used, the caller will have to transform the timeout beforehand.

internal/controllers/utils/utils.go
```diff
@@ -602,6 +601,20 @@ func ExtractTemplateDataFromConfigMap[T any](ctx context.Context, c client.Clien
...
 
+// ExtractTimeoutConfigFromConfigMap extracts the timeout config from the ConfigMap by key if exits,
+// converting it from integer to time.Duration. Returns an error if the value is not a valid integer.
+func ExtractTimeoutConfigFromConfigMap(cm *corev1.ConfigMap, key string) (time.Duration, error) {
+	if timeoutStr, exists := cm.Data[key]; exists {
+		timeoutInt, err := strconv.Atoi(timeoutStr)
+		if err != nil {
+			return 0, NewInputError("the value of key %s from ConfigMap %s is not an integer: %v", key, cm.GetName(), err)
+		}
+		return time.Duration(timeoutInt) * time.Minute, nil
```
irinamihai, line 612: In my opinion, we should just return `timeoutInt` here, no conversion, and leave it to the `TimeoutExceeded` function.

internal/controllers/utils/constants.go
```diff
@@ -68,6 +72,20 @@ const (
...
 )
 
+// Default timeout values
+const (
+	DefaultHardwareProvisioningTimeout = 90 * time.Minute
+	DefaultClusterProvisioningTimeout  = 90 * time.Minute
+	DefaultClusterConfigurationTimeout = 30 * time.Minute
+)
+
+// Timeout config keys
```
irinamihai, line 82: It would be helpful to add more information to the comment: that each of these areoptional keys in the respective ConfigMaps such that timeout is configured for each operation.

internal/controllers/utils/utils.go
Missxiaoguo: The type of parameters in the timeouts is time.Duration, same for the default values in constants.go. I didn't use int because I feel a dedicated type time.Duration represents the intervals more clearly
```
	DefaultHardwareProvisioningTimeout = 90 * time.Minute
	DefaultClusterProvisioningTimeout  = 90 * time.Minute
	DefaultClusterConfigurationTimeout = 30 * time.Minute
```

internal/controllers/utils/utils.go
Missxiaoguo: Will rename and update to use GetConfigMapField

config/samples/v1alpha1_clustertemplate.yaml
```diff
@@ -44,9 +44,9 @@ spec:
             type: string
           install-plan-approval:
             type: string
-            default: Automatic
+        type: object
       clusterInstanceParameters:
-        description: clusterInstanceParameters defines the params that are allowed in the ProvisioningRequest spec.clusterInstanceInput
+        description: clusterInstanceParameters defines the params that are allowed in the ProvisioningRequest
```
bartwensley, line 49: nit: to match above:
clusterInstanceParameters defines the available parameters for cluster provisioning

config/samples/clusterinstance-defaults-configmap.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: clustertemplate-sample.v1.0.0-clusterinstance-defaults
+  namespace: clustertemplate-a-v4-16
+data:
+  clusterProvisioningTimeout: "80" # optional
```
bartwensley, line 7: We should probably be using https://pkg.go.dev/time#Duration for this (e.g. 80m, 1h, 30s) and all the other timeout values

internal/controllers/provisioningrequest_controller.go
```diff
@@ -79,6 +80,15 @@ type nodeInfo struct {
 	interfaces     []*hwv1alpha1.Interface
 }
 
+// timeouts holds the timeout values, in minutes,
+// for hardware provisioning, cluster provisioning
+// and cluster configuration.
+type timeouts struct {
+	hardwareProvisioningMinutes int
+	clusterProvisioningMinutes  int
+	configurationMinutes        int
```
bartwensley, line 89: Should these be time.Duration instead? And remove "Minutes" from the names.

bartwensley (commented):
Looks good - my only real comment is that it would be better to use durations (e.g. 10s, 30m, 5h) instead of having all the timeouts in seconds. I know it was like this already, so don't feel you need to change it now.

config/samples/clusterinstance-defaults-configmap.yaml
Missxiaoguo: Good point about using time duration format for these config values. I think that looks better than just using a number.

internal/controllers/provisioningrequest_controller.go
Missxiaoguo: Yeah, I feel it's clearer to use time.Duration to represent the timeout. I had them as time.Duration before, so I will change them back.

Missxiaoguo:
> Looks good - my only real comment is that it would be better to use durations (e.g. 10s, 30m, 5h) instead of having all the timeouts in seconds. I know it was like this already, so don't feel you need to change it now.

Added a new commit for this change. Have tested in the real env.

PR #246, opened 2024-10-08
Refactoring applied configuration timeout mechanism to infer timeout from policy history

edcdavid:
This is to support getting the Applied configuration timedout from ACM policy history. This is designed to be reusable with any similar timeout mechanism based on history 
Last unit test implements:
image

TODO:
- ~~unit tests~~
- K8s Unit tests
- system tests


internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	if lastCompleted.After(lastInProgress) && lastCompleted != resetTime {
+		// Overall state is Completed
+		return false
+	}
+	if lastCompleted == resetTime && initialTime != resetTime {
+		// Overall state is in progress from the beginning
+		return now.Sub(initialTime) > timeout
+	}
+	// If no Events are present, we can't timeout
+	return false
```
Missxiaoguo, line 115: It's very unlikely that all policies would have no status unless there's issue with the ACM. But, in theory, I think if enforced policies don't have a compliance status within the configured time, it should result in a timeout.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
 }
+
+const (
+	// Policy Specific states
+	compliant    = "Compliant"
+	nonCompliant = "NonCompliant"
+	pending      = "Pending"
+)
+
+// BuildPoliciesEventHistory creates an EventHistory object from a policyList
```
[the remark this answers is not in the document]
edcdavid: There is a unit test for the IsTimedOut function (TestEventHistory_IsTimedOut) which gives some examples/scenarios. 
>Are the events retrieved from detailsPerTemplate.History listed from the latest to the oldest?
yes, The list of event is sorted from first to last chronologically in the IsTimedOut function

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
 			})).
 		Complete(r)
 }
+
+const (
```
irinamihai, line 2524: These should be taken directly from the governance-policy-propagator repo already imported, as in [here](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/provisioningrequest_controller.go#L1134).

internal/controllers/utils/events.go
edcdavid: If there is no event history, wouldn't it also mean that remediation hasn't started yet ? Before the cluster is ready I think to remember that there is not history or status.

Missxiaoguo: Yeah, I mean once the managed cluster is ready, policies should have a status shortly after, unless something is wrong. The cluster ready status is checked before the timeout, so if it reaches this point and there's no status for a long time, it's likely that something has gone wrong.

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	Timestamp time.Time
+	State     string
+}
+
+// EventHistory contains a slice of events
+type EventHistory struct {
+	History []*Event
+}
+
+// IsTimedOut proceses a process history an determines the overall state implied
```
irinamihai, line 44: Maybe `IsTimedOut looks at the history and determines...`

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	// Sort events chronologically from old to new
+	sort.Slice(h.History, func(i, j int) bool {
+		return h.History[i].Timestamp.Before(h.History[j].Timestamp)
+	})
+
+	// currentState stores the running value of the State (InProgress or Completed) for each ObjectID
+	currentState := map[string]string{}
+	// Initialize the current state to NonCompliant
+	for _, event := range h.History {
+		currentState[event.ObjectID] = InProgress
```
irinamihai, line 62: The comment above says `NonCompliant`, but the code sets it to `InProgress`.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+	// Policy Specific states
+	compliant    = "Compliant"
+	nonCompliant = "NonCompliant"
+	pending      = "Pending"
+)
+
+// BuildPoliciesEventHistory creates an EventHistory object from a policyList
+func BuildPoliciesEventHistory(policies *policiesv1.PolicyList) (h utils.EventHistory) {
+	for _, policy := range policies.Items {
```
irinamihai, line 2533: From my understanding, if a policy is `Compliant` it can be skipped.

Also, I don't think we need to build such a big structure with all the history from all the policies. I don't see why we can't call `IsTimedOut` for each policy or even `detailsPerTemplate` and if any of them has timed out, then that's the final result and we can return without processing the history of other policies.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		for _, detailsPerTemplate := range policy.Status.Details {
```
irinamihai, line 2534: Same as above, if the current detail is Compliant, it should be skipped.

internal/controllers/provisioningrequest_controller.go
Missxiaoguo: Yeah, I am concerned about the eventHistory as it could potentially become very long. It seems like there's room for optimization.

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	initialTime := time.Time{}
+	// resetTime is a recording a zero time
+	resetTime := time.Time{}
+	// lastCompleted records the last time the overall state for all event transitioned to Completed
+	lastCompleted := time.Time{}
+	// lastInProgress records the last time the overall state for all event transitioned to InProgress
+	lastInProgress := time.Time{}
+	// Process each event chronologically one by one and recalculate the
+	// current state (InProgress or Completed)
+	for _, event := range h.History {
```
irinamihai, line 74: It's not clear to me why we cannot rely on the policy controller to have computed the correct state and it has to be recreated. I understand it's trying to follow the k8s principles and relying just on observed data, but I think in this operator, we can safely use what we are observing in the status of other CRs that are not managed by the current operator.
I think we only care if the Policy is currently NonCompliant/Pending and to see since when.
cc @pixelsoccupied

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+			if state != Completed {
+				overallState = InProgress
+				break
+			}
+		}
+		// Record the lastCompleted overall state
+		if overallState == Completed {
+			lastCompleted = event.Timestamp
+		}
+		// Record the lastCompleted overall state
```
Missxiaoguo, line 97: nit: lastInProgress

internal/controllers/provisioningrequest_controller.go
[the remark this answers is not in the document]
Missxiaoguo: I thought may be helpful with an example of policy history output. 🤔 

```
status:
  compliant: NonCompliant
  details:
  - compliant: NonCompliant
    history:
    - eventName: ztp-common-v4-16.common-v4.16-config-policy.17dc923974703ccc
      lastTimestamp: ""
      message: NonCompliant
    templateMeta:
    creationTimestamp: null
    name: config-policy-config-1
  - compliant: Pending
    history:
    - eventName: ztp-common-v4-16.common-v4.16-config-policy.17dc923974703ccc
       lastTimestamp: ""
       message: NonCompliant
    templateMeta:
    creationTimestamp: null
    name: config-policy-config-2
```

internal/controllers/utils/events.go
edcdavid: The advantage of processing the history is to detect the times in history when the policies where all compliant, even if these are not the latest. Since these times reset the timeout timer. So in the example below, if the timeout is set at 6s, at the 6s time, we would time out if we look only at the current state (non compliant). By looking at the history, we can see that we were compliant before at the 5s time so the timer was reset, counting from zero again. So the next timeout would be at time 11s :
image
@irinamihai did I understand the issue raised in this diagram properly? or this is not the problem described?

pixelsoccupied: I think what Irina is saying is that we can simply read the policy CR directly...instead of doing a conversion to something custom (EventHistory)? Basically we really need `BuildPoliciesEventHistory` func?

edcdavid: I created EventHistory on purpose to de-correlate event processing from the specific k8s policies event history. It makes it easier to test and understand without mixing k8s implementation details with it. Also, if needed we could reuse the code if there is another similar object with events that we need to process a timeout from. The BuildPoliciesEventHistory is just the specific part of this process, it just very specific to policies events and convert this specific flavor of events into something generic that we do not have to retest. If we get the timeout for another type of historical events, this build function would be the only one needed.

internal/controllers/provisioningrequest_controller.go
edcdavid: Fixed

internal/controllers/utils/events.go
edcdavid: Fixed

internal/controllers/utils/events.go
edcdavid: Fixed

internal/controllers/utils/events.go
edcdavid: Fixed

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	})
+
+	// currentState stores the running value of the State (InProgress or Completed) for each ObjectID
+	currentState := map[string]string{}
+	// Initialize the current state to NonCompliant
+	for _, event := range h.History {
+		currentState[event.ObjectID] = InProgress
+	}
+	// initialTime records the initial Event for all ObjectIDs
+	initialTime := time.Time{}
```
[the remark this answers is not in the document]
edcdavid: Fixed

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+	// resetTime is a recording a zero time
+	resetTime := time.Time{}
```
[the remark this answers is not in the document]
edcdavid: Fixed

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+					item := utils.Event{
+						ObjectID:  templateFullName,
+						Timestamp: complianceHistory.LastTimestamp.Time,
+						State:     utils.InProgress,
+					}
+					h.History = append(h.History, &item)
+				}
+				if strings.HasPrefix(complianceHistory.Message, nonCompliant) {
+					item := utils.Event{
+						ObjectID:  templateFullName,
```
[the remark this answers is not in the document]
edcdavid: Fixed

internal/controllers/provisioningrequest_controller.go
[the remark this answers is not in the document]
edcdavid: @Missxiaoguo I added a unit test for the function getting the generic event from the policy history after @pixelsoccupied 's suggestion. https://github.com/openshift-kni/oran-o2ims/pull/246#discussion_r1805431009

internal/controllers/provisioningrequest_controller.go
```diff
@@ -2558,3 +2520,46 @@ func (r *ProvisioningRequestReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+
+const (
+	// Policy Specific states
+	compliant    = "Compliant"
+	nonCompliant = "NonCompliant"
+	pending      = "Pending"
+)
+
+// BuildPoliciesEventHistory creates an EventHistory object from a policyList
+func BuildPoliciesEventHistory(policies *policiesv1.PolicyList) (h utils.EventHistory) {
```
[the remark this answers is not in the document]
edcdavid: The functions in events are generic, not specific to policy history, same as https://github.com/openshift-kni/oran-o2ims/pull/246#discussion_r1805503409

internal/controllers/utils/events.go
```diff
@@ -0,0 +1,116 @@
...
+// EventHistory contains a slice of events
+type EventHistory struct {
+	History []*Event
+}
+
+// IsTimedOut proceses a process history an determines the overall state implied
+// by the events.
+// returns true if for any ObjectID:
+// - the overall state was InProgress from the initial time for more than timeout
+// - the overall state was InProgress from the last transition to InProgress for more than timeout
```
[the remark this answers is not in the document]
edcdavid: Let's continue the discussion in https://github.com/openshift-kni/oran-o2ims/pull/246#discussion_r1803846429. In my understanding it is not the same since the timeout will be different based on the history.

internal/controllers/provisioningrequest_controller.go
edcdavid: how would we accurately calculate the timeout in the case describe in the diagram? same as https://github.com/openshift-kni/oran-o2ims/pull/246#discussion_r1803865576

internal/controllers/provisioningrequest_controller.go
edcdavid: Performance impact should be limited, seems like only 10 events Max are stored per policy. But I agree, it could be a concern for scaling to managing 3500 SNOs. planning to close the PR for now

edcdavid:
Closing this PR for now. It provides for a slightly more accurate timeout by processing historical events and follows k8s recommendation to get state from objects in the cluster instead of storing state but also required to process more data (Max 10 events per policies ) so it might cause issues when scaling to 3500 SNOs (although a slim chance). Planning to revisit interest for this if we encounter issues in the future.

PR #247, opened 2024-10-08
CNF-14782: CNF-14783: policyTemplateParameters & clusterInstanceParam…

irinamihai:
Description:
- add readme to explain how modifying policyTemplateParameters and clusterInstanceParameters affects policy configuration

### DIFF
--- a/docs/cluster_configuration.md
+++ b/docs/cluster_configuration.md
@@ -1,0 +1,179 @@
+# Cluster Configuration
+
+## Preparation
+TO DO
+
+## Initial install and configuration
+TO DO
+
+## Day 2 configuration
+
+### Updates to the clusterInstanceParameters field under ProvisioningRequest spec.templateParameters
+
+A ProvisioningRequest can be edited to update:
+* the cluster labels and annotations
+```yaml
+apiVersion: o2ims.provisioning.oran.org/v1alpha1
+kind: ProvisioningRequest
+metadata:
+  finalizers:
+    - provisioningrequest.o2ims.provisioning.oran.org/finalizer
+  name: sno-ran-du-1
+spec:
+  description: Provisioning request for basic SNO with sample ACM policies
+  name: Dev-main-SNO-Provisioning-sno-ran-du-1
+  templateName: sno-ran-du
+  templateVersion: v1
+  templateParameters:
+    clusterInstanceParameters:
+      clusterName: sno-ran-du-1
+      extraAnnotations:
+        ManagedCluster:
+          test: test <<< added as a Day 2 change
+      extraLabels:
+        ManagedCluster:
+          sno-ran-du-policy: v1
+          test: test <<< added as a Day 2 change
+```
+* the node labels and annotations
+```yaml
+apiVersion: o2ims.provisioning.oran.org/v1alpha1
+kind: ProvisioningRequest
+metadata:
+  finalizers:
+    - provisioningrequest.o2ims.provisioning.oran.org/finalizer
+  name: sno-ran-du-1
+spec:
+  description: Provisioning request for basic SNO with sample ACM policies
+  name: Dev-main-SNO-Provisioning-sno-ran-du-1
+  templateName: sno-ran-du
+  templateVersion: v1
+  templateParameters:
+    clusterInstanceParameters:
+      clusterName: sno-ran-du-1
+      nodes:
+      - bmcCredentialsName:
+          name: sno-ran-du-1-bmc-secret
+        extraAnnotations:
+          BareMetalHost:
+            test: test <<< added as a Day 2 change
+        extraLabels:
+          BareMetalHost:
+            test: test <<< added as a Day 2 change
+```
+The `status.conditions` records the success/failure of the update.
+
+The cluster configuration goes to the `ManagedCluster` CR and the nodes configuration to the corresponding `BMH`s, as expected.
+
+**Note**: ManagedCluster and node extra labels&annotations are the only fields that can be edited post installation. All the other fields are immutable and are rejected by the IMS operator. These changes would be rejected anyway by webhooks put in place by other operators for cluster installation resources (ex: `ClusterDeployment`)
+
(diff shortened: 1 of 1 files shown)


docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+## Initial install and configuration
+TO DO
+
+## Day 2 configuration
+
+### Updates to ProvisioningRequest spec.templateParameters.clusterInstanceParameters
+
+A ProvisioningRequest can be edited to update:
+* the cluster labels and annotation
+```
```
sudomakeinstall2, line 15: nit:
```yaml
will add colors

File: docs/policy_configuration.md [hunk 2]
```diff
@@ -0,0 +1,160 @@
...
+        ManagedCluster:
+          test: test <<< added as a Day 2 change
+      extraLabels:
+        ManagedCluster:
+          clustertemplate-a-policy: v1
+          test: test <<< added as a Day 2 change
+```
+* the node labels and annotations
+```yaml
+spec:
```
bartwensley, line 30: I like it that you included the description, name, templateName in the example below - add it here too?

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+  finalizers:
+    - provisioningrequest.o2ims.provisioning.oran.org/finalizer
+  generation: 2
+  name: cnfdg11
+  resourceVersion: "99518665"
+  uid: 397a9bae-9483-417d-a808-3f60343760a4
+spec:
+  description: Provisioning request for basic SNO with sample ACM policies
+  name: Dev-main-SNO-Provisioning-sno-1
+  templateName: clustertemplate-a
```
bartwensley, line 67: We should probably come up with better example template names for our documentation. How about:

- sno-ran-du
- sno-ran-du-dell
- sno-ran-du-dell-spr
- sno-ran-du-dell-intel

I'm not sure what is best? @browsell - any thoughts?

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+  templateParameters:
+    nodeClusterName: site-sno-1
+    oCloudSiteId: local-west-12345
+    policyTemplateParameters:
+      sriov-network-vlan-1: "111"
+      sriov-network-vlan-2: "222"
+```
+**Note:** Only policy configuration values that are exposed in the associated ClusterTemplate under `spec.templateParameterSchema > policyTemplateParameters` can be updated through the ProvisioningRequest.
```
bartwensley, line 75: Should it be spec.templateParameterSchema.policyTemplateParameters?

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+    policyNamespace: ztp-clustertemplate-a-v4-16
+    remediationAction: enforce
+```
+
+**Notes**:
+* The format of the way the `nonCompliantAt` timestamps might move to another structure in the status, but it will still be recorded.
+* Some changes happen so fast that the Policy doesn't even switch to `NonCompliant`, so the IMS operator cannot record the event. In this case, IMS still holds a correct recording since all the policies are/remain Compliant.
+* Once an enforce `NonCompliant` Policy becomes `Compliant` again, the `status.policies` is updated, the `status.clusterDetails.nonCompliantAt` value removed and the `ConfigurationApplied` condition updated to show that the configuration is up to date:
+* When refactored, the start and end times of the configuration being NonCompliant will be recorded.
+```yaml
```
bartwensley, line 134: Add a sentence describing what the following yaml is about - is it an example of the status after the policies have all been applied?

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+            test: test <<< added as a Day 2 change
+        extraLabels:
+          BareMetalHost:
+            test: test <<< added as a Day 2 change
+```
+The `status.conditions` records the success/failure of the update.
+
+The cluster configuration goes to the `ManagedCluster` CR and the nodes configuration to the corresponding `BMH`s, as expected.
+
+**Note**: ManagedCluster  and node extra labels&annotations are the only fields that can be edited post installation. All the other fields are immutable and would be rejected by webhooks put in place by other operators for cluster installation resources (ex: `ClusterDeployment`)
```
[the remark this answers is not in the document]
irinamihai: We do, I meant to say that even if IMS wouldn't reject the changes, they would still end up being rejected by web hooks. I will rephrase it.

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
...
+```yaml
+apiVersion: o2ims.provisioning.oran.org/v1alpha1
+kind: ProvisioningRequest
+metadata:
+  finalizers:
+    - provisioningrequest.o2ims.provisioning.oran.org/finalizer
+  generation: 2
+  name: cnfdg11
+  resourceVersion: "99518665"
+  uid: 397a9bae-9483-417d-a808-3f60343760a4
```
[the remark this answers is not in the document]
irinamihai: I think it would make sense to keep the finalizer and name, but I agree that the others should be removed.

docs/policy_configuration.md
irinamihai: `sno-ran-du` is probably generic enough?

docs/policy_configuration.md
irinamihai: I didn't include `templateParameterSchema.policyTemplateParameters` because `policyTemplateParameters` is not an actual part of the `spec`, but part of the policy template schema and I wanted to make the distinction.

docs/policy_configuration.md
```diff
@@ -0,0 +1,160 @@
+# Policy Configuration
```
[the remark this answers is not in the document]
bartwensley: Thanks - should update the filename too.

bartwensley:
Looks good to me once the filename is updated.

irinamihai:
> Looks good to me once the filename is updated.

@bartwensley Done. Thank you.

PR #251, opened 2024-10-10
Aligns with the expected returned field names specified in the specification

alosadagrande:
The returned fields for some API endpoints are returning incorrect field names based on the specification (O-RAN 2IMS INTERFACE v6 spec). See for instance tables:

- Table 2.1.1.1.3-1: Attributes of the ResourceType
- Table 2.1.1.1.4-1: Attributes of the ResourcePool
- Table 2.1.1.1.5-1: Attributes of the Resource

This fix modifies the returned value and replaces:

- resourceTypeID by resourcetTypeId
- globalLocationID by globalLocationId
- resourcePoolID by resourcePoolId

With these changes, some faulty [O2IMS Compliance Test](https://github.com/o-ran-sc/it-test/tree/master/test_scripts/O2IMS_Compliance_Test) pass. 

PD. For testing you can use my ORAN-O2IMS operator image, which already contains the fix at quay.io/alosadag/oran-o2ims:v4.16.1

### DIFF
--- a/internal/service/alarm_handler_test.go
+++ b/internal/service/alarm_handler_test.go
@@ -226,7 +226,7 @@ var _ = Describe("Alarm handler", func() {
 					RespondWithList(
 						data.Object{
 							"name":           "spoke0",
-							"resourcePoolID": "spoke0",
+							"resourcePoolId": "spoke0",
 						},
 					),
 				)
@@ -429,7 +429,7 @@ var _ = Describe("Alarm handler", func() {
 						RespondWithList(
 							data.Object{
 								"name":           "spoke0",
-								"resourcePoolID": "spoke0",
+								"resourcePoolId": "spoke0",
 							},
 						),
 					),
@@ -509,7 +509,7 @@ var _ = Describe("Alarm handler", func() {
 						RespondWithList(
 							data.Object{
 								"name":           "spoke0",
-								"resourcePoolID": "spoke0",
+								"resourcePoolId": "spoke0",
 							},
 						),
 					),
--- a/internal/service/alarm_mapper.go
+++ b/internal/service/alarm_mapper.go
@@ -183,7 +183,7 @@ func (r *AlarmMapper) MapItem(ctx context.Context,
 	if err != nil {
 		// Instance is not available for cluster global alerts
 		alarmEventRecordId = fmt.Sprintf("%s_%s", alertName, resourcePoolName)
-		resourceID = resourcePool["resourcePoolID"].(string)
+		resourceID = resourcePool["resourcePoolId"].(string)
 	} else {
 		alarmEventRecordId = fmt.Sprintf("%s_%s_%s", alertName, resourcePoolName, alertInstance)
 		resource, err := r.fetchResource(ctx, resourcePoolName, alertInstance)
--- a/internal/service/resource_fetcher.go
+++ b/internal/service/resource_fetcher.go
@@ -160,7 +160,7 @@ func (b *ResourceFetcherBuilder) Build() (
 }
 
 // GetResourceTypeID generates a typeID from a search API object.
-func (h *ResourceFetcher) GetResourceTypeID(from data.Object) (resourceTypeID string, err error) {
+func (h *ResourceFetcher) GetResourceTypeID(from data.Object) (resourceTypeId string, err error) {
 	kind, err := data.GetString(from, "kind")
 	if err != nil {
 		return
@@ -177,7 +177,7 @@ func (h *ResourceFetcher) GetResourceTypeID(from data.Object) (resourceTypeID st
 		if err != nil {
 			return
 		}
-		resourceTypeID = fmt.Sprintf("node_%s_cores_%s", cpu, architecture)
+		resourceTypeId = fmt.Sprintf("node_%s_cores_%s", cpu, architecture)
 	}
 
 	return
--- a/internal/service/resource_handler.go
+++ b/internal/service/resource_handler.go
@@ -372,14 +372,14 @@ func (r *ResourceHandler) mapNodeItem(ctx context.Context, // nolint: unparam
 		return
 	}
 
-	resourcePoolID, err := data.GetString(from,
-		graphql.PropertyNode("resourcePoolID").MapProperty())
+	resourcePoolId, err := data.GetString(from,
+		graphql.PropertyNode("resourcePoolId").MapProperty())
 	if err != nil {
 		return
 	}
 
-	globalAssetID, err := data.GetString(from,
-		graphql.PropertyNode("globalAssetID").MapProperty())
+	globalAssetId, err := data.GetString(from,
+		graphql.PropertyNode("globalAssetId").MapProperty())
 	if err != nil {
 		return
 	}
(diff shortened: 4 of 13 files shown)


alosadagrande:
For testing you can use my ORAN-O2IMS operator image, which already contains the fix at quay.io/alosadag/oran-o2ims:**v4.16.1**

Notice that there are other attributes for `Resources` such as resourceId and globalAssetId that in the code are spelled as resourceID and globalAssetID. See `Table 2.1.1.1.5-1: Attributes of the Resource`. In those cases, there are no compliance tests associated.

On the other side there is inconsistency in the naming convention for the AlarmEventRecord ( see Table 2.1.1.3.2-1: Attributes of the AlarmEventRecord) . Only this one. Probably this is something that needs to be raised to the WG6.

internal/service/resource_fetcher.go
```diff
@@ -160,7 +160,7 @@ func (b *ResourceFetcherBuilder) Build() (
 }
 
 // GetResourceTypeID generates a typeID from a search API object.
-func (h *ResourceFetcher) GetResourceTypeID(from data.Object) (resourceTypeID string, err error) {
+func (h *ResourceFetcher) GetResourceTypeID(from data.Object) (resourceTypeId string, err error) {
```
alegacy, line 163: changing the name of the argument name to `resourceTypeId` has created a mismatch between that and the name of the function that spells id as `ID` rather than `Id`.   We should be consistent one way or another.   I understand that the JSON reported externally must be `Id`, and to be honest that's my preference for all acronyms in symbols, but the reason the function and other symbols are spelled as `ID` is because of the Google coding guidelines (https://google.github.io/styleguide/go/decisions#initialisms).   I have not seen any statement anywhere that says we are following that but it does seem to be used in some places in this repo (not everywhere).  

IMHO... `Id` is better than `ID`... especially when acronyms are embedded in the middle of a name.  

cc: @bartwensley

internal/service/alarm_mapper.go
```diff
@@ -222,7 +222,7 @@ func (r *AlarmMapper) MapItem(ctx context.Context,
 	to = data.Object{
 		"alarmEventRecordId": alarmEventRecordId,
 		"resourceID":         resourceID,
-		"resourceTypeID":     resourceTypeID,
+		"resourceTypeId":     resourceTypeId,
```
bartwensley, line 225: Unfortunately, the AlarmEventRecord is defined using "resourceTypeID" so we need to keep that for this string.

internal/service/alarm_handler_test.go
```diff
@@ -250,7 +250,7 @@ var _ = Describe("Alarm handler", func() {
 				// Verify first result:
 				Expect(items[0]).To(MatchJQ(`.alarmEventRecordId`, "alert1_spoke0"))
 				Expect(items[0]).To(MatchJQ(`.resourceID`, "spoke0"))
-				Expect(items[0]).To(MatchJQ(`.resourceTypeID`, ""))
+				Expect(items[0]).To(MatchJQ(`.resourceTypeId`, ""))
```
bartwensley, line 253: This should stay as resourceTypeID for the alarmEventRecord.

internal/service/alarm_handler_test.go
```diff
@@ -262,7 +262,7 @@ var _ = Describe("Alarm handler", func() {
 				// Verify seconds result:
 				Expect(items[1]).To(MatchJQ(`.alarmEventRecordId`, "alert2_spoke0_host0"))
 				Expect(items[1]).To(MatchJQ(`.resourceID`, "host0"))
-				Expect(items[1]).To(MatchJQ(`.resourceTypeID`, "type0"))
+				Expect(items[1]).To(MatchJQ(`.resourceTypeId`, "type0"))
```
bartwensley, line 265: This should stay as resourceTypeID for the alarmEventRecord.

internal/service/alarm_handler_test.go
```diff
@@ -525,7 +525,7 @@ var _ = Describe("Alarm handler", func() {
 				// Verify the result:
 				Expect(response.Object).To(MatchJQ(`.alarmEventRecordId`, "alert_spoke0"))
 				Expect(response.Object).To(MatchJQ(`.resourceID`, "spoke0"))
-				Expect(response.Object).To(MatchJQ(`.resourceTypeID`, ""))
+				Expect(response.Object).To(MatchJQ(`.resourceTypeId`, ""))
```
bartwensley, line 528: This should stay as resourceTypeID for the alarmEventRecord.

internal/service/alarm_notification_handler_test.go
```diff
@@ -113,7 +113,7 @@ var _ = Describe("alarm Notification handler", func() {
 
 				requestObj := data.Object{
 					"alarmEventRecordId": "a267bbd0-57aa-4ea1-b030-a300d420ef19",
-					"resourceTypeID":     "c1fe0c43-28e3-4b61-aac5-84bea67551ea",
+					"resourceTypeId":     "c1fe0c43-28e3-4b61-aac5-84bea67551ea",
```
bartwensley, line 116: This should stay as resourceTypeID for the alarmEventRecord.

internal/service/alarm_notification_handler_test.go
```diff
@@ -157,7 +157,7 @@ var _ = Describe("alarm Notification handler", func() {
 						"endpoint":                  "https",
 					},
 					"resourceID":         "1cf16c13-f5cb-497a-97d5-e909cba396a3",
-					"resourceTypeID":     "node_8_cores_amd64",
+					"resourceTypeId":     "node_8_cores_amd64",
```
bartwensley, line 160: This should stay as resourceTypeID for the alarmEventRecord.

internal/service/resource_fetcher.go
bartwensley: I would prefer the field names, variable names and function names to all match, but that could be deferred until later as there is probably quite a bit of cleanup to do here.

internal/service/resource_handler.go
```diff
@@ -412,12 +412,12 @@ func (r *ResourceHandler) mapNodeItem(ctx context.Context, // nolint: unparam
...
-		"resourceID":     resourceID,
-		"resourceTypeID": resourceTypeID,
+		"resourceId":     resourceID,
+		"resourceTypeId": resourceTypeId,
 		"description":    description,
 		"extensions":     extensionsMap,
-		"resourcePoolID": resourcePoolID,
-		"globalAssetID":  globalAssetID,
+		"resourcePoolId": resourcePoolId,
+		"globalAssetId":  globalAssetID,
```
bartwensley, line 420: Should probably change the globalAssetID variable to match as well to be consistent with the others.

bartwensley (changes requested):

internal/service/alarm_mapper.go
alosadagrande: OK. Honestly, it isn't easy to understand how the naming convention for the alarmEventRecord was chosen in the standard compared with other attributes.

bartwensley (approved):

PR #132, opened 2024-07-30
[doc] Adds operator deployment steps in OpenShift

alosadagrande:
This PR updates documentation:

- It divides the README into two main blocks: deployment via operator in OpenShift and local deployment (testing)
- Deployment of the IMS operator on top of OpenShift
- Configuration of the different microservices of the IMS operator using the ORANO2MS CR. There was no example of that CR in the repo.
- Currently only metadata, deployment manager and resource server. The rest could be added once they are validated in the operator deployment.
- An index is added to ease navigation through the information

Let me know your thoughts.

PR #260, opened 2024-10-15
Starting Metadata service automatically 

mmorency2021:
This is part of ORAN - Project infrastructure - Automatically start services
https://issues.redhat.com/browse/MGMT-18997

- Using Target deploy to install operator and enable services

- Using clusterID to deploy meta data server , user can edit later on the CR  to  add more parameters

### DIFF
--- a/docs/cluster_configuration.md
+++ b/docs/cluster_configuration.md
@@ -94,10 +94,10 @@ spec:
 Once the update is made, the `<cluster-name>-pg` ConfigMap in the `ztp-<cluster-template-namespace>` namespace gets updated with the new value. This ConfigMap is used by the ACM policies in their hub templates.
 ```yaml
 $  oc get clustertemplate -A
-NAMESPACE                 NAME                   AGE
-sno-ran-du-v4-16   sno-ran-du.v1   3d23h
+NAMESPACE                 NAME                     AGE
+sno-ran-du-v4-Y-Z         sno-ran-du.v4-Y-Z-1      3d23h
 
-$  oc get cm -n ztp-sno-ran-du-v4-16 sno-ran-du-1-pg -oyaml
+$  oc get cm -n ztp-sno-ran-du-v4-Y-Z <cluster name>-pg -oyaml
 apiVersion: v1
 data:
   cpu-isolated: 0-1,64-65
@@ -111,7 +111,7 @@ data:
 kind: ConfigMap
 metadata:
   name: sno-ran-du-1-pg
-  namespace: ztp-sno-ran-du-v4-16
+  namespace: ztp-sno-ran-du-v4-Y-Z
 ```
 
 Once a policy matched with a ManagedCluster deployed through a ProvisioningRequest becomes `NonCompliant`, it's reflected in the ProvisioningRequest `status.policies` and the time when it becomes `NonCompliant` is also recorded. The `ConfigurationApplied` condition reflects that the configuration is being applied.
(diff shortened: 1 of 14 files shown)


bartwensley:
Please take a look at the comments on https://github.com/openshift-kni/oran-o2ims/pull/132 - according to the comments there, the token is not necessary and the image isn't required in the Inventory CR.

config/manager/metadata-server.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: orano2ims
+    app.kubernetes.io/part-of: oran-o2ims
+  name: metadata-server
+  namespace: oran-o2ims
+spec:
+  alarmSubscriptionServerConfig:
+    enabled: false
+  cloudId: ae6bd821-a4a4-4e53-89d1-94609968e40e
```
alegacy, line 15: This `cloudId` is meant to be the `globalCloudId`... as discussed internally this should be made optionally and excluded from this CR and only provided by the end user when the system is ready to be registered with the SMO, this is because the system cannot register with the SMO until the administrator has setup this cloud in the SMO and has acquired the globalCloudId.  

So i think what should happen here is that this field be made optional, and renamed to `globalCloudId` so that its intent is clear.

config/manager/metadata-server.yaml
alegacy, line 1: This file should probably be named `inventory.yaml` since metadata-server is only one of the servers being started, but not the entire purpose of this CR.

Makefile
```diff
@@ -160,9 +160,14 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
 ##@ Deployment
+## settings parameter for deployment 
+
+# getting the cloud id and the IngressHost 
+
+CLOUD_ID := $(shell oc get clusterversion -o jsonpath='{.items[].spec.clusterID}{"\n"}')
 
 ifndef ignore-not-found
-  ignore-not-found = false
+  ignore-not-found = true
```
alegacy, line 170: This should be removed before merging... and can be specified at the command line like this:

```
make IMAGE_TAG_BASE=<your repo name here> undeploy docker-push install deploy ignore-not-found=true
```

Makefile
```diff
@@ -160,9 +160,14 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	rm Dockerfile.cross
 
 ##@ Deployment
+## settings parameter for deployment 
+
+# getting the cloud id and the IngressHost 
+
+CLOUD_ID := $(shell oc get clusterversion -o jsonpath='{.items[].spec.clusterID}{"\n"}')
```
alegacy, line 167: These new lines can all be removed.  See comments in the metadata-server.yaml file.

config/manager/metadata-server.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: o2ims.oran.openshift.io/v1alpha1
+kind: Inventory
+metadata:
+  annotations:
+  labels:
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: orano2ims
+    app.kubernetes.io/part-of: oran-o2ims
+  name: metadata-server
```
alegacy, line 10: This name should be changed to something more generic... like `default` or `cluster` which are two examples that seem to be used by other operators that automatically create an instance of themselves.   I prefer `default`... @bartwensley @browsell thoughts?

browsell: +1 for default

PR #263, opened 2024-10-16
CNF-14785: Validate new spoke CR to be managed by the ACM PG

irinamihai:


docs/cluster_configuration.md
```diff
@@ -257,3 +257,63 @@ For updating a manifest in an existing ACM PolicyGenerator, the following steps
       policyNamespace: ztp-clustertemplate-a-v4-16
       remediationAction: enforce
     ```
+
+### Adding a new manifest to an existing ACM PolicyGenerator
+
+This usecase is identical to the [previous one](#updates-to-an-existing-acm-policygenerator-manifest), with the following distinctions:
+* If the new manifest does not have a corresponding `source-cr` file, the CSP should add a new yaml file to the `custom-crs` directory.
```
edcdavid, line 264: what does CSP mean?

docs/cluster_configuration.md
```diff
@@ -257,3 +257,63 @@ For updating a manifest in an existing ACM PolicyGenerator, the following steps
       policyNamespace: ztp-clustertemplate-a-v4-16
```
Missxiaoguo, line 257: Not related to this PR! Should these namespaces be changed to `ztp-sno-ran-du-v4-16` which match the PG namespace?

docs/cluster_configuration.md
irinamihai: Cloud Service Provider

docs/cluster_configuration.md
irinamihai: Yes, good catch. I'll update all of them in a future PR. Thank you.

PR #264, opened 2024-10-17
CNF-14630: CNF14631: Add initial upgrade handler

sudomakeinstall2:
- Add release field to ClusterTemplate, sepcifying openshift release for the template
- Add upgradeDefaults to ClusterTemplate.Templates, specifying configmap name for upgrade information to this templater
- Add initital upgrade handler. After detecting that clusterTemplate.release is higher than deployed managedCluster.OpenshiftVersion, upgrade handler creates and ibgu to upgrade the cluster. The upgrade information is read from upgradeDefaults configmap specified in the clusterTemplate.


internal/controllers/provisioningrequest_controller.go
```diff
@@ -284,6 +284,24 @@ func (t *provisioningRequestReconcilerTask) run(ctx context.Context) (ctrl.Resul
 		if !utils.IsClusterProvisionCompleted(t.object) || requeue {
 			return requeueWithLongInterval(), nil
 		}
+
+		shouldUpgrade, err := t.IsUpgradeRequested(ctx, renderedClusterInstance.GetName())
+		fmt.Println(shouldUpgrade)
+		if err != nil {
+			return requeueWithError(err)
+		}
+		if utils.IsClusterProvisionCompleted(t.object) && shouldUpgrade {
```
jc-rh, line 293: have you done e2e test including finalize? After post pivot, we would be getting the new OCP version. handleUpgrade would be getting skipped. This condition probably should be more like "shouldUpgrade || upgradeInProgress"?

sudomakeinstall2: not with finalize.
The condition will change after Angie's PR is merged and with the next story.
Probably as per document we will add a new condition `ClusterUpgraded` and we can check with that.

internal/controllers/upgrade_handler.go
```diff
@@ -0,0 +1,157 @@
...
+	siteconfig "github.com/stolostron/siteconfig/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (t *provisioningRequestReconcilerTask) IsUpgradeRequested(
```
Missxiaoguo, line 19: Maybe add a brief comment describing that if release version in the template is greater than the current ocp version, an upgrade will be triggered

internal/controllers/upgrade_handler.go
```diff
@@ -0,0 +1,157 @@
...
+			ctx,
+			fmt.Sprintf(
+				"Upgrade initiated. Created IBGU %s in the namespace %s",
+				ibgu.GetName(),
+				ibgu.GetNamespace(),
+			),
+		)
+
+		utils.SetStatusCondition(&t.object.Status.Conditions,
+			utils.PRconditionTypes.ClusterProvisioned,
```
Missxiaoguo, line 83: Based on the doc, it looks like we are considering to add a didicated condition for upgrades?

internal/controllers/utils/conditions.go
```diff
@@ -136,3 +138,14 @@ func IsSmoRegistrationCompleted(cr *inventoryv1alpha1.Inventory) bool {
 		string(InventoryConditionTypes.SmoRegistrationCompleted))
 	return condition != nil && condition.Status == metav1.ConditionTrue
 }
+
+// IsClusterUpgradeInProgress checks if the cluster upgrade condition status is in progress
+func IsClusterUpgradeInProgress(cr *provisioningv1alpha1.ProvisioningRequest) bool {
+	condition := meta.FindStatusCondition(cr.Status.Conditions, (string(PRconditionTypes.UpgradeCompleted)))
```
jc-rh, line 144: This probably works in most cases. It might be a bit more reliable if we do this based on the existence of the IBGU CR.

jc-rh: The essence of this check is more like "if there is any upgrade stuff that needs to be taken care of by the handleUpgrade function".

internal/controllers/upgrade_handler.go
```diff
@@ -0,0 +1,164 @@
...
+			message,
+		)
+	} else {
+		utils.SetStatusCondition(&t.object.Status.Conditions,
+			utils.PRconditionTypes.UpgradeCompleted,
+			utils.CRconditionReasons.Completed,
+			metav1.ConditionTrue,
+			"Upgrade is completed",
+		)
+		err := t.client.Delete(ctx, ibgu)
```
irinamihai, line 133: I think deleting the IBGU should move at the end. If the delete fails, then the ProvisioningRequest status is not updated to UpgradeCompleted True, even if it should.

internal/controllers/upgrade_handler.go
```diff
@@ -0,0 +1,164 @@
+package controllers
```
irinamihai, line 1: I think we should rename this file to `provisioningrequest_cluster_upgrade.go` or something similar, but to have the controller using these functions prefixed. It would make it clear that it's used for provisioning.

internal/controllers/upgrade_handler.go
```diff
@@ -0,0 +1,164 @@
...
+		for _, action := range cluster.FailedActions {
+			message += fmt.Sprintf("Action %s failed: %s\n", action.Action, action.Message)
+		}
+		return true, message
+	}
+	return false, ""
+}
+
+func IsIBGUProgressing(cr *ibgu.ImageBasedGroupUpgrade) bool {
+	condition := meta.FindStatusCondition(cr.Status.Conditions, "Progressing")
```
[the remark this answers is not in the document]
jc-rh: ConditionTypes struct is not exposed in api unfortunately

internal/controllers/upgrade_handler.go
jc-rh: I already made quite a bit additional changes in this file. I will do the renaming in the subsequent PR.

internal/controllers/upgrade_handler.go
jc-rh: I've re-arranged this function quite a bit. I will see if I can address this in my version without making it too complicated.

internal/controllers/clustertemplate_controller.go
```diff
@@ -245,6 +260,25 @@ func (t *clusterTemplateReconcilerTask) validateClusterTemplateCR(ctx context.Co
...
+func validateUpgradeDefaultsConfigmap(
+	ctx context.Context, c client.Client, name, namespace, key string,
+) error {
+	ibgu, err := utils.GetIBGUFromUpgradeDefaultsConfigmap(ctx, c, name, namespace, key, "name", namespace)
+	if err != nil {
+		return fmt.Errorf("failed to get IBGU from upgrade defaults configmap: %w", err)
+	}
+	opts := []client.CreateOption{}
+	opts = append(opts, client.DryRunAll)
+	err = c.Create(ctx, ibgu, opts...)
```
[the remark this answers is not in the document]
jc-rh: This is a dry run

PR #267, opened 2024-10-21
CNF-14859: Alarms openapi spec and generated server side code to start alarm server

pixelsoccupied:
This PR introduces the OpenAPI spec which covers all the endpoints exposed for both external (o-ran) and internal clients. 

- With OpenAPI spec we generate server-side code using `https://github.com/oapi-codegen/oapi-codegen` 
- Enabled all the basic input validation and all error return o-ran specified `ProblemDetails`
- Fully configured server to allow us to start implementing codegen interfaces

Bonus: included an example to generate client-side code


internal/service/alarms/exmaple/client-codegen-config/oapi-codegen-config.yaml
alegacy, line 1: typo in directory name... `exmaple` -> `example`

internal/service/alarms/exmaple/client-codegen-config/overlay.yaml
alegacy, line 1: Is there a reason/requirement why we are keeping this client config?

internal/service/alarms/internal/alarms_server.go
```diff
@@ -0,0 +1,90 @@
...
+	p := api.ProblemDetails{
+		Detail: "invalid `filter` parameter syntax",
+		Status: http.StatusBadRequest,
+	}
+	return api.GetAlarms400ApplicationProblemPlusJSONResponse(p), nil
+}
+
+func (a AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequestObject) (api.GetAlarmResponseObject, error) {
+	// TODO implement me
+	alrm := api.AlarmEventRecord{
```
alegacy, line 52: could we spell out `alarm` rather than `alrm`

internal/service/alarms/exmaple/client-codegen-config/overlay.yaml
pixelsoccupied: No reason really...just as an example in case anyone wants to use the same library and generate some code. I'll just remove it and note in the readme instead.

internal/service/alarms/main.go
```diff
@@ -0,0 +1,127 @@
...
+		ErrorHandlerFunc: getOranReqErrFunc(),
+	}
+
+	// Create the handler
+	handler := api.HandlerWithOptions(alarmServerStrict, opt)
+
+	// Server config
+	srv := &http.Server{
+		Handler:      handler,
+		Addr:         net.JoinHostPort("0.0.0.0", "8080"),
```
alegacy, line 59: we should restrict this to `127.0.0.1` since we have a proxy sitting in front of us.

internal/service/alarms/main.go
alegacy, line 1: Is there a reason this is called `main.go`... i thought we were still going to start this as a sub-command of the `oran-o2ims-operator` binary?

internal/service/alarms/exmaple/client-codegen-config/oapi-codegen-config.yaml
pixelsoccupied: removed this directory

internal/service/alarms/exmaple/client-codegen-config/overlay.yaml
pixelsoccupied: updated

internal/service/alarms/internal/alarms_server.go
pixelsoccupied: updated

internal/service/alarms/main.go
pixelsoccupied: updated

internal/service/alarms/main.go
pixelsoccupied: No reason. The `main` func is just there so we can quickly start the server until we fully integrate with cli (at which point we can just remove that as well). 

Renamed to `serve.go` to make it clear.

internal/service/alarms/internal/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+
+// LogDuration log time taken to complete a request.
+// TODO: This is just get started with middleware but should be replaced with something that's more suitable for production i.e OpenTelemetry
+// https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/examples/prometheus/main.go
+func LogDuration() Middleware {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			startTime := time.Now()
+			next.ServeHTTP(w, r)
+			fmt.Printf("%q took %s\n", r.RequestURI, time.Since(startTime))
```
alegacy, line 24: These errors should go through a logging framework.

internal/service/alarms/internal/middleware.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+// AlarmsOapiValidation to validate all incoming requests as specified in the spec
+func AlarmsOapiValidation() Middleware {
+	// This also validates the spec file
+	swagger, err := generated.GetSwagger()
+	if err != nil {
+		// TODO: handle this error
+		fmt.Printf("failed to get swagger")
```
alegacy, line 35: The error should be wrapped for logging.

internal/service/alarms/internal/middleware.go
pixelsoccupied: logging instead of printing everywhere now.

internal/service/alarms/internal/middleware.go
pixelsoccupied: updated

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+        '403':
+          description: Forbidden.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+        '404':
```
mlguerrero12, line 65: I've seen this 404 for lists when filtering is used and there no matches. But we can also return an empty list with 200 OK for this case (with filtering). If so, we won't need this response code.

Both are fine in my opinion but given that the specification explicitly calls for 200 OK for empty lists. I think we can go for the second approach. (200 OK empty list with and without filtering).

Unless your intention with this response code was to question the existence of the endpoint. Hmm, I'd say 500 is better in that case.

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+        - alarms
+      parameters:
+        - in: path
+          name: alarmEventRecordId
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        '200':
```
mlguerrero12, line 92: missing response codes? 400, 500

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+        '404':
+          description: The specified AlarmEventRecord was not found.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+
+    patch:
+      operationId: AckAlarm
+      summary: Modify exactly one alarm to acknowledge
```
mlguerrero12, line 107: to acknowledge and clear

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+          schema:
+            type: string
+            format: uuid
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AlarmEventRecordModifications'
+      responses:
```
mlguerrero12, line 123: Specification mentions 409 and 412 codes. Also 400 is important here

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/AlarmSubscriptionInfo'
+        '400':
+          description: Bad request
+          content:
```
mlguerrero12, line 168: don't see 403 or 404. I think we should be consistent with response codes for list operations

internal/service/alarms/api/openapi.yaml
```diff
@@ -0,0 +1,756 @@
...
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+        '500':
+          description: Internal server error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '#/components/schemas/ProblemDetails'
+      callbacks:
```
mlguerrero12, line 209: This is not part of this API. We will need to generate a client for this

internal/service/alarms/serve.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Register the handler
+	api.HandlerWithOptions(alarmServerStrictHandler, opt)
+
+	// Server config
+	srv := &http.Server{
+		Handler:      r,
+		Addr:         net.JoinHostPort("127.0.0.1", "8080"),
+		ReadTimeout:  5 * time.Second,
```
mlguerrero12, line 63: let's make this constants

internal/service/alarms/serve.go
```diff
@@ -0,0 +1,130 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
```
mlguerrero12, line 8: shouldn't we use the log package of the repo?

internal/service/alarms/serve.go
```diff
@@ -0,0 +1,130 @@
...
+// gracefulShutdown allow graceful shutdown with timeout
+func gracefulShutdown(srv *http.Server) error {
+	// Create shutdown context with 10 second timeout
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Attempt graceful shutdown
+	if err := srv.Shutdown(ctx); err != nil {
+		// TODO: handle this error
+		slog.Error(fmt.Sprintf("failed graceful shutdown: %v", err))
```
mlguerrero12, line 104: error is logged outside as well

internal/service/alarms/internal/middleware.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Clear out the servers array in the swagger spec, that skips validating
+	// that server names match. We don't know how this thing will be run.
+	swagger.Servers = nil
+
+	return oapimiddleware.OapiRequestValidatorWithOptions(swagger, &oapimiddleware.Options{
+		Options: openapi3filter.Options{
+			AuthenticationFunc: openapi3filter.NoopAuthenticationFunc, // No auth needed even when we have something in spec
+		},
+		ErrorHandler: func(w http.ResponseWriter, message string, statusCode int) {
```
mlguerrero12, line 47: Shouldn't we write a generic error handler function like this and here create getOranReqErrFunc and getOranRespErrFunc?

mlguerrero12 (commented):
Nice work @pixelsoccupied. I left some comments.

internal/service/alarms/api/openapi.yaml
pixelsoccupied: > I've seen this 404 for lists when filtering is used and there no matches. But we can also return an empty list with 200 OK for this case (with filtering)

Good point! Basically if no match found do we simply return an empty list instead of 404?  Will follow spec and remove 404.

internal/service/alarms/api/openapi.yaml
pixelsoccupied: I think decided that we dont support clearing (diverging from the spec). This endpoint will simply take the `alarmAcknowledged` for patch request. Do we think we should follow the spec but in code we just consider any other value as no-op?

internal/service/alarms/internal/middleware.go
pixelsoccupied: Sure moved it into func

internal/service/alarms/serve.go
pixelsoccupied: The project logger file seems to be unused and most of the project for example using `slog` where possible which i agree with. 

I dropped "log" and instead using "slog" everywhere

internal/service/alarms/serve.go
pixelsoccupied: updated

internal/service/alarms/serve.go
pixelsoccupied: yeah...accidentally removed the return `err`. Added it back and instead wrapped error with additional detail

internal/service/alarms/api/openapi.yaml
pixelsoccupied: Added

internal/service/alarms/api/openapi.yaml
pixelsoccupied: added 403. i thought we dont need 404 since we are returning empty list?

internal/service/alarms/api/openapi.yaml
pixelsoccupied: Yeah basically the idea is that `CreateSubscription` will generate a callback depending on `subscriber.callbackUrl` as seen below. 

The codegen gives us `AlarmEventNotification` go struct. We just need to create Post call using the callback url.

internal/service/alarms/api/openapi.yaml
pixelsoccupied: added

internal/service/alarms/api/openapi.yaml
mlguerrero12: I see. Didn't know only ack was going to be supported

internal/service/alarms/api/openapi.yaml
mlguerrero12: That was in case you wanted to leave 404 in the previous list

internal/service/alarms/api/openapi.yaml
mlguerrero12: I've seen these callbacks defined separately, so that we can share it with customers who are gonna generate the server code. From the same spec we could generate the client (and the go struct we need). My point is to have these things separated since they don't belong to the same server.

pixelsoccupied: As discussed on call...CreateSubscription is the "owner" of the callback.

PR #271, opened 2024-10-23
CNF-14959: Validate changes to schemas in the ClusterTemplate CR

irinamihai:


docs/cluster-configuration.md
```diff
@@ -216,6 +216,7 @@ For updating a manifest in an existing ACM PolicyGenerator, the following steps
     * Create a new version of the [clusterinstance-defaults-v2](samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/clusterinstance-defaults-v2.yaml) `ConfigMap` - [clusterinstance-defaults-v3](samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/clusterinstance-defaults-v3.yaml):
         * Update the name to `clusterinstance-defaults-v3` (the namespace stays `sno-ran-du-v4-Y-Z`).
         * Update the `sno-ran-du-policy` ManagedCluster `extraLabel` from `v1` to `v2`.
+        **Note:** An alternative to creating a new ConfigMap, is to change the label directly in the ProvisioningRequest by updating `spec.templateParameters.clusterInstanceParameters.extraLabels.ManagedCluster.sno-ran-du-policy` from `v1` to `v2`.
```
bartwensley, line 219: I don't think this is an alternative as it would require the SMO to do this - we only want the SMO to have to switch to the new cluster template version.

docs/cluster-configuration.md
```diff
@@ -343,3 +344,98 @@ policies:
...
+        * `metadata.name` is updated from `policytemplate-defaults-v1` to `policytemplate-defaults-v2`.
+        * update the defaults to reflect the new schema and thus the needed configuration values, in our case: `sriov-network-pfNames-1` and `sriov-network-pfNames-2`.
+    
+    * Create a new version of the [clusterinstance-defaults-v3](samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/clusterinstance-defaults-v3.yaml) `ConfigMap` - [clusterinstance-defaults-v4](samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/clusterinstance-defaults-v4.yaml):
+        * Update the name to `clusterinstance-defaults-v4` (the namespace stays `sno-ran-du-v4-Y-Z`).
+        * Update the `sno-ran-du-policy` ManagedCluster `extraLabel` from `v2` to `v3`.
+
+        **Note:** An alternative to creating a new ConfigMap, is to change the label directly in the ProvisioningRequest by updating `spec.templateParameters.clusterInstanceParameters.extraLabels.ManagedCluster.sno-ran-du-policy` from `v2` to `v3`.
```
bartwensley, line 402: I don't think we should mention this as an alternative - the SMO should only have to switch to the new cluster template version.

bartwensley (commented):
Looks good - just minor comments.

docs/cluster-configuration.md
irinamihai: Ok, understood. I'll remove it. Thank you.

docs/cluster-configuration.md
```diff
@@ -343,3 +343,97 @@ policies:
...
+
+The following steps need to be taken:
+1. Upversion the cluster template content:
+    * A new ACM PG is created - [sno-ran-du-pg-v4-Y-Z-v3](samples/git-setup/policytemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-pg-v4-Y-Z-v3.yaml):
+        * `metadata.name` is updated from `sno-ran-du-pg-v4-Y-Z-v2` to `sno-ran-du-pg-v4-Y-Z-v3` (the `ztp-sno-ran-du-v4-Y-Z` namespace is kept).
+        * `policyDefaults.placement.labelSelector.sno-ran-du-policy` is updated from `v2` to `v3` such that the policy binding is updated.
+        * All policy names are updated from `v2` to `v3` (example: `v2-subscriptions-policy` -> `v3-subscriptions-policy`).
+        * The following manifests are added under the `v3-subscriptions-policy`:
+        ```yaml
+        - path: source-crs/SriovNetworkNodePolicy-SetSelector.yaml
```
Missxiaoguo, line 362: I just noticed that the SriovNetworkNodePolicy CRs for du_fh and du_mh are missing in the initial PolicyGenerator (PG), but their SriovNetworks are included. Based on the legacy ZTP, they are all part of the DU profile. I think we should include both of them in the Day 0 PG, so that the VFs of the configured interfaces can be created and attached to the application.

For day2 schema update usecase, maybe we can add a new pair of SriovNetwork and SriovNetworkNodePolicy?

Missxiaoguo (commented):
/lgtm

PR #274, opened 2024-10-24
CNF-15056: Improve watches in the PR controller

irinamihai:
Description:
- Handle ClusterTemplate Create, Update and Delete events in the provisioningrequest controller.
- Trigger provisioningrequest reconciliation if such events come from a ClusterTemplate used by a ProvisioningRequest.
- For ClusterTemplate Create events, only process them further if the ClusterTemplate status has been filled in. If the status is empty it will not impact the ProvisioningRequest since we need a validated ClusterTemplate.
- Switch to using `handler.EnqueueRequestsFromMapFunc` for watching ClusterInstances, ClusterTemplates, Policies, NodePools, ManagedClusters.
- Watch ManagedClusters only if there's a change in their availability.
- Unrelated changes: remove some logs from the  `doesK8SResourceExist` and `CreateK8sCR` functions.

### DIFF
--- a/internal/controllers/provisioningrequest_setup.go
+++ b/internal/controllers/provisioningrequest_setup.go
@@ -7,7 +7,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


internal/controllers/provisioningrequest_setup.go
```diff
@@ -288,7 +289,7 @@ func (r *ProvisioningRequestReconciler) findManagedClusterForProvisioningRequest
 
 // findPoliciesForProvisioningRequests creates reconciliation requests for the ProvisioningRequests
 // whose associated ManagedClusters have matched policies Updated or Deleted.
-func findPoliciesForProvisioningRequests[T deleteOrUpdateEvent](
+func findPoliciesForProvisioningRequests[T deleteUpdateEvent](
```
Missxiaoguo, line 292: I wonder if we can use `EnqueueRequestsFromMapFunc` for policies as well since we only watch on delete and update events for policies.

PR #275, opened 2024-10-25
Upgrade schema and reconciler update

jc-rh:
This replaces #264

- Add release field to ClusterTemplate, sepcifying openshift release for the template
- Add upgradeDefaults to ClusterTemplate.Templates, specifying configmap name for upgrade information to this templater
- Add initital upgrade handler. After detecting that clusterTemplate.release is higher than deployed managedCluster.OpenshiftVersion, upgrade handler creates and ibgu to upgrade the cluster. The upgrade information is read from upgradeDefaults configmap specified in the clusterTemplate.
- Upgrade status reporting through conditions
- Upgrade failure / template version revert handling
- Controller setup update for IBGU monitoring
- Set owner reference if owner is cluster scoped

Doc and sample config udpate will be in a subsequent PR


Missxiaoguo (commented):
/lgtm

internal/controllers/provisioningrequest_upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+		return requeueWithError(fmt.Errorf("error getting IBGU: %w", err))
+	}
+
+	if isIBGUProgressing(ibgu) {
+		utils.SetStatusCondition(&t.object.Status.Conditions,
+			utils.PRconditionTypes.UpgradeCompleted,
+			utils.CRconditionReasons.InProgress,
+			metav1.ConditionFalse,
+			"Upgrade is in progress",
+		)
```
Missxiaoguo, line 111: I think `provisioningStatus.state` should be updated to progressing with a detailed message to align with the current process? This can be done in a follow-up PR.

jc-rh: Do we still need a separate upgrade condition then? I asked about this very early on and the sense I got was a separate condition was more preferred.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -79,8 +81,16 @@ func (t *provisioningRequestReconcilerTask) renderClusterInstanceTemplate(
...
 				var disallowedChanges []string
-				if len(updatedFields) != 0 {
-					disallowedChanges = append(disallowedChanges, updatedFields...)
+
+				for _, updatedField := range updatedFields {
+					// Add "AgentClusterInstall" to ClusterInstance.SuppressedManifests in order to
+					// prevent unnecessary updates to ACI.
+					if updatedField == "clusterImageSetNameRef" &&
+						crProvisionedCond.Reason == string(utils.CRconditionReasons.Completed) {
+						suppressedManifests = append(suppressedManifests, "AgentClusterInstall")
```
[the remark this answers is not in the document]
jc-rh: The template has to be usable in both install and upgrade cases. This is to prevent the provisioning failed state after updating the template ref to the new version.

internal/controllers/provisioningrequest_upgrade.go
Missxiaoguo: What I meant is just for the overall status, when upgrade condition is changed to InProgress, we would update the overall provisioningStatus to progressing as well, same for the completion.

internal/controllers/provisioningrequest_clusterinstall.go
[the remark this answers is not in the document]
Missxiaoguo: oh, okay, so for the upgrade case, clusterImageSetNameRef will need to be updated to the new version even if it's not actually used?

PR #282, opened 2024-10-29
Fix RBAC role for ImageBasedUpgrade

sudomakeinstall2:
This adds the missing RBAC role for provisioning request controller to access and modify IBGUs.


alegacy:
retested on my system.  it works now.

PR #285, opened 2024-10-29
Update gitops repo samples for upgrade related changes

jc-rh:


Missxiaoguo:
LGTM, except I’m not sure if we want to add a comment for `upgradeDefaults: upgrade-defaults-v1` indicating that it is optional and intended for upgrade use only, or will there be document to highlight/describe the field/files that should be prepared and added for upgrades?

Missxiaoguo:
Another sample to update (at least for required field)
https://github.com/openshift-kni/oran-o2ims/blob/main/config/samples/v1alpha1_clustertemplate.yaml

jc-rh:
> Another sample to update (at least for required field) https://github.com/openshift-kni/oran-o2ims/blob/main/config/samples/v1alpha1_clustertemplate.yaml

It's updated in this PR

jc-rh:
> LGTM, except I’m not sure if we want to add a comment for `upgradeDefaults: upgrade-defaults-v1` indicating that it is optional and intended for upgrade use only, or will there be document to highlight/describe the field/files that should be prepared and added for upgrades?

Good question. Not sure if it should be optional or mandatory. Will address this in the next doc PR once we agree on the seedgen and template update workflow.

bartwensley:
> > LGTM, except I’m not sure if we want to add a comment for `upgradeDefaults: upgrade-defaults-v1` indicating that it is optional and intended for upgrade use only, or will there be document to highlight/describe the field/files that should be prepared and added for upgrades?
> 
> Good question. Not sure if it should be optional or mandatory. Will address this in the next doc PR once we agree on the seedgen and template update workflow.

The upgradeDefaults needs to be optional. A simple example is the first release used to install an SNO will never be used in an upgrade. I think I would prefer not to add this to the cluster templates in the version_4.Y.Z directory, but instead create a new version4.Y.Z+1 directory as an example for the upgrade. But would like to hear what others think.

docs/samples/git-setup/clustertemplates/version_4.Y.Z/kustomization.yaml
```diff
@@ -11,6 +11,7 @@ resources:
 - sno-ran-du/ns.yaml
 - sno-ran-du/pull-secret.yaml
 - sno-ran-du/policytemplates-defaults-v1.yaml
+- sno-ran-du/policytemplates-defaults-v2.yaml
```
Missxiaoguo, line 14: policyTemplate defaults v2 is being used in template v4. It seems like sno-ran-du-v4-Y-Z-4.yaml and clusterinstance-defaults-v4.yaml are missing in kustomization.yaml too.

docs/samples/git-setup/clustertemplates/version_4.Y.Z+1/sno-ran-du/sno-ran-du-v4-Y-Z-1.yaml
```diff
@@ -0,0 +1,286 @@
+apiVersion: o2ims.provisioning.oran.org/v1alpha1
```
Missxiaoguo, line 1: Is this for seed cluster deployment? I wonder if it might be better not to mix the seed and target deployment files in the same repo.
Maybe @browsell can comment on this. Did we recommend separated repos for seed and target clusters for IBU?

jc-rh: It could be. In general, the upgradable template may not be the first version for each Z release though.
As regards to seedgen, my understanding is that the seedgen gitops setup should be a fresh copy of the target version of the real one. We can discuss whether we need to avoid upversioning the template just for upgrade defaults.

Missxiaoguo:
@jc-rh The clusterinstance schema and default configmap were updated in https://github.com/openshift-kni/oran-o2ims/pull/289, which was just merged last Friday. The new templates need to match the latest version now.

jc-rh:
> @jc-rh The clusterinstance schema and default configmap were updated in #289, which was just merged last Friday. The new templates need to match the latest version now.

Updated

docs/samples/git-setup/policytemplates/version_4.Y.Z+1/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+generators:
+- sno-ran-du/sno-ran-du-pg-v4-Y-Z-v1.yaml
```
bartwensley, line 2: Should this be "Z+1"?

jc-rh: Yes, updated

PR #287, opened 2024-10-29
NO-ISSUE: Update or remove hardcoded hwmgr references

donpenney:


Missxiaoguo:
When do we plan to remove the checks for `hwmgr` namespace as well (making the hw plugin mandatory)? I think with the hwmgr check, it's still possible to bypass the hw plugin?

donpenney:
> When do we plan to remove the checks for `hwmgr` namespace as well (making the hw plugin mandatory)? I think with the hwmgr check, it's still possible to bypass the hw plugin?

If they're not needed for unit tests, I assume they could be removed any time.

Missxiaoguo:
> 
> If they're not needed for unit tests, I assume they could be removed any time.

Okay, sounds good! Looks like we are ready to remove all workarounds (https://issues.redhat.com/browse/CNF-13704). I don't think the checks are needed for unittests. I will take a look.

PR #290, opened 2024-10-31
Add logic to fetch prometheus rules from clusters

mlguerrero12:
Prometheus rules are obtained from managed clusters to subsequently fill out the alarm dictionaries


internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+				wg.Done()
+				resultChannel <- result{
+					resourceTypeID: resource.id,
+					rules:          rules,
+					err:            err,
+				}
+			}()
+
+			switch resource.model {
+			case resourceTypeCluster:
```
alegacy, line 68: This will have to be revisited as the method used to determine if a resource type is a cluster will likely be more complex that just looking for the word "cluster" in the model.

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+	defer cancel()
+
+	var rules monitoringv1.PrometheusRuleList
+	err = clusterClient.List(ctxWithTimeout, &rules)
+	if err != nil {
+		return nil, fmt.Errorf("error listing rules for managed cluster %s: %w", cluster.Name, err)
+	}
+
+	if len(rules.Items) == 0 {
+		return nil, fmt.Errorf("no rules found for managed cluster %s", cluster.Name)
```
alegacy, line 155: Is this really an error or merely a possibility that there are no rules on this cluster?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+		r.RulesMap[res.resourceTypeID] = res.rules
+		slog.Info("loaded rules for resource type", "ResourceType ID", res.resourceTypeID, "rules count", len(res.rules))
+	}
+
+	// TODO: Load future alarm dictionary struct
+	// TODO: Load alarm dictionary into DB
+}
+
+// getRulesForResourceTypeCluster fetches Prometheus rules for a cluster resource type
+func (r *AlarmDictionary) getRulesForResourceTypeCluster(ctx context.Context, version string) ([]*monitoringv1.PrometheusRule, error) {
```
alegacy, line 96: Suggest `getRulesForCluster` is descriptive enough.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Yes, for sure

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: I don't think it is possible to delete the rules for core platform monitoring. So, if there are none for cluster type, then something is wrong. That's how I see it

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Ack. Thanks.

mlguerrero12: done

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+	// TODO: List Resource Type from Resource Server
+	resourceTypeList := getResourceTypes()
+
+	type result struct {
+		resourceTypeID int
+		rules          []*monitoringv1.PrometheusRule
+		err            error
+	}
+
+	// If there are many resource types, we can use a semaphore to limit the number of concurrent goroutines
```
pixelsoccupied, line 49: how are we limiting the concurrency? Shouldnt there be there something `semaphore := make(chan struct{}, maxConcurrency)`...which potentially makes wg redundant? 

Unless meant to just synchronize (in that case code is good but comment is not i think)?

internal/service/alarms/internal/dictionary/dictionary.go
pixelsoccupied: this should not be an error imo also. Though I would add some additional logs to understand the CR name, cluster name and maybe the count? Really anything that helps debug.

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, fmt.Errorf("no managed cluster found with version %s", version)
+	}
+
+	return &managedClusters.Items[0], nil
+}
+
+var getClientForCluster = clients.NewClientForCluster
+
+// getRules gets Prometheus rules from a managed cluster
+func (r *AlarmDictionary) getRules(ctx context.Context, cluster *clusterv1.ManagedCluster) (*monitoringv1.PrometheusRuleList, error) {
```
pixelsoccupied, line 139: I would update the name to say "getPromRules". Cause the actual rules (a.k.a alarms definition) embedded in the spec under .spec.groups[].rules`

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+	rules, err := r.getRules(ctx, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return rules.Items, nil
+}
+
+// getManagedCluster finds a single managed cluster with the given version
+func (r *AlarmDictionary) getManagedCluster(ctx context.Context, version string) (*clusterv1.ManagedCluster, error) {
```
pixelsoccupied, line 111: This will be integrated later yeah? Could put a "todo" comment to specify that?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+		r.RulesMap[res.resourceTypeID] = res.rules
+		slog.Info("loaded rules for resource type", "ResourceType ID", res.resourceTypeID, "rules count", len(res.rules))
+	}
+
+	// TODO: Load future alarm dictionary struct
+	// TODO: Load alarm dictionary into DB
+}
+
+// getRulesForCluster fetches Prometheus rules for a cluster resource type
+func (r *AlarmDictionary) getRulesForCluster(ctx context.Context, version string) ([]*monitoringv1.PrometheusRule, error) {
```
pixelsoccupied, line 96: would update to getPromRulesFromCluster

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+	err = clusterClient.List(ctxWithTimeout, &rules)
+	if err != nil {
+		return nil, fmt.Errorf("error listing rules for managed cluster %s: %w", cluster.Name, err)
+	}
+
+	if len(rules.Items) == 0 {
+		return nil, fmt.Errorf("no rules found for managed cluster %s", cluster.Name)
+	}
+
+	return &rules, nil
```
pixelsoccupied, line 158: Sinnce the actual rules are in the spec of Prom...could you add another function to extract them? Not really needed for this PR is focusing on alarmDict...but would add a "todo"

internal/service/alarms/internal/dictionary/suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestConfig(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Dictionary Suite")
+}
```
pixelsoccupied, line 13: Ultimately there will be only one suite (alarms_suite?) file all our tests? Sorry about the dumb q but forgot how ginkgo recommends structuring projects

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,187 @@
...
+	id      int
+	model   string
+	version string
+}
+
+// TODO: Replace with actual resource type list from Resource Server
+func getResourceTypes() []resourceType {
+	return []resourceType{
+		{
+			id:      1,
```
pixelsoccupied, line 172: for all IDs let's use UUID

internal/clients/clients.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	conf, err := clientcmd.RESTConfigFromKubeConfig(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config: %w", err)
+	}
+
+	c, err := client.New(conf, client.Options{Scheme: GetSchemeForCluster()})
```
pixelsoccupied, line 69: Instead of creating many clients for many kubeconfigs...is it possible to have multiple kubeconfing merged into one and thne switch context as needed? Client i think is also concurrency safe. https://pkg.go.dev/k8s.io/client-go@v0.26.3/tools/clientcmd

pixelsoccupied (commented):
Good stuff! Thanks!

internal/clients/clients.go
```diff
@@ -0,0 +1,108 @@
...
+	return c, nil
+}
+
+// GetSchemeForHub returns the scheme for the hub cluster client
+func GetSchemeForHub() *runtime.Scheme {
+	scheme := runtime.NewScheme()
+	utilruntime.Must(corev1.AddToScheme(scheme))
+	utilruntime.Must(clusterv1.AddToScheme(scheme))
+	utilruntime.Must(clusterv1beta1.AddToScheme(scheme))
+	utilruntime.Must(clusterv1beta2.AddToScheme(scheme))
```
pixelsoccupied, line 52: Something to note that we need to apply the RBACs to fetch the data (managedcluster stuff maybe not since we get admin kubeconfig)...depending on the service account. I think we can all share the same account that operators are using but something to think about.

pixelsoccupied: Since we are not ready to integrate/deploy to cluster yet it's not really important but maybe a todo to note this.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Alright. It is not up to us (alarm server) to interpret the presence or absence of rules for given resource type. I will just log (Debug level) it.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: That would be the case if we have a large number of resource types, but I don't expect that to be the case. The comment is more for completeness.

I see that the comment might be confusing, the idea is not to replace the wg with a buffered channel, but limit one specific section inside the routines.  I will delete the comment for now.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Ack, thanks.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: ack

internal/service/alarms/internal/dictionary/suite_test.go
mlguerrero12: Yes, at some point, I was thinking about running a simulated api server with envtest, that's why I put it here. Will change it.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: This is throw away code. Everything around resource type will be modified at some point.

internal/clients/clients.go
mlguerrero12: but not here, we could create a story and write that comment

internal/clients/clients.go
mlguerrero12: I don't really see the benefit of doing that. Besides, we won´t be accessing the clusters often. Also, we would need to cover scenarios such as the deletion of a managed cluster. And I'm using concurrency to get the rules. This would mean to stop to merge the configs and then continue.

This is simple and not expensive. It creates a client (not expensive in my opinion), get the stuff and goes out of scope so it can be collected by the GC.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Done

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Done

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Done

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Done

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: I ended up extracting the Rules.

internal/service/alarms/internal/dictionary/suite_test.go
mlguerrero12: Best practice is to have a test suite per package. I left it unchanged.

internal/clients/clients.go
```diff
@@ -0,0 +1,108 @@
...
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+)
+
+const (
+	ListRequestTimeout   = 30 * time.Second
+	SingleRequestTimeout = 10 * time.Second
+
+	secretTypeLabel      = "hive.openshift.io/secret-type"
+	secretTypeLabelValue = "kubeconfig"
```
pixelsoccupied, line 28: could add a comment about where these are coming from just to make it clear for the future? 

Also are the timeouts related to these two consts if not maybe create new const block (nit)?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,190 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift-kni/oran-o2ims/internal/clients"
+)
+
+const (
+	managedClusterVersionLabel = "openshiftVersion-major-minor"
+	localClusterLabel          = "local-cluster"
+
+	resourceTypeCluster = "Cluster"
```
pixelsoccupied, line 22: - nit: maybe a separate block since it's unrelated to acm values (this is o-ran specific stuff)? Also maybe a comment about what this is how we will have more types in the future?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Extract rules from PrometheusRule list
+	var rules []monitoringv1.Rule
+	for _, promRule := range promRules {
+		for _, group := range promRule.Spec.Groups {
+			rules = append(rules, group.Rules...)
+		}
+	}
+
+	return rules, nil
```
pixelsoccupied, line 114: maybe a log see say which cluster, the version and count?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// getManagedCluster finds a single managed cluster with the given version
+func (r *AlarmDictionary) getManagedCluster(ctx context.Context, version string) (*clusterv1.ManagedCluster, error) {
+	// Match managed cluster with the given version and not local cluster
+	selector := labels.NewSelector()
+	versionSelector, _ := labels.NewRequirement(managedClusterVersionLabel, selection.Equals, []string{version})
+	localClusterRequirement, _ := labels.NewRequirement(localClusterLabel, selection.NotEquals, []string{"true"})
```
pixelsoccupied, line 122: can we return the errors?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -0,0 +1,190 @@
...
+	promRules, err := r.getPromRules(ctx, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract rules from PrometheusRule list
+	var rules []monitoringv1.Rule
+	for _, promRule := range promRules {
+		for _, group := range promRule.Spec.Groups {
+			rules = append(rules, group.Rules...)
```
pixelsoccupied, line 110: Might be helpful to print out the alert names?

internal/clients/clients.go
pixelsoccupied: sounds good

internal/service/alarms/serve.go
```diff
@@ -32,7 +34,28 @@ const (
...
+
+	// Get client for hub
+	hubClient, err := clients.NewClientForHub()
+	if err != nil {
+		slog.Error("error creating client for hub", "error", err)
+		os.Exit(1)
+	}
+
+	alarmsDict := dictionary.New(hubClient)
+	alarmsDict.Load(ctx)
```
pixelsoccupied, line 58: Is there any clean up we need to do after this...like closing the client maybe?

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: The error happens when the rules for creating a label requirement constructor are violated. So, once you get this right, it won't fail. Unit tests are verifying this.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: let's do this in a future pr for filling out the dicts

internal/service/alarms/serve.go
mlguerrero12: Good question. I wondered the same, especially for the managed cluster clients. No, we don't have to close anything.

internal/service/alarms/internal/dictionary/dictionary.go
pixelsoccupied: just heads up when we upload this data into DB...it will reject anything that's not UUID. But we can update this in the follow up PR during DB integration.

PR #295, opened 2024-11-02
Fix suppressedManifests getting removed

sudomakeinstall2:
For upgrade process AgentClusterInstall should be added to
ClusterInstance's suppressedManifests in order to prevent updates to it.
This commit fixes suppressedManifests getting removed on subsequent
reconciles after adding on the first reconcile.


sudomakeinstall2:
on hold untill
https://github.com/openshift-kni/oran-o2ims/pull/288
is merged

PR #298, opened 2024-11-04
CNF-15491: Handle ProvisioningRequest deletion

Missxiaoguo:
Updates included:
- Add new provisioningState `deleting`, set when ProvisioningRequest is marked for deletion.
- Add the finializer back to handle dependents deletion, ensuring dependents are removed before removing finalizer from ProvisioningRequest to complete the deletion.
- Watch for deletion of cluster namespace so reconcilation can be triggered when ns is deleted.

### DIFF
--- a/internal/controllers/utils/conditions.go
+++ b/internal/controllers/utils/conditions.go
@@ -110,6 +110,12 @@ func SetProvisioningStateFulfilled(cr *provisioningv1alpha1.ProvisioningRequest)
 	cr.Status.ProvisioningStatus.ProvisioningDetails = "Provisioning request has completed successfully"
 }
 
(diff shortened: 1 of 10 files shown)


internal/controllers/utils/conditions.go
```diff
@@ -110,6 +110,12 @@ func SetProvisioningStateFulfilled(cr *provisioningv1alpha1.ProvisioningRequest)
 	cr.Status.ProvisioningStatus.ProvisioningDetails = "Provisioning request has completed successfully"
 }
 
+// SetProvisioningStateDeleting updates the provisioning state to failed with detailed message
```
bartwensley, line 113: failed -> deleting

Missxiaoguo: Ah...

PR #302, opened 2024-11-06
Adds missing mandatory AlarmDefinition attributes

alosadagrande:
These fields are missing as part of the AlarmDefinition table (Table 3.2.6.2.9‑1 Definition of type AlarmDefinition) in the O-RAN WG6 O2-IMS INTERFACE document. They're mandatory fields that needs to be sent back to pass the conformance tests.

cc/ @pixelsoccupied @alegacy

### DIFF
--- a/internal/service/alarm_definition_handler_test.go
+++ b/internal/service/alarm_definition_handler_test.go
@@ -78,6 +78,9 @@ var _ = Describe("Alarm definition handler", func() {
 					Expect(item).To(HaveKey("managementInterfaceId"))
 					Expect(item).To(HaveKey("pkNotificationField"))
 					Expect(item).To(HaveKey("alarmAdditionalFields"))
+					Expect(item).To(HaveKey("alarmLastChange"))
+					Expect(item).To(HaveKey("alarmChangeType"))
+					Expect(item).To(HaveKey("clearingType"))
 				}
 			})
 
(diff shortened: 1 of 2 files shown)


internal/service/alarm_definition_handler_test.go
alosadagrande, line 1: @pixelsoccupied not sure if they also need to be added to the test handler.

internal/service/alarm_definition_handler_test.go
```diff
@@ -78,6 +78,9 @@ var _ = Describe("Alarm definition handler", func() {
 					Expect(item).To(HaveKey("managementInterfaceId"))
 					Expect(item).To(HaveKey("pkNotificationField"))
 					Expect(item).To(HaveKey("alarmAdditionalFields"))
+					Expect(item).To(HaveKey("alarmLastChange"))
```
alosadagrande, line 81: @pixelsoccupied not sure if they also need to be added to the test handler.

bartwensley:
Original cc used wrong syntax...
/cc @alegacy 
/cc @pixelsoccupied

internal/service/alarm_definition_handler_test.go
pixelsoccupied: @alosadagrande you can consider all the current code Alarm server as deprecated. so feel free to make the minimal changes you need for your tests to pass (if the tests dont need to be updated to pass CI feel free to revert it)

alosadagrande: ok, i'll leave it like that since it passes my tests :)

PR #303, opened 2024-11-06
MGMT-18997: Use downward API to acquire namespace

alegacy:
Rather than get the namespace from the service account static files we are able to get it from an environment variable.  This benefits being able to run the controller from the command line for testing purposes.

### DIFF
--- a/bundle/manifests/oran-o2ims.clusterserviceversion.yaml
+++ b/bundle/manifests/oran-o2ims.clusterserviceversion.yaml
@@ -1397,6 +1397,10 @@ spec:
                   value: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
                 - name: HWMGR_PLUGIN_NAMESPACE
                   value: oran-hwmgr-plugin
+                - name: OCLOUD_MANAGER_NAMESPACE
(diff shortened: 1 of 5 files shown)


mlguerrero12:
I think you need to update the clusterversion of the bundle `make bundle`

alegacy:
> I think you need to update the clusterversion of the bundle `make bundle`

yes, of course.  Sorry... hit commit too quickly!

mlguerrero12:
/lgtm


Question outside the scope of this pr: 

Why is the manager watching resources in all namespaces? What you are introducing on this pr is a common pattern to narrow the scope of an operator https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-watch-namespaces-dynamically.

alegacy:
> Why is the manager watching resources in all namespaces? What you are introducing on this pr is a common pattern to narrow the scope of an operator https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-watch-namespaces-dynamically.

That's a good question.  I don't have the answer/history on this... my assumption is that we are watching other cluster resources on the provisioning side of things that do not live in our namespace.

PR #304, opened 2024-11-06
Remove NodePool Status Updates from Cloud Manager

tliu2021:
This update removes node pool status updates from the Cloud Manager to maintain a clear separation of responsibilities: the Cloud Manager exclusively updates spec fields, while the hwMgr plugin is responsible for status fields.

The hwMgr plugin uses the ObservedGeneration status field to detect changes in the node pool spec and adjust its operations accordingly, without needing to modify spec fields. This design strengthens single-responsibility principles, reducing potential update conflicts.


donpenney (commented):
/lgtm

PR #306, opened 2024-11-08
CNF-14862: Configure AlertManager to send alerts

mlguerrero12:
Add a webhook receiver to AlertManager configuration so that alarm server is notified to the internal endpoint.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -408,7 +408,15 @@ sync-submodules:
 ##@ O-RAN Alarms Server
 
 .PHNOY: alarms
-alarms: clean-postgres run-postgres run-alarms-migrate run-alarms ##Run full alarms stack
+alarms: clean-postgres clean-am-service run-postgres run-alarms-migrate create-am-service run-alarms ##Run full alarms stack
+
+create-am-service: ##Creates alarm manager service and endpoint to expose a DNS entry.
+	oc apply -k ./internal/service/alarms/k8s/base --wait=true
+	@echo "Service and Endpoint for alarm manager created."
+
+clean-am-service: ##Deletes alarm manager service and endpoint.
+	-oc delete -k ./internal/service/alarms/k8s/base --wait=true
+	@echo "Service and Endpoint for alarm manager deleted."
 
 .PHONY: run-alarms
 run-alarms: go-generate binary ##Run alarms server locally
--- a/internal/service/alarms/internal/alertmanager/alertmanager.go
+++ b/internal/service/alarms/internal/alertmanager/alertmanager.go
@@ -1,0 +1,42 @@
+package alertmanager
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	namespace  = "open-cluster-management-observability"
+	secretName = "alertmanager-config"
+	secretKey  = "alertmanager.yaml"
+)
+
+//go:embed alertmanager.yaml
+var alertManagerConfig []byte
+
+// Setup updates the alertmanager config secret with the new configuration
+func Setup(ctx context.Context, cl client.Client) error {
+	// ACM recreates the secret when it is deleted, so we can safely assume it exists
+	var secret corev1.Secret
+	err := cl.Get(ctx, client.ObjectKey{Namespace: namespace, Name: secretName}, &secret)
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s/%s: %w", namespace, secretName, err)
+	}
+
+	// Verify that Data has the key "alertmanager.yaml"
+	if _, ok := secret.Data[secretKey]; !ok {
+		return fmt.Errorf("%s not found in secret %s/%s", secretKey, namespace, secretName)
+	}
+
+	secret.Data[secretKey] = alertManagerConfig
+	err = cl.Update(ctx, &secret)
+	if err != nil {
+		return fmt.Errorf("failed to update secret %s/%s: %w", namespace, secretName, err)
+	}
+
+	return nil
+}
--- a/internal/service/alarms/internal/alertmanager/alertmanager.yaml
+++ b/internal/service/alarms/internal/alertmanager/alertmanager.yaml
@@ -1,0 +1,30 @@
+---
+global:
+  resolve_timeout: 5m
+route:
+  group_wait: 30s
+  group_interval: 1m
+  repeat_interval: 4h
+  # Default receiver - Alerts without cluster identifier label will still be sent to this receiver
+  receiver: oran_alarm_receiver
+  routes:
+  # Route to skip alerts
+  - receiver: "null"
+    matchers:
+      # Always firing alert to verify alertmanager is working
+      - Watchdog
+  - receiver: oran_alarm_receiver
+    group_by:
+      # This label is guaranteed to be present in all alerts
+      - managed_cluster
+    matchers:
+      # Match alerts with a non-empty managed_cluster label
+      - managed_cluster=~".+"
+receivers:
+  - name: oran_alarm_receiver
+    webhook_configs:
+    - send_resolved: true
+      # Hardcoded 8080 port (same as http server)
+      # TODO: delete port from url once we have a proper k8s service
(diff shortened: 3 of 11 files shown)


internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,147 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	namespace  = "open-cluster-management-observability"
+	secretName = "alertmanager-config"
+	secretKey  = "alertmanager.yaml"
+
+	URL          = "http://alarm-server.oran-o2ims.svc.cluster.local:%s/internal/v1/caas-alerts/alertmanager"
+	templateFile = "./internal/service/alarms/internal/alertmanager/alertmanager.tmpl"
```
alegacy, line 19: This will only work when running locally as a standalone binary.   This would have to be packaged into our image.

mlguerrero12: Indeed. I changed this part to make it work for both

internal/service/alarms/serve.go
```diff
@@ -60,6 +60,15 @@ func Serve() error {
 	if err != nil {
 		return fmt.Errorf("error creating client for hub: %w", err)
 	}
+
+	// We need port for now because we are using a service with clusterIP None and an endpoint to expose a DNS entry.
+	// These resources are created within the kustomize folder.
+	// TODO: Remove port once application is run within a pod
+	err = alertmanager.Setup(ctx, hubClient, port)
```
pixelsoccupied, line 67: No port simply means `80` no?

Makefile
```diff
@@ -421,15 +429,17 @@ run-alarms-migrate: binary ##Migrate all the way up
 
 .PHONY: run-postgres
 run-postgres: ##Run O-RAN postgres
-	oc apply -k ./internal/service/postgres/k8s/overlays/dev
+	oc apply -k ./internal/service/kustomize/postgres/overlays/dev
```
pixelsoccupied, line 432: can we revert back to the original? I think the idea is that /pg is service and eventually the content under /k8s will be replaced with Go code that's called from the operator

internal/service/kustomize/alertmanager/README.md
```diff
@@ -0,0 +1,10 @@
+# Service for Alertmanager
```
pixelsoccupied, line 1: can we move this under `/alarms/k8s` to assert that these only relevant for  alarms service? (and eventually replace them with actual Go code as needed when integrating with operator)

pixelsoccupied (commented):
Looks good!

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	routes = append(routes, r)
+
+	// Routes to group alerts by cluster identifier
+	for _, label := range clusterIDLabels {
+		r = RouteConfig{Receiver: oranAlarmReceiver}
+		r.GroupBy = append(r.GroupBy, label)
+		r.Matchers = append(r.Matchers, fmt.Sprintf("%s=~\".+\"", label))
+		routes = append(routes, r)
+	}
```
pixelsoccupied, line 122: Is there chance we may miss alerts if we start adding filter at this stage? I think we should accept all the alerts now and then start adding filter in the subsequent PRs?

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: Thanks. I'll have a look at it

internal/service/alarms/internal/alertmanager/alertmanager.tmpl
```diff
@@ -0,0 +1,33 @@
+---
+global:
+  resolve_timeout: {{ .ResolveTimeout }}
+route:
+  group_wait: {{ .Route.GroupWait }}
+  group_interval: {{ .Route.GroupInterval }}
+  repeat_interval: {{ .Route.RepeatInterval }}
+  receiver: "{{ .Route.Receiver }}"
```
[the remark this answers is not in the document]
mlguerrero12: Not even the URL. Perhaps it is better to hardcode everything. I'll think about it

internal/service/alarms/serve.go
mlguerrero12: yes

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: These will group by labels that identify the cluster. If these labels do not exist, the alert is sent anyway due to the parent receiver.  So, we won't miss anything.

pixelsoccupied: Gotcha! All good as long as we dont miss anything. 

But just a heads up - i think we are guaranteed to get `managed_cluster_id` and the alert name for all the notifications. If that's not the case we might need to reach out to acm folks right? @browsell

mlguerrero12: I've seen alerts without the label `managed_cluster`. They had instead the label `clusterID`. That's why I'm creating a group for it. I will report it to see if it is a bug.

browsell: Are those from the local cluster ?

mlguerrero12: don't remember exactly but alerts coming from the local cluster also come with the `managed_cluster` label

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: updated

internal/service/alarms/internal/alertmanager/alertmanager.tmpl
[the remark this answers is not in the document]
mlguerrero12: updated

Makefile
mlguerrero12: done

internal/service/kustomize/alertmanager/README.md
mlguerrero12: done

internal/service/alarms/k8s/base/endpoints.yaml
```diff
@@ -0,0 +1,9 @@
+---
+kind: Endpoints
+apiVersion: v1
+metadata:
+  name: alarm-server
+subsets:
+  - addresses:
+      # Change me
+      - ip: 10.6.34.87
```
pixelsoccupied, line 9: let's not checkin internal IPs...maybe just put a generic IP e.g "0.0.0.0" ?

internal/service/alarms/internal/alertmanager/alertmanager.yaml
```diff
@@ -0,0 +1,30 @@
...
+  group_interval: 1m
+  repeat_interval: 4h
+  # Default receiver - Alerts without cluster identifier label will still be sent to this receiver
+  receiver: oran_alarm_receiver
+  routes:
+  # Route to skip alerts
+  - receiver: "null"
+    matchers:
+      # Always firing alert to verify alertmanager is working
+      - Watchdog
```
pixelsoccupied, line 15: I think skipping `Watchdog` is fine but just want to make sure @browsell is aware of this since we initially sort of agreed to collect all the alerts regardless of it being a noise.

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+const (
+	namespace  = "open-cluster-management-observability"
+	secretName = "alertmanager-config"
+	secretKey  = "alertmanager.yaml"
+)
+
+//go:embed alertmanager.yaml
+var alertManagerConfig []byte
```
pixelsoccupied, line 19: looks clean!

pixelsoccupied (commented):
I think we are ready to merge! just a few tiny comments

pixelsoccupied:
/approve

will make any follow up changes as needed to AM config but I think we can get started with building on top of this!

PR #308, opened 2024-11-12
MGMT-18997: Get default domain from status rather than spec

alegacy:
This changes how we acquire the default domain so that we look at the ingress controller's Status field rather than its Spec field since in some cases it may not be explicitly set in the Spec which causes it to go get it from another source.  Ultimately, we do not care where it gets it from so we look at its final decision which is stored in the Status field.

### DIFF
--- a/internal/controllers/utils/utils.go
+++ b/internal/controllers/utils/utils.go
@@ -240,14 +240,14 @@ func GetIngressDomain(ctx context.Context, c client.Client) (string, error) {
 		return "", fmt.Errorf("default ingress controller object not found: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controllers/utils/utils.go
```diff
@@ -240,14 +240,14 @@ func GetIngressDomain(ctx context.Context, c client.Client) (string, error) {
 		return "", fmt.Errorf("default ingress controller object not found: %w", err)
 	}
 
-	spec := ingressController.Object["spec"].(map[string]interface{})
+	spec := ingressController.Object["status"].(map[string]interface{})
```
pixelsoccupied, line 243: `spec :=` should be `status :=`?

PR #309, opened 2024-11-12
CNF-15608: Support cluster provisioning without hwTemplate

Missxiaoguo:
In the case where hwTemplate is not provided, hw provisioning is skipped and only cluster installation and configuration will be performed.  The hardware-related parameters (such as bmcAddress, bmcCreds - username and password are all base64-encoded strings, bootMacAddress, nodeNetwork macAddress) should instead be passed in the ProvisioningRequest. The ClusterTemplate schema should include these parameters, allowing them to be specified in the ProvisioningRequest.

Updates:
- change `templates.hwTemplate` to be optional
- skip hw provisioning when hwTemplate is not provided
- create BMC secret when hwTemplate is not provided
- add example ClusterTemplate without hwTemplate

Todo:
Consider validating the schema to ensure that the hardware-related parameters are present and marked as required.

E2E testing has been done for both with and without hwtemplate cases.

### DIFF
--- a/docs/cluster-provisioning.md
+++ b/docs/cluster-provisioning.md
@@ -31,7 +31,7 @@ The [CRD](../config/crd/bases/o2ims.provisioning.oran.org_clustertemplates.yaml)
 - version: Defines the version of the ClusterTemplate.
 - description: A description of the ClusterTemplate.
 - templates:
-    - hwTemplate: References the ConfigMap containing the hardware template used for node allocation.
+    - hwTemplate: (Optional) References the ConfigMap containing the hardware template used for node allocation. See note below.
     - clusterInstanceDefaults: References the ConfigMap containing default values for ClusterInstance.
     - policyTemplateDefaults: References the ConfigMap containing default values for ACM policy templates.
 - templateParameterSchema: Specifies the OpenAPI v3 schema that defines the accepted and required parameters for provisioning a cluster. This schema is used to validate the parameters passed in the ProvisioningRequest.
@@ -62,6 +62,10 @@ status:
     status: "True"
     type: ClusterTemplateValidated
 ```
+
+> [!NOTE]
+>  `spec.templates.hwTemplate` is optional. In scenarios where the hwTemplate ConfigMap is not provided, hardware provisioning will not be performed, and hardware-related parameters for each node
+> (e.g, bmcAddress, bmcCredentialsDetails, bootMACAddress, nodeNetwork.interfaces[*].macAddress) should be specified in the ProvisioningRequest. See this [example](samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml) of a ClusterTemplate without `hwTemplate`.
 
 ## ProvisioningRequest CR
 
--- a/docs/samples/git-setup/clustertemplates/version_4.Y.Z/kustomization.yaml
+++ b/docs/samples/git-setup/clustertemplates/version_4.Y.Z/kustomization.yaml
@@ -24,3 +24,5 @@ resources:
 # sno-ran-du.v4-Y-Z-4 ClusterTemplate:
 - sno-ran-du/sno-ran-du-v4-Y-Z-4.yaml
 - sno-ran-du/clusterinstance-defaults-v4.yaml
+# sno-ran-du.v4-Y-Z-1-no-hwtemplate ClusterTemplate:
+- sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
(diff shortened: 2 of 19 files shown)


docs/cluster-provisioning.md
```diff
@@ -63,6 +63,9 @@ status:
     type: ClusterTemplateValidated
 ```
 
+> [!NOTE]
```
bartwensley, line 66: You could also mark the hwTemplate (line 34 above) as optional and say "see note below" or something like that.

docs/samples/git-setup/clustertemplates/version_4.Y.Z/kustomization.yaml
```diff
@@ -15,6 +15,8 @@ resources:
 # sno-ran-du.v4-Y-Z-1 ClusterTemplate:
 - sno-ran-du/sno-ran-du-v4-Y-Z-1.yaml
 - sno-ran-du/clusterinstance-defaults-v1.yaml
+# sno-ran-du.v4-Y-Z-1-no-hwtemplate ClusterTemplate:
+- sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
```
bartwensley, line 19: nit: Would it be better to put this at the end of the list as it really is separate from the other example templates?

docs/samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
```diff
@@ -0,0 +1,257 @@
...
+            type: string
+          hugepages-count:
+            type: string
+          install-plan-approval:
+            type: string
+        type: object
+      #####################################################################################
+      # In scenarios where hardware provisioning is not required, the provisioningRequest
+      # must provide additional hardware-related parameters to support cluster installation.
+      # The following fields should be added to the schema for ProvisioningRequest,
```
bartwensley, line 42: Maybe "added to the clusterInstanceParameters schema, as shown in the example"?

docs/samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
```diff
@@ -0,0 +1,257 @@
...
+      # as shown in the example:
+      # - bmcAddress
+      # - bmcCredentialsName
+      # - bmcCredentialsDetails
+      # - bootMACAddress
+      # - nodeNetwork.interfaces[*].macAddress
```
bartwensley, line 48: Should you mention that these are all under the "nodes" section?

docs/samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
```diff
@@ -0,0 +1,257 @@
...
+                  type: string
+                bmcCredentialsName:
+                  description: BmcCredentialsName is the name of the secret containing
+                    the BMC credentials (requires keys "username" and "password").
+                  properties:
+                    name:
+                      type: string
+                  required:
+                  - name
+                  type: object
```
bartwensley, line 133: I think we're requiring the ProvisioningRequest to have the username/password - why do we need the secret here as well? Ahh... this will be used to provide a name for the created secret. Hmm... could we auto-generate the secret name (e.g. based on the cluster name and node name)? Would be simpler for the user and avoid errors where the user attempts to use the same secret name for multiple nodes.

donpenney: The ProvisioningRequest wouldn't have the BMC username/password, afaik. That would ultimately come from the hardware manager when nodes are allocated, and the plugin would create a corresponding secret

browsell: In this use case, there is no H/W manager.

docs/samples/git-setup/clustertemplates/version_4.Y.Z/kustomization.yaml
browsell: +1

browsell:
LGTM with the comments addressed

docs/samples/git-setup/clustertemplates/version_4.Y.Z/sno-ran-du/sno-ran-du-v4-Y-Z-1-no-hwtemplate.yaml
Missxiaoguo: `bmcCredentialsName` is optional field. Yes, currently, we auto-generate the secret name as <nodename>-bmc-secret if secret name is not provided. Right, it seems unnecessary since we’re auto-generating it anyway. I don't see a reason why users would need a different secret name 🤔 .

Missxiaoguo: Excluded `bmcCredentialsName` parameter from schema.

PR #310, opened 2024-11-13
Consolidate inventory subscription server into resource server

alegacy:
This change consolidates the inventory subscription server into the resource server so that both functions are handled by the same server. This is to facilitate future changes to the persistence strategy that will be simpler if both of these functions are colocated.


internal/cmd/server/start_resource_server.go
```diff
@@ -173,6 +189,31 @@ func (c *ResourceServerCommand) run(cmd *cobra.Command, argv []string) error {
...
+	if err != nil {
+		c.logger.ErrorContext(
+			ctx,
+			"Failed to get global cloud identifier flag",
+			"flag", globalCloudIDFlagName,
+			"error", err.Error(),
+		)
+		return exit.Error(1)
+	}
+	if globalCloudID == "" {
```
pixelsoccupied, line 203: Beyond the scope of this PR but just a heads up there maybe couple of cobra features we can leverage to reduce boilerplate code. 

There's a convenient way make a flag [required](https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#required-flags) by cobra so it auto errors out. There are also ways to allow for default values + automatically fill out vars for us (so we dont need to be call flags.GetString explicitly).

pixelsoccupied (commented):
/approve

internal/cmd/server/start_resource_server.go
alegacy: Thanks, ya... i'm aware of some of those features and hope that we incorporate them to simplify the code!

PR #314, opened 2024-11-16
CNF-15660: Filter out policies outside the ztp-<clustertemplate-ns> namespace

Missxiaoguo:
The controller should only handle/consider the child policies whose parent policies are from the `ztp-<clustertemplate-ns> `namespace. Other bound policies should neither trigger the reconcilation nor be considered in the policy configuration process.

### DIFF
--- a/internal/controllers/provisioningrequest_controller_test.go
+++ b/internal/controllers/provisioningrequest_controller_test.go
@@ -1052,10 +1052,10 @@ var _ = Describe("ProvisioningRequestReconcile", func() {
 			// Create Non-compliant enforce policy
 			policy = &policiesv1.Policy{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "ztp-clustertemplate-a-v4-15.v1-sriov-configuration-policy",
+					Name:      "ztp-clustertemplate-a-v4-16.v1-sriov-configuration-policy",
 					Namespace: "cluster-1",
 					Labels: map[string]string{
-						utils.ChildPolicyRootPolicyLabel:       "ztp-clustertemplate-a-v4-15.v1-sriov-configuration-policy",
+						utils.ChildPolicyRootPolicyLabel:       "ztp-clustertemplate-a-v4-16.v1-sriov-configuration-policy",
 						utils.ChildPolicyClusterNameLabel:      "cluster-1",
 						utils.ChildPolicyClusterNamespaceLabel: "cluster-1",
 					},
--- a/internal/controllers/provisioningrequest_setup.go
+++ b/internal/controllers/provisioningrequest_setup.go
@@ -251,12 +251,42 @@ func (r *ProvisioningRequestReconciler) enqueueProvisioningRequestForPolicy(
 	// ManagedCluster to which the policy is matched.
 	provisioningRequest, okCR := clusterInstance.GetLabels()[provisioningRequestNameLabel]
 	if okCR {
-		r.Logger.Info(
-			"[enqueueProvisioningRequestForPolicy] Add new reconcile request for ProvisioningRequest ",
-			"name", provisioningRequest)
-		requests = append(requests, reconcile.Request{
-			NamespacedName: types.NamespacedName{Name: provisioningRequest},
-		})
+		provReq := &provisioningv1alpha1.ProvisioningRequest{}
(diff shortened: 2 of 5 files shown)


internal/controllers/provisioningrequest_controller_test.go
```diff
@@ -1052,10 +1052,10 @@ var _ = Describe("ProvisioningRequestReconcile", func() {
 			// Create Non-compliant enforce policy
 			policy = &policiesv1.Policy{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "ztp-clustertemplate-a-v4-15.v1-sriov-configuration-policy",
+					Name:      "ztp-clustertemplate-a-v4-16.v1-sriov-configuration-policy",
```
irinamihai, line 1055: In the spirit of our examples, maybe these should be actually renamed to `v4.Y`

internal/controllers/provisioningrequest_setup.go
```diff
@@ -251,12 +251,42 @@ func (r *ProvisioningRequestReconciler) enqueueProvisioningRequestForPolicy(
...
+				// The provisioning request could have been deleted
+				return nil
+			}
+			r.Logger.Error("[enqueueProvisioningRequestForPolicy] Error getting ProvisioningRequest. ", "Error: ", err)
+			return nil
+		}
+
+		clusterTemplates := &provisioningv1alpha1.ClusterTemplateList{}
+		if err := r.List(ctx, clusterTemplates); err != nil {
+			r.Logger.Error("[enqueueProvisioningRequestForPolicy] Error listing ClusterTemplate. ", "Error: ", err)
```
irinamihai, line 266: `Error listing ClusterTemplates`?

internal/controllers/provisioningrequest_setup.go
```diff
@@ -251,12 +251,42 @@ func (r *ProvisioningRequestReconciler) enqueueProvisioningRequestForPolicy(
...
+			return nil
+		}
+
+		ctRefName := getClusterTemplateRefName(
+			provReq.Spec.TemplateName, provReq.Spec.TemplateVersion)
+		ctRefNamespace := ""
+		for _, ct := range clusterTemplates.Items {
+			if ctRefName == ct.Name {
+				// Break if found, as the metadata name of Clustertemplate is unique across all namespaces.
```
irinamihai, line 275: `ClusterTemplate`

internal/controllers/utils/provision.go
```diff
@@ -360,3 +360,15 @@ func ValidateDefaultInterfaces[T any](data T) error {
 	}
 	return nil
 }
+
+func GetParentPolicyNameAndNamespace(childPolicyName string) (policyName, policyNamespace string) {
```
irinamihai, line 364: To maintain consistency, this function needs a description above.

internal/controllers/utils/provision.go
```diff
@@ -360,3 +360,15 @@ func ValidateDefaultInterfaces[T any](data T) error {
...
+	// The format of a child policy name is parent_policy_namespace.parent_policy_name.
+	// Extract the parent policy name and namespace by splitting the child policy name into two substrings separated by "."
+	// and we are safe to split with the separator "." as the namespace is disallowed to contain "."
+	res := strings.SplitN(childPolicyName, ".", 2)
+	return res[1], res[0]
+}
+
+func IsParentPolicyInZtpClusterTemplateNs(policyNamespace, ctNamespace string) bool {
```
irinamihai, line 372: Same comment as above.

irinamihai (commented):
Just some nits.
Looks good to me otherwise.

internal/controllers/utils/provision.go
Missxiaoguo: This function is small and its name clearly conveys its purpose. I didn’t think adding a description was necessary, but I can add one - no big deal.

internal/controllers/utils/provision.go
Missxiaoguo: Same as above.

internal/controllers/provisioningrequest_controller_test.go
Missxiaoguo: An exact match would be `v4.Y.Z`.  It might be good to have but is not a must-have. It would be a seperate PR to update all testcases if needed.

PR #320, opened 2024-11-19
CNF-15661: Include inform policies in configuration timeout calculation

Missxiaoguo:
For the ORAN use case, inform policies should only be used to validate cluster configurations. When all enforce and inform policies are compliant, the cluster configuration is considered complete, and provisioning can be fulfilled. Updated logic now includes inform policies in the configuration timeout calculation.


Missxiaoguo:
/hold
Try to finish e2e testing after fixing the hub cluster issue.

Missxiaoguo:
/unhold
Tested e2e for both timeout and non-timeout cases.

PR #333, opened 2024-11-20
Get rules for hub cluster

mlguerrero12:
Add logic to fetch the rules of the hub cluster

### DIFF
--- a/internal/service/alarms/internal/dictionary/dictionary.go
+++ b/internal/service/alarms/internal/dictionary/dictionary.go
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	cr_client "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/k8s_client"
 )
@@ -21,15 +21,18 @@ const (
 )
 
 // TODO: create a file with the resource type names once they are defined by the Resource Server
-const resourceTypeCluster = "Cluster"
+const (
+	resourceTypeCluster = "Cluster"
+	resourceTypeHub     = "Hub"
+)
 
 type AlarmDictionary struct {
-	Client client.Client
+	Client cr_client.Client
 
 	RulesMap map[int][]monitoringv1.Rule
 }
 
-func New(client client.Client) *AlarmDictionary {
+func New(client cr_client.Client) *AlarmDictionary {
 	return &AlarmDictionary{
 		Client:   client,
 		RulesMap: make(map[int][]monitoringv1.Rule),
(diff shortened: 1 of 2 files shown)


mlguerrero12:
@pixelsoccupied, @alegacy, @browsell, please have a look when you have time

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -189,5 +205,9 @@ func getResourceTypes() []resourceType {
 			model:   resourceTypeCluster,
 			version: "4.16",
 		},
+		{
+			id:    4,
+			model: resourceTypeHub,
+		},
```
pixelsoccupied, line 211: I think is good but will not useful for now? To start with we are only just looking at managed clusters/SNO right...i.e specifically ignoring hub alerts? (we have mentioned this in design docs/epic i think)

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -10,7 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	cr_client "sigs.k8s.io/controller-runtime/pkg/client"
```
pixelsoccupied, line 13: how about `k8s_client`?

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -21,15 +21,18 @@ const (
...
 
 type AlarmDictionary struct {
-	Client client.Client
+	Client cr_client.Client
 
 	RulesMap map[int][]monitoringv1.Rule
 }
 
-func New(client client.Client) *AlarmDictionary {
+func New(client cr_client.Client) *AlarmDictionary {
```
pixelsoccupied, line 35: I think we should send back these errors to the init code? If anything of this fails we shouldnt start the server

pixelsoccupied (commented):
Just a few comments but looks good otherwise!  Just a heads up...we tackling hub alerts at a later phase

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: there is already a k8s_client. This is fine I think

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: No, I don't think so. For some reason the collection of rules for a resource type can fail but that doesn't mean we need to stop. The server can still work. DB can have data already. What I think is we should implement a resync mechanism so that I tries to populate the dict again.

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: Sure

mlguerrero12: done

internal/service/alarms/internal/dictionary/dictionary.go
mlguerrero12: There might be a resource type for which there are no managed clusters. Maybe the resource server is not in sync. An error will be returned but we can still continue. I think this section is not critical for the alarm server to continue or not. In my opinion, we should simply log and continue.

This is outside the scope of this PR. We need too continue this discussion for sure

pixelsoccupied: > An error will be returned but we can still continue

Yeah this is not true for all error right? yeah sounds let's move to another pr as needed.

PR #334, opened 2024-11-20
CNF-15750: Add DB schema for resource server

alegacy:
This creates an initial baseline schema for the resource server's inventory based objects. The objects will be managed by the resource server as an intermediate storage between the underlying data sources and the API front-end.


internal/service/alarms/migration.go
```diff
@@ -0,0 +1,26 @@
+package alarms
```
pixelsoccupied, line 1: ah good call...flows much nicer by bringing this out to serve.go level. mapping the two commands to the two main files!

pixelsoccupied (commented):
/lgtm

PR #336, opened 2024-11-22
CNF-14955: Adding postgres resources to controller

alegacy:
This adds the ability to manage the Kubernetes resources related to the Postgres database.  This is a first iteration which does not include the management of the PV/PVC for the persistent data.  It handles only the establishment of the deployment and the user credentials.  PV/PVC handling will come in a later commit.


api/inventory/v1alpha1/inventory_types.go
```diff
@@ -138,6 +138,11 @@ type InventorySpec struct {
 	// KubeRbacProxyImage is the full reference of the container image that implements the kube-rbac-proxy.
 	//+optional
 	KubeRbacProxyImage *string `json:"kubeRbacProxyImage"`
+
+	// PostgresImage is the full reference of the container image that implements the Postgres database.
+	//+optional
+	PostgresImage *string `json:"postgresImage"`
```
pixelsoccupied, line 144: Generally images are set using the operator env vars. Having it in env also signals users that they may not swap it out/not-configurable. Maybe this can be removed?

internal/controllers/utils/types.go
```diff
@@ -31,6 +32,7 @@ var InventoryConditionTypes = struct {
 	DeploymentServerError:     "DeploymentServerError",
 	ResourceServerError:       "ResourceServerError",
 	SmoRegistrationCompleted:  "SmoRegistrationCompleted",
+	DatabaseServerAvailable:   "DatabaseServerAvailable",
```
pixelsoccupied, line 35: On top of Available should we also have DB inprogres + failed status?

internal/controllers/inventory_controller_postgres.go
```diff
@@ -0,0 +1,305 @@
...
+			existing.Data[utils.ResourcesPasswordEnvName] = []byte(utils.GetPasswordOrRandom(utils.ResourcesPasswordEnvName))
+		}
+
+		err = utils.CreateK8sCR(ctx, t.client, &existing, t.object, utils.UPDATE)
+		if err != nil {
+			return fmt.Errorf("failed to create passwords: %w", err)
+		}
+	}
+
+	return nil
```
pixelsoccupied, line 211: Would nice have some additional success logs and/or debug logs in the funcs (here and others too). Would helpful in prod (and in dev) incase any of the crucial steps fail.

internal/service/postgres/files/postgresql.conf
```diff
@@ -27,3 +27,8 @@ log_destination = 'stderr'
 logging_collector = on
 log_directory = 'pg_log'
 log_filename = 'postgresql.log'
+
+# TLS
+ssl = on
+ssl_cert_file = '/secrets/tls/tls.crt'
+ssl_key_file = '/secrets/tls/tls.key'
```
pixelsoccupied, line 34: I havnt poked around with ssl but would it better to follow official example? https://github.com/sclorg/postgresql-container/tree/master/examples/enable-ssl . Looks they are going with `/opt/app-root/src/certs..`

pixelsoccupied (commented):
Good stuff! Looks forward to using it!

internal/controllers/inventory_controller.go
```diff
@@ -1055,7 +1066,7 @@ func (t *reconcilerTask) deployServer(ctx context.Context, serverName string) (u
 					{
 						Name:            utils.ServerContainerName,
 						Image:           image,
-						ImagePullPolicy: corev1.PullAlways,
+						ImagePullPolicy: corev1.PullIfNotPresent,
```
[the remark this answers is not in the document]
pixelsoccupied: This is usually a recommendation from OCP to use PullIfNotPresent. Especially important for SNO context where resource are constrained. Cant find the docs though where it was mentioned...but this is the "correct" way 🤔

api/inventory/v1alpha1/inventory_types.go
alegacy: Ok, I'm in favor of removing it given that we have the env variable method available to us.   I'll also take out the kube-rbac-proxy one while i'm at it.

api/inventory/v1alpha1/inventory_types.go
```diff
@@ -138,6 +138,11 @@ type InventorySpec struct {
 	// KubeRbacProxyImage is the full reference of the container image that implements the kube-rbac-proxy.
 	//+optional
 	KubeRbacProxyImage *string `json:"kubeRbacProxyImage"`
+
+	// PostgresImage is the full reference of the container image that implements the Postgres database.
+	//+optional
```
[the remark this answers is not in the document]
alegacy: Given that we want to remove the image attribute i'll hold off on the struct until i do the pv/pvc stuff.

internal/controllers/utils/types.go
alegacy: I never understood the benefit of having a laundry list of different statuses rather than having just a single state representation in a single condition with an appropriate failure reason... maybe for some things it makes sense but generally I find it awkward.   

Having said that, I can throw in the failed... but the inprogress would be super transient so probably not worth the hassle.

internal/service/postgres/files/postgresql.conf
alegacy: we could move them but it really doesn't make a difference since we are able to specify their location in the conf file.  The direct benefit to us is that we get to reuse the code that mounts the secret across all of our servers since they are all currently mounting to /secrets/tls/...

I'd rather just leave it as is.

internal/controllers/inventory_controller_postgres.go
```diff
@@ -0,0 +1,305 @@
...
+			"app":        serverName,
+		},
+	}
+
+	postgresImage := os.Getenv(utils.PostgresImageName)
+	if t.object.Spec.PostgresImage != nil {
+		postgresImage = *t.object.Spec.PostgresImage
+	}
+	if postgresImage == "" {
+		return fmt.Errorf("missing %s environment variable value", utils.PostgresImageName)
```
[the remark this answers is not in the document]
alegacy: Given that @pixelsoccupied suggested to remove the spec attribute i'll leave this as-is.

internal/controllers/utils/types.go
pixelsoccupied: Yeah status is a interesting one. Generally it's simply meant for the outside world to read and do something with it...so it's really up to us to decide what'ss useful for them. API convention doc has some good pointers about it https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties (a great doc in general) 

Just Fail should be good enough in this case!

internal/controllers/inventory_controller_postgres.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		t.logger.ErrorContext(
+			ctx,
+			"Failed to make config configmap for Database.",
+			slog.String("error", err.Error()),
+		)
+		return
+	}
+
+	err = utils.CreateK8sCR(ctx, t.client, config, t.object, utils.UPDATE)
```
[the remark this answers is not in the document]
alegacy: changed

internal/controllers/inventory_controller_postgres.go
```diff
@@ -0,0 +1,305 @@
...
+								Drop: []corev1.Capability{"ALL"},
+							},
+						},
+						Image:           postgresImage,
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						Ports: []corev1.ContainerPort{
+							{
+								Name:          utils.DatabaseTargetPort,
+								Protocol:      corev1.ProtocolTCP,
+								ContainerPort: 5432,
```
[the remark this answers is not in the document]
alegacy: changed

api/inventory/v1alpha1/inventory_types.go
alegacy: changed

internal/controllers/inventory_controller.go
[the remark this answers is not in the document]
mlguerrero12: It might be. However, I was wondering why this change on this pr.

These are for the servers, right? It won't run in the SNO. Just a concern because we are in the developing phase. I usually modify the quay image (same name) and recreate the operator to pick up latest changes.

internal/controllers/inventory_controller_postgres.go
alegacy: added

alegacy:
/hold

having an issue with the bundle-run testing

alegacy:
/hold cancel

...false alarm, problem between keyboard and chair.

internal/controllers/inventory_controller.go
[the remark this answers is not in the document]
alegacy: @mlguerrero12  If your version tag is set to `latest` it will automatically pull the latest even if it is set to `IfNotPresent`.

internal/controllers/inventory_controller.go
```diff
@@ -592,6 +592,22 @@ func (t *reconcilerTask) run(ctx context.Context) (nextReconcile ctrl.Result, er
 		return
 	}
 
+	// Create the database
+	err = t.createDatabase(ctx)
+	if err != nil {
+		err2 := t.updateORANO2ISMUsedConfigStatus(ctx, utils.InventoryDatabaseServerName,
```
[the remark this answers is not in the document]

internal/controllers/inventory_controller.go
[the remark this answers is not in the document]
mlguerrero12: @alegacy that's not k8s works https://kubernetes.io/docs/concepts/containers/images/#required-image-pull. Am I missing something in our code?

PR #341, opened 2024-11-22
fix 'bad matcher format: Watchdog' and some post merge cleanup

pixelsoccupied:
/cc @mlguerrero12 
/cc @alegacy

### DIFF
--- a/internal/service/alarms/internal/alertmanager/alertmanager.yaml
+++ b/internal/service/alarms/internal/alertmanager/alertmanager.yaml
@@ -12,7 +12,7 @@ route:
   - receiver: "null"
     matchers:
       # Always firing alert to verify alertmanager is working
-      - Watchdog
+      - alertname=Watchdog
   - receiver: oran_alarm_receiver
     group_by:
       # This label is guaranteed to be present in all alerts
--- a/internal/service/alarms/serve.go
+++ b/internal/service/alarms/serve.go
@@ -68,11 +68,6 @@ func Serve() error {
 		return fmt.Errorf("error creating client for hub: %w", err)
 	}
 
-	err = alertmanager.Setup(ctx, hubClient)
-	if err != nil {
-		return fmt.Errorf("error configuring alert manager: %w", err)
-	}
-
 	alarmsDict := dictionary.New(hubClient)
 	alarmsDict.Load(ctx)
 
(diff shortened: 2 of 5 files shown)


internal/service/alarms/internal/alertmanager/alertmanager.yaml
```diff
@@ -12,7 +12,7 @@ route:
   - receiver: "null"
     matchers:
       # Always firing alert to verify alertmanager is working
-      - Watchdog
+      - alertname="Watchdog"
```
pixelsoccupied, line 15: Wonder if we should Unmarshal it to Go struct to possible check for any config error validation.

mlguerrero12, line 15: I fixed but didn't push the commit. Thanks! Quotes are not needed. Only in for the null receiver

internal/service/alarms/serve.go
```diff
@@ -13,11 +13,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/alertmanager"
+
```
mlguerrero12, line 17: let's put it together with the other imports

pixelsoccupied: I don't modify it manually actually. Simply run `golangci-lint run --fix --config .golangci.yml` and let it do its thing.

internal/service/alarms/internal/alertmanager/alertmanager.yaml
pixelsoccupied: Updated! Wonder if we should Unmarshal it to Go struct to possibly validate any error from plain text file

internal/service/alarms/serve.go
mlguerrero12: weird, we've been grouping this kind of imports together. Lint locally and CI do not complain.

internal/service/alarms/internal/alertmanager/alertmanager.yaml
mlguerrero12: I think once we have integration tests we could detect this kind of issues. Besides, inside the matchers, you can put whatever you want.

internal/service/alarms/serve.go
pixelsoccupied: updated manually

PR #343, opened 2024-11-22
CNF-14144: Add migration init container for resource server

alegacy:
This commit adds the init container for the resource server.  The container runs prior to starting the resource server container so that its database schema is created and migrated up to the latest schema version.  The init container is run any time the resource server pod is recreated, and if there is a transient failure in the init due to a race condition in bringing up the database server it is automatically retried until it succeeds.

### DIFF
--- a/internal/controllers/utils/utils.go
+++ b/internal/controllers/utils/utils.go
@@ -182,6 +182,11 @@ func HasApiEndpoints(serverName string) bool {
 		serverName == InventoryAlarmSubscriptionServerName
 }
 
+// HasDatabase determines whether a server owns a logical database instance
+func HasDatabase(serverName string) bool {
+	return serverName == InventoryResourceServerName
+}
+
 func GetDeploymentVolumes(serverName string) []corev1.Volume {
 	if HasApiEndpoints(serverName) {
 		tlsDefaultMode := int32(os.FileMode(0o400))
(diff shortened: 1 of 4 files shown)


internal/controllers/utils/utils.go
```diff
@@ -198,6 +203,21 @@ func GetDeploymentVolumes(serverName string) []corev1.Volume {
 		}
 	}
 
+	if HasDatabase(serverName) || serverName == InventoryDatabaseServerName {
```
mlguerrero12, line 206: why do we need this volume for `InventoryDatabaseServerName`?

internal/service/common/db/migration.go
```diff
@@ -96,6 +96,12 @@ func NewHandler(cfg MigrationConfig) (*MigrationHandler, error) {
 		connStr += fmt.Sprintf("&x-migrations-table=%s", cfg.MigrationsTable)
 	}
 
+	if _, err := os.Stat(utils.DefaultServiceCAFile); err == nil {
```
mlguerrero12, line 99: why do we need this?

internal/controllers/utils/utils.go
alegacy: It is required by the init script that creates the users and logical databases.

mlguerrero12: I see, so this was not working before, right? This is not part of the migration init container

internal/service/common/db/migration.go
alegacy: Why do we need the `os.Stat(...)` check?   I had to add this to make sure that standalone testing isn't affected.  Once we get enough code in-place so that everyone can test in-cluster then we won't need this.

internal/controllers/utils/utils.go
alegacy: Let me take another look at this.

alegacy: Sorry, not sure what I was thinking there... I've removed it as it was unnecessary.  Thanks for catching this.

internal/service/common/db/migration.go
mlguerrero12: shouldn't we fail if this file is not found and sslMode is `verify-full`?

alegacy: It will fail on its own because the database connection won't initialize.

mlguerrero12: yeah, I was thinking about a more specific log. Buy, yes it doesn't matter much

PR #344, opened 2024-11-24
CNF-14144: Add resource server API controller

alegacy:
This adds an initial implementation of the new resource server API controller based on the alarm server API controller.  The endpoint implementations all return "not implemented", but since this implementation is not yet instantiated by default it has no impact on the existing inventory API functionality.

### DIFF
--- a/internal/service/alarms/serve.go
+++ b/internal/service/alarms/serve.go
@@ -17,7 +17,7 @@ import (
 	common "github.com/openshift-kni/oran-o2ims/internal/service/common/api"
 	"github.com/openshift-kni/oran-o2ims/internal/service/common/db"
 
-	api "github.com/openshift-kni/oran-o2ims/internal/service/alarms/api/generated"
+	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/api/generated"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/alertmanager"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/dictionary"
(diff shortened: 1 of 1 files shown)


internal/service/alarms/serve.go
```diff
@@ -52,6 +51,11 @@ func Serve() error {
 		cancel()
 	}()
 
+	password, exists := os.LookupEnv(utils.ResourcesPasswordEnvName)
```
mlguerrero12, line 54: AlarmsPasswordEnvName?

internal/service/alarms/serve.go
```diff
@@ -138,47 +148,10 @@ func Serve() error {
 		return fmt.Errorf("error starting server: %w", err)
 	case <-ctx.Done():
 		slog.Info("Shutting down server")
-		if err := gracefulShutdown(srv); err != nil {
-			return err
+		if err := common.GracefulShutdown(srv); err != nil {
```
mlguerrero12, line 151: let's simply return here. No need to wrap the error again in my opinion

alegacy: Our linter configuration requires that we wrap this error since they are in different packages.

internal/service/alarms/serve.go
```diff
@@ -85,20 +89,26 @@ func Serve() error {
...
-			ResponseErrorHandlerFunc: getOranRespErrFunc(),
+			RequestErrorHandlerFunc:  common.GetOranReqErrFunc(),
+			ResponseErrorHandlerFunc: common.GetOranRespErrFunc(),
 		},
 	)
 
 	r := http.NewServeMux()
 
+	// This also validates the spec file
+	swagger, err := api.GetSwagger()
```
mlguerrero12, line 100: nit: remove api from the import or use generated as in serve.go from resources. Just for consistency.

mlguerrero12 (commented):
/lgtm

internal/service/alarms/serve.go
alegacy: fixed.

internal/service/resources/serve.go
```diff
@@ -0,0 +1,136 @@
...
+	port         = "8080"
+	readTimeout  = 5 * time.Second
+	writeTimeout = 10 * time.Second
+	idleTimeout  = 120 * time.Second
+
+	username = "resources"
+	database = "resources"
+)
+
+// Serve start alarms server
```
pixelsoccupied, line 36: start `resource` server?

pixelsoccupied (commented):
/approve

PR #346, opened 2024-11-25
Use query builder for GetAlarmRecordID endpoint

mlguerrero12:
Use BOB SQL Toolkit to build SQL queries


mlguerrero12:
c.c. @pixelsoccupied @alegacy

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -21,16 +23,21 @@ type AlarmsRepository struct {
...
-	query := `
-SELECT alarm_event_record_id, alarm_definition_id, probable_cause_id, alarm_raised_time, 
-       alarm_changed_time, alarm_cleared_time, alarm_acknowledged_time, 
-       alarm_acknowledged, perceived_severity, extensions, resource_type_id
-FROM alarm_event_record 
-WHERE alarm_event_record_id = $1
-`
+	query, args, err := psql.Select(
+		sm.Columns(
+			AlarmEventRecordID, AlarmDefinitionID, ProbableCauseID,
```
alegacy, line 28: Would be good if we could somehow get this list automatically from the model object in the `alarms.internal.db.models` package.

pixelsoccupied: +1 for that! Hope there's way from Bob to easily convert that? Otherwise we might look into using reflection and just extract the "db" tag values?

mlguerrero12: Yes, reflection is the only option. I'll give it a try tomorrow. If it is too messy I'll stick with this

pixelsoccupied: Looks like we leverage the codegen from the lib https://bob.stephenafamo.com/docs/code-generation/usage (looks really nice tbh)

mlguerrero12: Done, please have another look @pixelsoccupied @alegacy

internal/service/alarms/internal/db/repo/utils.go
alegacy, line 1: Would you mind moving these common utils and tests under `service/common/repo`

alegacy: ...basically anything that is not directly specific to the alarms code should live under common so that it is easier/cleaner to share between resources & alarms.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -21,17 +23,21 @@ type AlarmsRepository struct {
...
-       alarm_changed_time, alarm_cleared_time, alarm_acknowledged_time, 
-       alarm_acknowledged, perceived_severity, extensions, resource_type_id
-FROM alarm_event_record 
-WHERE alarm_event_record_id = $1
-`
+	record := models.AlarmEventRecord{}
+	tags := getAllDBTagsFromStruct(&record)
+
+	query, args, err := psql.Select(
+		sm.Columns(tags.columns()...),
```
pixelsoccupied, line 30: hmmm probably the original way makes more sense (hardcoding the vals). 

The problem with this that we are sort of forced to  pull all the columns regardless of if we need them. The performance hit (pulling too much useless data + reflection) but worse...degraded readability is probably not worth the tradeoff with simply hardcoding the names of the cols.

mlguerrero12: I have another function where you can specify the name of the fields you want to get. I figured for this single object we could use all of them. For others, we will limit the fields

pixelsoccupied: Ah cool. Do you mind specifying the cols? 

If you specify it...there's no point in reflect right?

mlguerrero12: I meant the fields of the struct

pixelsoccupied:
/approve 

let's see if we need any changes later but i think its great to build on top this now! (e.g https://bob.stephenafamo.com/docs/models/table)

PR #349, opened 2024-11-25
Handle error: No handware node for cluster instance

tliu2021:
The following changes are included:
1. Return an error if no handware node is found for the cluster instance.
2. Update the unit tests to ensure the expected hardware node is created.


Missxiaoguo (commented):
/lgtm

PR #351, opened 2024-11-26
CNF-14954: Add alarm server resources to operator

alegacy:
This adds the new alarm server implementation to the operator's set of responsibilities to manage.  This handles setting up the prerequisites to starting the server such as the service account, the rbac rules and bindings, etc.  This effectively obsoletes the existing alarm subscription server implementation which was currently not being started by default.


internal/controllers/inventory_controller.go
```diff
@@ -321,50 +323,69 @@ func (t *reconcilerTask) setupDeploymentManagerServerConfig(ctx context.Context,
...
+	if err != nil {
+		t.logger.ErrorContext(
+			ctx,
+			"Failed to create alarm server cluster role",
+			slog.String("error", err.Error()),
+		)
+		return
+	}
+
+	err = t.createAlarmServerClusterRoleBinding(ctx)
```
pixelsoccupied, line 350: Curious why we need create these for the server? Cant we simply fetch operator's SA and skip creating rbacs and stuff?

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -1,29 +1,41 @@
...
 var alertManagerConfig []byte
 
 // Setup updates the alertmanager config secret with the new configuration
 func Setup(ctx context.Context, cl client.Client) error {
+	// Get the ACM namespace
+	acmNamespace, err := utils.GetACMNamespace(ctx, cl)
+	if err != nil {
+		return fmt.Errorf("failed to get ACM namespace: %w", err)
+	}
+	namespace := fmt.Sprintf("%s-observability", acmNamespace)
```
pixelsoccupied, line 34: think its better to just hardcode this ACM namespace? might be worth decoupling from searchAPI i think

pixelsoccupied (commented):
Just few nits but looking really good!

internal/service/alarms/internal/alertmanager/alertmanager.go
alegacy: The problem is that the AMC namespace isn't fixed.  In my lab it is `open-cluster-management` and in another lab it is `rhacm`.  I figured (maybe incorrectly) that the alertmanager would be in whatever namespace the search-api is in + the `-observability` suffix.

internal/controllers/inventory_controller.go
alegacy: That's a good question.  They each have their own rbac requirements so from a security point of view of it is best to limit them to only the minimum rules that they need rather than giving all of them the broader permissions of the operator.  Then there would be the issue of the TLS certs/keys.  Our rbac proxy needs to have certs/keys that align with their FQDN hostname so sharing a set of TLS certs/keys across all of the servers wouldn't work properly.  

Unless you had something else in mind I don't see how we can avoid most of the things that are setup on a per-server basis.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: > ACM namespace isn't fixed

tbh i think its pretty much guaranteed to be fixed if done through official channel. And for observability stack if `open-cluster-management-observability` doesn't exist the whole operator may actually fail

Interesting that you have a ns `rhacm`! Like if we dont have `open-cluster-management` and `open-cluster-management-observability` probably a symptom of bigger problem (acm and observability not installed correctly)!

internal/controllers/inventory_controller.go
alegacy: The operator needs to be granted those permissions if it is to delegate to a server.  i.e., it can't grant something that it itself doesn't already have permission to do.

api/inventory/v1alpha1/inventory_types.go
```diff
@@ -148,10 +148,10 @@ type InventorySpec struct {
 	//+optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resource Server Configuration",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	ResourceServerConfig ResourceServerConfig `json:"resourceServerConfig,omitempty"`
-	// AlarmSubscriptionServerConfig contains the configuration for the alarm server.
+	// AlarmServerConfig contains the configuration for the alarm server.
 	//+optional
-	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Alarm Subscription Server Configuration",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
-	AlarmSubscriptionServerConfig AlarmSubscriptionServerConfig `json:"alarmSubscriptionServerConfig"`
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Alarm Server Configuration",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	AlarmServerConfig AlarmServerConfig `json:"alarmSubscriptionServerConfig"`
```
mlguerrero12, line 154: tag seems to be incorrect

config/crd/bases/o2ims.oran.openshift.io_inventories.yaml
```diff
@@ -42,8 +42,8 @@ spec:
               alarmSubscriptionServerConfig:
```
mlguerrero12, line 42: due to above tag

internal/service/alarms/internal/dictionary/dictionary.go
```diff
@@ -116,7 +116,11 @@ var getClientForCluster = k8s_client.NewClientForCluster
 func (r *AlarmDictionary) processCluster(ctx context.Context, version string) ([]monitoringv1.Rule, error) {
 	cluster, err := r.getManagedCluster(ctx, version)
 	if err != nil {
-		return nil, err
+		// TODO remove when we only process cluster version that actually exist
```
mlguerrero12, line 119: the error from processCluster is simply logged. It shouldn't stop the server

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: I agree, this is clearly stated in the documentation. Let's not try to support other options. Our system should expect everything to be in `open-cluster-management-observability`

alegacy: can you point me to the documentation?

mlguerrero12: https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.11/html/observability/observing-environments-intro#configuring-alertmanager

internal/service/alarms/internal/dictionary/dictionary.go
alegacy: i'll recheck

internal/controllers/inventory_controller.go
```diff
@@ -321,50 +323,69 @@ func (t *reconcilerTask) setupDeploymentManagerServerConfig(ctx context.Context,
...
+			"Failed to create alarm server RBAC proxy cluster role binding",
 			slog.String("error", err.Error()),
 		)
 		return
 	}
 
-	// Create the Service needed for the alarm subscription server.
-	err = t.createService(ctx, utils.InventoryAlarmSubscriptionServerName, utils.DefaultServicePort, utils.DefaultTargetPort)
+	// Create the Service needed for the alarm server.
+	err = t.createService(ctx, utils.InventoryAlarmServerName, utils.DefaultServicePort, utils.DefaultTargetPort)
```
mlguerrero12, line 373: our servers are listening on port 8080. DefaultServicePort is 8000.

internal/service/alarms/internal/dictionary/dictionary.go
alegacy: You are correct.  The error is simply logged... not sure why I thought it wasn't.  I'll revert this part of the change.  Thanks.

internal/controllers/inventory_controller.go
alegacy: i'll change them to 8000 to align with the existing servers that we already deploy to avoid further confusion.

internal/service/alarms/internal/alertmanager/alertmanager.go
alegacy: The documentation never says that the namespace MUST be a specific value.  It simply walks through an example of how to configure it and it happens to have chosen that as a name.    Having said that, if both of you feel that hardcoding it is an acceptable strategy then we can certainly do that.

internal/controllers/utils/utils.go
```diff
@@ -298,12 +299,35 @@ func GetSearchURL(ctx context.Context, c client.Client) (string, error) {
...
+	service, err := GetSearchAPI(ctx, c)
+	if err != nil {
+		return "", err
+	}
+	return service.Namespace, nil
+}
+
+// GetServerDatabasePasswordName retrieves name of the environment variable used to store the server's database password
+func GetServerDatabasePasswordName(serverName string) (string, error) {
+	if name, ok := DatabaseNameMappings[serverName]; ok {
```
mlguerrero12, line 313: There are already `utils.AlarmsPasswordEnvName`and `utils.ResourcesPasswordEnvName`. Cleaner to do if serverName == InventoryAlarmServerName return utils.Alarms......

No need to return errors or have a name mapping

alegacy: This is to lookup the name of the logical database for each server.  The variables that you referenced hold the passwords for those servers.

internal/controllers/inventory_controller.go
```diff
@@ -1283,14 +1429,14 @@ func (t *reconcilerTask) createIngress(ctx context.Context) error {
 								},
 							},
 							{
-								Path: "/o2ims-infrastructureMonitoring/v1/alarmSubscriptions",
+								Path: "/o2ims-infrastructureMonitoring",
```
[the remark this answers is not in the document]
alegacy: No, it will directly everything starting with this path prefix to the alarm server.

internal/controllers/utils/utils.go
mlguerrero12: you create a secret called fmt.Sprintf("%s-passwords", utils.InventoryDatabaseServerName). That secret has as keys the constants I mention. 

Now, here you are trying to find the name of the key related to the server which are the same constants.

internal/controllers/inventory_controller.go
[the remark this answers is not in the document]
mlguerrero12: it's been a while since I last worked with ingress, but if I remember correctly, when a path matches multiple rules, it uses the more specific one. Wouldn't be have problems with "/o2ims-infrastructureMonitoring" and "/". It will match both, but everything will be sent to the alarm backend. Not sure, will look it up as well.

internal/controllers/utils/utils.go
alegacy: I can change it to `if` statements but we'll see need to return an error if the server name doesn't match one of the 2 supported values.

internal/controllers/inventory_controller.go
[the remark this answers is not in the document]
mlguerrero12: I misread `o2ims-infrastructureMonitoring` - ignore

PR #352, opened 2024-11-27
Generate and use client to make api calls towards resource server

mlguerrero12:
- Add missing Model property to ResourceType
- Generate client from resource server api spec
- Move k8s client to commons


mlguerrero12:
@pixelsoccupied @alegacy, PTAL

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return resp.JSON200, nil
+}
+
+// getClient gets the client for the resource server
+func getClient() (*resources.ClientWithResponses, error) {
+	var err error
+	once.Do(func() {
+		// TODO: update the URL to the correct one once the service is deployed
+		url = os.Getenv("RESOURCE_SERVER_URL")
```
alegacy, line 47: The service is already deployed.  It's not using the new API, but theoretically it should still work.  

See `GetServiceURL` in https://github.com/openshift-kni/oran-o2ims/pull/351 to build the URL.

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,65 @@
...
+			err = fmt.Errorf("environment variable RESOURCE_SERVER_URL must be set")
+		}
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Default transport (global var within net/http package) is used here. TCP connections are reused.
+	// Minimal overhead is introduced by creating a new client for each request.
+	hc := http.Client{}
```
alegacy, line 58: The resource server has HTTPS enabled so the transport will have to load the Service CA from a file.  It would be best if that was not done on each request, but rather once when the process loads.

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,65 @@
...
+
+	c, errNewClient := resources.NewClientWithResponses(url, resources.WithHTTPClient(&hc))
+	if err != nil {
+		err = fmt.Errorf("failed to create client: %w", errNewClient)
+	}
```
alegacy, line 63: Not sure why we are using `errNewClient` here rather than `err`.  If the client creation fails we should return.

internal/service/alarms/serve.go
```diff
@@ -67,13 +68,22 @@ func Serve() error {
...
 
+	// Get Resource Types
+	resourceTypes, err := resourceserver.GetResourceTypes(ctx)
+	if err != nil {
+		// TODO: return error - Ignore for now
+		// Mock data for testing
+		resourceTypes = &[]resources.ResourceType{}
+	}
+
+	// Load dictionary
```
alegacy, line 84: We should try to keep the top level `serve` function as minimalistic as possible.  Creating clients, initializing structs, starting servers, etc... we should keep the actual work part (i.e., calling APIs, reading/writing to DB, etc...) within separate functions/files.

internal/service/alarms/internal/resourceserver/resourceserver.go
mlguerrero12: Thanks. Good to know. I created this pr without taking into account the actual provisioning. Will update it

internal/service/alarms/internal/resourceserver/resourceserver.go
mlguerrero12: will review this based on previous comment

internal/service/alarms/serve.go
mlguerrero12: agree, the idea here is to avoid to call this twice. It will be used in the dictionary and in the server as well. Will see what I can do

internal/service/common/clients/resources/client.gen.go
```diff
@@ -0,0 +1,2792 @@
...
+	}
+}
+
+// ClientWithResponsesInterface is the interface specification for the client with responses above.
+type ClientWithResponsesInterface interface {
+	// GetAllVersionsWithResponse request
+	GetAllVersionsWithResponse(ctx context.Context, params *GetAllVersionsParams, reqEditors ...RequestEditorFn) (*GetAllVersionsResponse, error)
+
+	// GetCloudInfoWithResponse request
+	GetCloudInfoWithResponse(ctx context.Context, params *GetCloudInfoParams, reqEditors ...RequestEditorFn) (*GetCloudInfoResponse, error)
```
pixelsoccupied, line 1748: Cool so basically we will reuse the oran endpoints for our stuff?

internal/service/common/clients/resources/oapi-codegen.yaml
```diff
@@ -0,0 +1,17 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/oapi-codegen/oapi-codegen/HEAD/configuration-schema.json
+package: resources
+output: client.gen.go
```
pixelsoccupied, line 3: shouldnt this whole thing be under `alarms` dir since we are the client of resource server?

pixelsoccupied (commented):
good stuff! alarms + resource server connectivity incoming!

internal/service/common/clients/resources/oapi-codegen.yaml
alegacy: sort of depends on how we look at this... if we had 2 clients wanting to use it then we'd likely want to put in under `resources`... but for now only `alarms` is a user so it seems better there because of that.    In the end it doesn't matter much until there is a 2nd client.

internal/service/alarms/internal/resourceserver/resourceserver.go
mlguerrero12: updated

internal/service/alarms/internal/resourceserver/resourceserver.go
mlguerrero12: updated

internal/service/alarms/internal/resourceserver/resourceserver.go
mlguerrero12: fixed

internal/service/alarms/serve.go
mlguerrero12: updated

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -13,13 +13,15 @@ import (
...
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/clients/resources"
 	"github.com/openshift-kni/oran-o2ims/internal/service/common/utils"
 )
 
 // All DB interaction code goes here
 
 type AlarmsRepository struct {
-	Db *pgxpool.Pool
+	Db            *pgxpool.Pool
+	ResourceTypes *[]resources.ResourceType
```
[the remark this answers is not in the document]
mlguerrero12: fixed

internal/service/common/clients/resources/client.gen.go
mlguerrero12: I think so

internal/service/common/clients/resources/oapi-codegen.yaml
mlguerrero12: moved it within alarms directory

mlguerrero12:
addressed comments, PTAL again @alegacy, @pixelsoccupied

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,122 @@
...
+	url := utils.GetServiceURL(utils.InventoryResourceServerName)
+
+	// Use for local development
+	resourceServerURL := os.Getenv(resourceServerURLEnvName)
+	if resourceServerURL != "" {
+		url = resourceServerURL
+	}
+
+	// Set up transport
+	tr := &http.Transport{
```
alegacy, line 44: We have a utility to do this.   See `GetDefaultTLSConfig` in utils

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,122 @@
...
+		data, err := os.ReadFile(utils.DefaultServiceCAFile)
+		if err != nil {
+			// Should never happen
+			return nil, fmt.Errorf("failed to read CA file: %w", err)
+		}
+
+		tr.TLSClientConfig.RootCAs.AppendCertsFromPEM(data)
+	}
+
+	hc := http.Client{Transport: tr}
```
alegacy, line 63: We have a utility to do this. See `GetDefaultTLSConfig` in utils

alegacy: or `GetDefaultBackendTransport` depending on what you need

mlguerrero12: I'm using `GetDefaultBackendTransport` now. Thanks

internal/service/alarms/internal/resourceserver/resourceserver.go
```diff
@@ -0,0 +1,122 @@
...
+
+	tokenPath := utils.DefaultBackendTokenFile
+
+	// Use for local development
```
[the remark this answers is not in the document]
mlguerrero12: I'd like to keep them for now

PR #355, opened 2024-11-27
CNF-14948: implement alertmanager endpoint to receive and update db with events (no subscriber notification)

pixelsoccupied:
when a list of alerts come in 
- check if event table and the current payload match. If missing set them to resolved
- collect alarm definitions potentially inject them into alarm event record (e.g alarm def id)
- Insert and potentially update events tables 
- Move anything thats's resolved into archive 

Note: Between the last and second last steps we will notify subscriber


internal/service/alarms/internal/db/models/alarm_event_record.go
```diff
@@ -3,30 +3,35 @@ package models
...
-	PerceivedSeverity     int                    `db:"perceived_severity"`
-	Extensions            map[string]interface{} `db:"extensions"`
-	ResourceID            uuid.UUID              `db:"resource_id"`
-	ResourceTypeID        uuid.UUID              `db:"resource_type_id"`
-	NotificationEventType int                    `db:"notification_event_type"`
-	AlarmStatus           string                 `db:"alarm_status"`
-	Fingerprint           string                 `db:"fingerprint"`
-	AlarmSequenceNumber   int64                  `db:"alarm_sequence_number"`
-	CreatedAt             time.Time              `db:"created_at"`
+	AlarmEventRecordID    *uuid.UUID                  `db:"alarm_event_record_id"`
```
alegacy, line 14: Why is this being made nullable?  It is the primary key so it should always be there.    Is it because we are relying on the DB to set it on the insert?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +59,193 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+		name := alertmanager.GetAlertName(*alert.Labels)
+		resourceTypeId := alertmanager.GetResourceTypeID(*alert.Labels)
+		b = append(b, psql.ArgGroup(name, resourceTypeId))
+	}
+
+	return b
+}
+
+// ArchiveResolvedAlarmEventRecord Delete resolved and move to archive
+func (ar *AlarmsRepository) ArchiveResolvedAlarmEventRecord(ctx context.Context) error {
```
alegacy, line 192: This should be run in a transaction otherwise we run the risk of having an inconsistent view of the data... if the process dies or is restarted at exactly between lines 205 and 206 then we'll have deleted data but not moved it to the archive.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +59,193 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+
+	// Set the column list for the InsertQuery
+	query.Expression.Columns = columns
+
+	// Add the values for each record
+	values := make([]bob.Mod[*dialect.InsertQuery], 0, len(records))
+	for _, record := range records {
+		// Build the values for the row
+		args := []any{
+			record.AlarmRaisedTime, record.AlarmChangedTime, record.AlarmClearedTime,
```
alegacy, line 113: It's going to be tricky 👀 to inspect this to make sure the column names and the column data match, could you align the line breaks so that the same column names and data fields are on the same lines.  Maybe 3 per line each.

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +67,104 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+// getPerceivedSeverity am's `severity` to oran's PerceivedSeverity
+func getPerceivedSeverity(labels map[string]string) api.PerceivedSeverity {
+	input, ok := labels["severity"]
+	if !ok {
+		slog.Warn("Could not find severity label. This may impact the value of PerceivedSeverity")
+		return api.INDETERMINATE
+	}
+	slog.Info("Found alert severity", "severity", input)
+
+	// TODO: "info" semes like common one but no oran mapping
```
alegacy, line 147: "semes" -> "seems"

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +67,104 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+	}
+
+	return val
+}
+
+// getPerceivedSeverity am's `severity` to oran's PerceivedSeverity
+func getPerceivedSeverity(labels map[string]string) api.PerceivedSeverity {
+	input, ok := labels["severity"]
+	if !ok {
+		slog.Warn("Could not find severity label. This may impact the value of PerceivedSeverity")
```
alegacy, line 142: maybe dump out the other fields so we can track down which alert is causing this.

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +67,104 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+	case "critical":
+		return api.CRITICAL
+	case "major":
+		return api.MAJOR
+	case "minor":
+		return api.MINOR
+	case "warning":
+		return api.WARNING
+	default:
+		slog.Warn("Could not map", "severity", input)
```
alegacy, line 160: maybe dump out the other fields so we can track down which alert is causing this.

internal/service/alarms/internal/db/models/alarm_event_record.go
pixelsoccupied: > Is it because we are relying on the DB to set it on the insert

Yeah exactly. Normally when inserting the live data...we are auto generating uuid with db. but when ready to move to archive it will already have a UUID. This var is basically used to determine if we working with archive or live table and dynamically update the queries.

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -44,7 +44,7 @@ CREATE OR REPLACE FUNCTION update_alarm_event_sequence()
     RETURNS TRIGGER AS $$
 BEGIN
     -- Update sequence if status changes to 'resolved' or if alarm_changed_time is updated
-    IF (NEW.status = 'resolved' AND OLD.status IS DISTINCT FROM 'resolved')
+    IF (NEW.alarm_status = 'resolved' AND OLD.alarm_status IS DISTINCT FROM 'resolved')
```
alegacy, line 47: If we're going to be referencing these literals (e.g., 'resolved') frequently then it may be best to define some constants in a file specific to the alertmanager.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: Yeah actually a great idea! I was actually suffering from it during dev (it was deleting but wasn't inserting cause of an unrelated error) and was gonna look into transaction before posting the PR but totally forgot lol

pixelsoccupied:
/hold 
making some updates

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
pixelsoccupied: hmmm you mean something like this https://github.com/openshift-kni/oran-o2ims/blob/main/internal/service/alarms/api/generated/alarms.generated.go#L38-L39 ?

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: updated to use transaction!

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -190,33 +187,42 @@ func getGetAlertNameAndResourceTypeID(am *api.AlertmanagerNotification) []bob.Ex
 
 // ArchiveResolvedAlarmEventRecord Delete resolved and move to archive
 func (ar *AlarmsRepository) ArchiveResolvedAlarmEventRecord(ctx context.Context) error {
```
alegacy, line 189: Might want to pluralize the function name since it is handling multiple records.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: added. printing all the labels now.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: added. printing all the labels now.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: Removed it! Spoke with Brent about the possible mapping of Info. We are using "Warning" for that.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +59,193 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+			}
+		}
+
+		values = append(values, im.Values(psql.Arg(args...)))
+	}
+
+	// Apply the values to the query
+	query.Apply(values...)
+
+	// constrains only applicable for non-archive table
```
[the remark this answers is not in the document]
pixelsoccupied: Updated! Lol i wrote the same thing in the table's comments as well :P

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: Good idea! Wonder if there's a CI check we can use to catch this. But will follow 3 per line for all.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: Updated

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -21,9 +21,8 @@ CREATE TABLE IF NOT EXISTS alarm_event_record (
     extensions JSONB, -- Additional data for extensibility
 
     -- O-RAN additional data to create AlarmEventNotification
-    resource_id UUID NOT NULL, -- Same as manager_cluster_id for caas alerts
-    resource_type_id UUID NOT NULL, -- Derived from manager_cluster_id
-    notification_event_type INT NOT NULL, -- Should be enum calculated from current alert
+    resource_id UUID, -- Same as manager_cluster_id for caas alerts
```
pixelsoccupied, line 24: Specifically removed `notification_event_type` from table as this can be calculated when we are ready to send notification to subscriber using `alarm_acknowledged` and `alarm_status`

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +66,103 @@ func Setup(ctx context.Context, cl client.Client) error {
 	slog.Info("Successfully configured alertmanager")
 	return nil
 }
+
+// ConvertAmToAlarmEventRecordModels get alarmEventRecords based on the alertmanager notification and AlarmDefinition
+func ConvertAmToAlarmEventRecordModels(am *api.AlertmanagerNotification, aDefinitionRecords []models.AlarmDefinition) []models.AlarmEventRecord {
```
mlguerrero12, line 71: I would create a type labels map[string][string] and then have

func(r *labels) Get.....() {}

for each function you have. It's cleaner I think

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +66,103 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+
+		// See if possible to pick up additional info from its definition
+		for _, def := range aDefinitionRecords {
+			if def.AlarmName == GetAlertName(*alert.Labels) { // TODO: match with the resourceTypeID too once init data is ready
+				record.AlarmDefinitionID = def.AlarmDefinitionID
+				record.ProbableCauseID = def.ProbableCauseID
+			}
+		}
+
+		// Note: note only thing we are missing is `AlarmChangedTime` since it's not available from alertmanager
```
mlguerrero12, line 96: as far as I could see, EndsAt is different than 0000.... when the alert is resolved, changed time must have this value since changedtime is updated whenever any other field changes

internal/service/alarms/internal/alarms_server.go
```diff
@@ -91,11 +92,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
 
+// AmNotification Used internally, alertmanager from ACM calls here.
+// Note: the errors returned can also be view under alertmanager pod logs but also logging here for convenience
 func (a *AlarmsServer) AmNotification(ctx context.Context, request api.AmNotificationRequestObject) (api.AmNotificationResponseObject, error) {
-	// TODO: Implement the logic to handle the AM notification
-	slog.Debug("Received AM notification", "groupLabels", request.Body.GroupLabels)
-	for _, alert := range request.Body.Alerts {
-		slog.Debug("Alert", "fingerprint", alert.Fingerprint, "startsAt", alert.StartsAt, "status", alert.Status)
+	// Audit the table with full list of alerts in the current payload. If missing set them to resolve
+	if err := a.AlarmsRepository.ResolveNotificationIfNotInCurrent(ctx, request.Body); err != nil {
```
mlguerrero12, line 99: Are we sure that a single notification has all alerts? Say, alert1, alert2 and alert3 are firing and that the group_interval is 1m. If there are no changes, nothing will be reported after 1 m, but what if a new alert comes in in between. Will am wait until it completes 1m to report all of them? or does report it right away? group_interval applies when the parent receiver has no group_by?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +60,201 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+		return nil
+	})
+}
+
+// ResolveNotificationIfNotInCurrent find and only keep the alerts that are available in the current payload
+func (ar *AlarmsRepository) ResolveNotificationIfNotInCurrent(ctx context.Context, am *api.AlertmanagerNotification) error {
+	query := psql.Update(
+		um.Table(models.AlarmEventRecordView.Name(ctx)),
+		um.SetCol("alarm_status").ToArg(api.Resolved),
+		um.Where(psql.Quote("fingerprint").NotIn(getGetAlertFingerPrints(am)...)),
```
mlguerrero12, line 236: or if StartsAt is different, it could be a new alert (same fingerprint) that started after again. Assuming, we were off when the previous one went to Resolved and this one is present. Makes sense?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +60,201 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+	)
+
+	sql, params, err := query.Build()
+	if err != nil {
+		return fmt.Errorf("failed to build alarm definitions query when processing AM notification: %w", err)
```
mlguerrero12, line 241: this is not for alarms definitions?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +60,201 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+		return sql, args, fmt.Errorf("failed build upsert query for tabls %s: %w", tableName, err)
+	}
+
+	return sql, args, nil
+}
+
+// GetAlarmDefinitions needed to build out aer
+func (ar *AlarmsRepository) GetAlarmDefinitions(ctx context.Context, am *api.AlertmanagerNotification) ([]models.AlarmDefinition, error) {
+	query := psql.Select(
+		sm.Columns("alarm_name", "alarm_definition_id", "probable_cause_id", "resource_type_id"),
```
mlguerrero12, line 157: didn't we agree on using the name of the fields from the structure?  We can't have endpoints with different approaches.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -91,11 +92,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+
+	// Get the definition data based on current set of Alert names and managed cluster ID
+	alarmDefinitions, err := a.AlarmsRepository.GetAlarmDefinitions(ctx, request.Body)
+	if err != nil {
+		msg := "failed to get AlarmDefinitions"
+		slog.Error(msg, "err", err)
+		return nil, fmt.Errorf("%s: %w", msg, err)
+	}
+
+	// Combine possible definitions with events
```
mlguerrero12, line 113: perhaps we could pass aerModels to GetAlarmDefinitions (and then to getGetAlertNameAndResourceTypeID)?

internal/service/alarms/internal/alarms_server.go
```diff
@@ -91,11 +92,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+	aerModels := alertmanager.ConvertAmToAlarmEventRecordModels(request.Body, alarmDefinitions)
```
mlguerrero12, line 114: I wonder if we should first build the aerModels, withouth the alarm definitions, then pass it to the GetAlarmDefinitions because we use it to get related alarm definitions and then add the for you have in ConvertAmToAlarmEventRecordModels to map the alerts with the definitions

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +66,103 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+			ResourceID:        GetResourceID(*alert.Labels),
+			ResourceTypeID:    GetResourceTypeID(*alert.Labels),
+		}
+
+		// Update Extensions with things we didn't really process
+		record.Extensions = getExtensions(*alert.Labels, *alert.Annotations)
+
+		// See if possible to pick up additional info from its definition
+		for _, def := range aDefinitionRecords {
+			if def.AlarmName == GetAlertName(*alert.Labels) { // TODO: match with the resourceTypeID too once init data is ready
```
mlguerrero12, line 90: it will be possible to have the same alert name but from different resource type IDs (belong to different dicts). If 2 or more alerts with the same name and different managed cluster version come, it seems to me that they will be assigned the same alarmdefinitionID, am I right?

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -61,3 +66,103 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+	records := make([]models.AlarmEventRecord, 0, len(am.Alerts))
+	for _, alert := range am.Alerts {
+		slog.Info("Converting Alertmanager alert", "alert name", GetAlertName(*alert.Labels))
+		record := models.AlarmEventRecord{
+			AlarmRaisedTime:   *alert.StartsAt,
+			AlarmClearedTime:  alert.EndsAt,
+			PerceivedSeverity: getPerceivedSeverity(*alert.Labels),
+			AlarmStatus:       string(*alert.Status),
+			Fingerprint:       *alert.Fingerprint,
+			ResourceID:        GetResourceID(*alert.Labels),
```
mlguerrero12, line 81: I believe we should ignore alerts that do not have a valid resourceID (managed_cluster label), what do we get by storing them?

internal/service/alarms/internal/db/models/alarm_event_record.go
```diff
@@ -3,30 +3,34 @@ package models
...
+	Fingerprint           string                      `db:"fingerprint"`
+	AlarmSequenceNumber   int64                       `db:"alarm_sequence_number"`
+	CreatedAt             time.Time                   `db:"created_at"`
 }
 
 func (r *AlarmEventRecord) TableName() string {
 	return "alarm_event_record"
 }
+
+var AlarmEventRecordView = psql.NewView[AlarmEventRecord]("public", "alarm_event_record")
```
mlguerrero12, line 36: why this view?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -51,3 +60,201 @@ func (ar *AlarmsRepository) GetAlarmEventRecordWithUuid(ctx context.Context, uui
...
+		values = append(values, im.Values(psql.Arg(args...)))
+	}
+
+	// Apply the values to the query
+	query.Apply(values...)
+
+	// constraints only applicable for non-archive table
+	if tableName == models.AlarmEventRecordView.Name(context.Background()).String() {
+		query.Apply(im.OnConflictOnConstraint("unique_fingerprint_alarm_raised_time").DoUpdate(
+			im.SetExcluded("alarm_changed_time"),
```
mlguerrero12, line 136: say we have an alert with a valid resourceID but we failed to collect the alarm dict, I suppose we will still report it to the user. Then, before the next notification comes, we manage to collect the alarm dict. We will update the alarm defintion ID which also means that we would have to update the alarm_changed_time since one of the values has changed, makes sense?

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: Basically there's no specific "changesAt" time in alertmanager (unlike StartsAt or EndsAt...couldnt find it in AM project about it either). So instead of inferring something (e.g keep track of changes in the row) we are letting it stay null. Touched base with Brent about this and he said to let go through as null.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: hmmm so so we are also dealing with 'annotation' (as well as 'label') later to combine them together for extension. Might be a bit strange if we do type labels but leave the rest as regular map. IMO this is already quite minimal? Otherwise maybe another PR for that change?

mlguerrero12: that's another thing, why are we starting with all possible data in extensions? We should start empty and adding stuff later in my opinion

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: I know, but a update trigger might do the trick, it can check the values for us

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: Yeah its actually pretty valuable to have everything in there to actually understand the alerts that are in the DB (without it its simply UUIDs in the DB)

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: Yeah that's easy to do (i originally had this tbh) but again...given its not coming from alertmanager i was told just keep it as null. But we can simply do that from table so no code changes needed...I'll take that on once im back from PTO as would need test out.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: Yeah. Update to check for both fingerprint and startAt.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: fixed

internal/service/alarms/internal/db/migrations/000004_create_alarm_event_record_archive_table.up.sql
```diff
@@ -1,21 +1,21 @@
...
+    perceived_severity INT,
     extensions JSONB,
-    resource_id UUID NOT NULL,
-    resource_type_id UUID NOT NULL,
-    notification_event_type INT NOT NULL,
-    alarm_status VARCHAR(20) DEFAULT 'firing' NOT NULL,
-    finger_print TEXT NOT NULL,
+    resource_id UUID,
+    resource_type_id UUID,
+    notification_event_type INT,
```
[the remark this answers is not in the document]
pixelsoccupied: updated

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: Yeah basically the idea is that we collect all the alerts for now. And try to see if anything is missing later instead of  preemptively dropping rows. Given that we are now collecting them...i'm able to tell which ones are faulty ones just with quick in the DB. Cause in theory they should all have managed_cluster so nothing should be ignored

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: alarm_changed_time is currently not being added since alertmanager is not sending it

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: yeah i think

internal/service/alarms/internal/alarms_server.go
pixelsoccupied: Yeah either way should work. Found it simpler to get all the definitions from the current alertmanager payload instead from aer. 

The idea is that aer is basically an combination of alertmanager playload + definition (which also happens to be the two params for `ConvertAmToAlarmEventRecordModels`.

internal/service/alarms/internal/db/repo/alarms_repository.go
browsell: Isn't there an updatedAt field in the alert ?

pixelsoccupied: Hey Brent yeah it does but only if we call the API (pull). The notification doesnt have it https://prometheus.io/docs/alerting/latest/configuration/#webhook_config

pixelsoccupied: https://github.com/prometheus/alertmanager/ issues/2142#issuecomment-567087632 -- more info here

pixelsoccupied:
/hold 

had a quick chat with Brent and will update `alarm_changed_time` to be the same `alarm_raised_time` instead of keeping it null. updating now

pixelsoccupied:
/hold

testing with latest changes

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: updated

internal/service/alarms/internal/db/models/alarm_event_record.go
pixelsoccupied: removed

pixelsoccupied:
/unhold

ready for a review. this PR now fully connects the init data (alarms dict and def derived from `clusterserver`) with the caas events. as part of the integration...took the opportunity to clean up the `dictionary` package (now called `dictionary_definition`) by splitting up the large func into smaller pieces along with better better error handling when Go routines are launched. When processing events, code is now using internal lib that uses reflect to get the DB tag.

Tested end to end in cluster (dev env switched to cluster)

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +149,229 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+	})
+}
+
+// ResolveNotificationIfNotInCurrent find and only keep the alerts that are available in the current payload
+func (ar *AlarmsRepository) ResolveNotificationIfNotInCurrent(ctx context.Context, am *api.AlertmanagerNotification) error {
+	m := models.AlarmEventRecord{}
+	dbTags := utils.GetAllDBTagsFromStruct(m)
+	query := psql.Update(
+		um.Table(m.TableName()),
+		um.SetCol(dbTags["AlarmStatus"]).ToArg(api.Resolved),
```
mlguerrero12, line 347: As per our discussion about the clearing type of Alarms, change the perceived_severity to CLEARED here. This corresponds to the IMS-Initiated Alarm Clear section for automatically cleared alarms.

Also, you should also change the `alarmChangedTime` field. The specification says that this should change if any of the Alarm Record fields changes. So, this is not something that should be sent by the AlertManager but something is completely under our control.

There are 2 ways to change an Alarm Record, when it is acknowledged and when it is cleared. So, similar to what we do in the PATCH Alarm Record, save the current timestamp here since your are clearing the alarm.

internal/service/alarms/cmd/serve.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms"
 )
 
-var config alarms.AlarmsServerConfig
+var config alarms.ServerConfig
```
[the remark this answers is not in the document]
pixelsoccupied: https://www.jetbrains.com.cn/en-us/help/inspectopedia/GoNameStartsWithPackageName.html

Basically all references to names in a package will be done using the package name, so one can omit that name from the identifiers

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: /hold

true. updating this now

internal/service/alarms/internal/alarms_server.go
```diff
@@ -363,11 +366,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
 	return nil, fmt.Errorf("not implemented")
 }
 
+// AmNotification Used internally, alertmanager from ACM calls here.
```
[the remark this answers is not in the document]
pixelsoccupied: updated

internal/service/alarms/internal/alarms_server.go
```diff
@@ -363,11 +366,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+// Note: the errors returned can also be view under alertmanager pod logs but also logging here for convenience
 func (a *AlarmsServer) AmNotification(ctx context.Context, request api.AmNotificationRequestObject) (api.AmNotificationResponseObject, error) {
-	// TODO: Implement the logic to handle the AM notification
-	slog.Debug("Received AM notification", "groupLabels", request.Body.GroupLabels)
-	for _, alert := range request.Body.Alerts {
-		slog.Debug("Alert", "fingerprint", alert.Fingerprint, "startsAt", alert.StartsAt, "status", alert.Status)
+	// Audit the table with full list of alerts in the current payload. If missing set them to resolve
+	if err := a.AlarmsRepository.ResolveNotificationIfNotInCurrent(ctx, request.Body); err != nil {
+		msg := "failed to resolve notification that are not present"
+		slog.Error(msg, "err", err)
```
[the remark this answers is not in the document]
pixelsoccupied: updated

internal/service/alarms/cmd/serve.go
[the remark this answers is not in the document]
mlguerrero12: I know those things. However, something that I value more in a project is consistency. Nothing wrong about changing this. All the opposite. But please change it for the other servers as well.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -363,11 +364,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+		msg := "failed to get AlarmDefinitions"
+		slog.Error(msg, "error", err)
+		return nil, fmt.Errorf("%s: %w", msg, err)
+	}
+
+	// Combine possible definitions with events
+	aerModels := alertmanager.ConvertAmToAlarmEventRecordModels(request.Body, alarmDefinitions, a.ClusterServer.ClusterIDToResourceTypeID)
+
+	// Insert and update AlarmEventRecord
+	if err := a.AlarmsRepository.UpsertAlarmEventRecord(ctx, aerModels); err != nil {
```
mlguerrero12, line 389: I don't see why we need to upsert the current list. Above, you resolve the ones that are not in the list. That is good. Then, something that I'm missing is resolving the ones that explicitly come with firing status resolved. They are gonna be gone anyway in the next notification. But we don't know when this will happen. The `group_interval` states when a change is reported (new alert added to the list or one removed/resolved) but if nothing happens, this notification won't arrive when this timeout expires.

But anyway, the alerts are static, once they are reported, they are not gonna change. Only thing that changes is the firing status. You can perfectly identify with a query in the DB which alerts do not have a record in the DB. These are completely new, and you could simply add them. No need to process all of them

internal/service/alarms/cmd/serve.go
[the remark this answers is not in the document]
pixelsoccupied: yeah the problem is that rest of them are not is the package. Let's move this file under `api` dir and revert the name instead.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -363,11 +364,43 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+		slog.Error(msg, "error", err)
+		return nil, fmt.Errorf("%s: %w", msg, err)
+	}
+
+	//TODO: Get subscriber
+
+	//TODO: Notify subscriber
+
+	// Removed resolved events and move to archive
+	if err := a.AlarmsRepository.ArchiveResolvedAlarmEventRecords(ctx); err != nil {
```
mlguerrero12, line 400: Why do we need to archive the resolved alarms? At first, I thought this table was to archive the alerts that were purged. A place to have an extra backup for users to check. But you are moving everything that has been resolved. 

The Alarm Records Endpoints use the regular alarm_record table. There must be resolved and firing alarms for the users to query whatever they want using filters. With this, they won't be able to query resolved alerts. They should, these are supposed to be gone only after the retention period has expired.

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -63,3 +68,101 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+			ObjectID:          GetObjectID(*alert.Labels),
+			ObjectTypeID:      clusterIDToResourceTypeID[GetClusterID(*alert.Labels)],
+		}
+
+		// Update Extensions with things we didn't really process
+		record.Extensions = getExtensions(*alert.Labels, *alert.Annotations)
+
+		// See if possible to pick up additional info from its definition
+		for _, def := range aDefinitionRecords {
+			if def.AlarmName == GetAlertName(*alert.Labels) && def.ObjectTypeID == record.ObjectTypeID {
```
mlguerrero12, line 93: you need to include severity

internal/service/alarms/internal/alarms_server.go
browsell: There is a requirement to have an historical alarm DB

mlguerrero12: I understand that but that doesn't mean we should delete the resolved ones from the main Alarm Record table. Not before the retention period expires.

So, this method `ArchiveResolvedAlarmEventRecords` is deleting the resolved alarms from the main table `alarm_event_record` and then moving those to the archive table. This means that the main table will only have unresolved alarms. Our Alarm Record enpoints are using only the main alarm record table so queries for resolved alerts will be empty.

The job of archiving the resolved alerts should be done by the k8s Job we will implement in charge of purging the resolved alerts from the main table. Only then, they become purged and we could safely archive them without affecting expected workflow.

Another thing, the specification says that we need to purge resolved alarm and also active acknowledged alerts.

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -63,3 +68,101 @@ func Setup(ctx context.Context, cl client.Client) error {
 	slog.Info("Successfully configured alertmanager")
 	return nil
 }
+
+// ConvertAmToAlarmEventRecordModels get alarmEventRecords based on the alertmanager notification and AlarmDefinition
+func ConvertAmToAlarmEventRecordModels(am *api.AlertmanagerNotification, aDefinitionRecords []models.AlarmDefinition, clusterIDToResourceTypeID map[uuid.UUID]uuid.UUID) []models.AlarmEventRecord {
```
mlguerrero12, line 73: This goes better in the db/models/converters.go

internal/service/alarms/internal/alertmanager/alertmanager.yaml.template
```diff
@@ -19,13 +19,6 @@ route:
...
-  - receiver: oran_alarm_receiver
-    group_by:
-      # This label is guaranteed to be present in all alerts
-      - managed_cluster
-    matchers:
-      # Match alerts with a non-empty managed_cluster label
-      - managed_cluster=~".+"
 receivers:
   - name: oran_alarm_receiver
     webhook_configs:
```
mlguerrero12, line 24: should we delete send_resolved: true and rely only on the fact that resolved alerts are not present?

internal/service/common/utils/repository.go
```diff
@@ -101,8 +101,6 @@ func Delete[T db.Model](ctx context.Context, db DBQuery, whereExpr psql.Expressi
 		return 0, fmt.Errorf("failed to build delete query for '%s': %w", record.TableName(), err)
 	}
 
-	slog.Debug("executing statement", "sql", sql, "args", args)
```
mlguerrero12, line 104: I suppose you are deleting this because the bulk operations have a lot of values. However, by experience, it is good to have visibility to what we send to the DB. We will need them.

internal/service/common/utils/utils.go
```diff
@@ -174,3 +174,21 @@ func CompareObjects[T db.Model](a, b T, excluded ...string) DBTag {
 
 	return tags
 }
+
+// GetColumns gets the DB column names for the given fields
+func GetColumns[T db.Model](record T, fields []string) []string {
```
mlguerrero12, line 179: perhaps something like this func (r DBTag) Columns() []any  line 15 but for string

internal/service/common/utils/utils.go
```diff
@@ -174,3 +174,21 @@ func CompareObjects[T db.Model](a, b T, excluded ...string) DBTag {
...
+	tags := GetDBTagsFromStructFields(record, fields...)
+	columns := make([]string, len(fields))
+	for i, field := range fields {
+		columns[i] = tags[field]
+	}
+	return columns
+}
+
+func GetColumnsAsAny(slice []string) []any {
```
mlguerrero12, line 188: what we do is to call GetDBTagsFromStructFields with the fields you want and then .Columns

internal/service/cluster/collector/collector_k8s.go
```diff
@@ -18,9 +18,9 @@ import (
 )
 
 const vendorLabelName = "vendor"
-const openshiftVersionLabelName = "openshiftVersion"
-const clusterIDLabelName = "clusterID"
-const localClusterLabelName = "local-cluster"
+const OpenshiftVersionLabelName = "openshiftVersion"
```
mlguerrero12, line 21: we shouldn't take this from here. Nothing wrong to have our constants. But if you want something common for both. Use internal/controllers/utils/constants.go. We are adding common constants there. Some at the end of the file

internal/service/common/utils/repository.go
pixelsoccupied: yeah this was interesting issue. not sure if you tried this in the cluster...but basically after generating these large logs...slog log package effectively breaks. Like the server would start but no more logs are shown after a certain point. I was able to see the logs in crictl...but not on the console. 

yeah the logs are giant. like one call to upload the defs is basically few pages long.

internal/service/alarms/serve.go
```diff
@@ -201,3 +193,35 @@ func Serve(config *AlarmsServerConfig) error {
 
 	return nil
 }
+
+// UpdateAlarmDictionaryAndAlarmsDefinitionData reach out to cluster server and sync DB with alarm dict and def
+func UpdateAlarmDictionaryAndAlarmsDefinitionData(ctx context.Context, hubClient client.Client, a *internal.AlarmsServer) error {
```
mlguerrero12, line 198: I'm always in favor of refactoring code to improve stuff. But this is too large (the whole dictionary refactoring) to be part of a single commit. To make review easier, please consider having a separate commit for this. Allain has been merging stuff with multiple commits (as far as I know, there is a label to avoid the squash). Or create a subsequent PR for this.

Or if you want to leave it like this, let's try to converge on the main purpose of this pr first. Then, I will switch context and review this refactoring.

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -10,8 +10,8 @@ CREATE SEQUENCE IF NOT EXISTS alarm_sequence_seq
 CREATE TABLE IF NOT EXISTS alarm_event_record (
     -- O-RAN
     alarm_event_record_id UUID PRIMARY KEY DEFAULT gen_random_uuid(), -- Unique identifier for each event record
-    alarm_definition_id UUID NOT NULL, -- From alarm_definition table
-    probable_cause_id UUID NOT NULL, -- From alarm_definition table
+    alarm_definition_id UUID, -- From alarm_definition table
```
mlguerrero12, line 13: we've been trying to align this with the spec. These fields are not nullable according to O-CLOUD-IM.0-R004-v02.00

internal/service/alarms/internal/alarms_server.go
browsell: ok, so for the resolved alarms they should not be moved until retention period expires. 
I assume we would ignore these on a resync with the alert server ? 

As for the purge description in the spec. 
We cannot purge active acknowledged alarms as they still exist in the alert server, we would just readd on the next sync.

mlguerrero12: Correct. That's why I'm suggesting @pixelsoccupied  to identify which Alerts are already in the DB and simply ignore them instead of upserting the rows. If we are assuming that it is safe to resolve records that are not in the reported list from the Alert server (AlertManager), we could also assume that the ones that are present in the list and not in the DB are new ones. We could work only with these ones.

Regarding the active acknowledged alarms, that's a challenged indeed. Perhaps, we could also check the archive table to avoid the re triggering since you are mentioning, we don't control the Alert server (AlertManager). Those alerts will still be active. Or like you are suggesting, simply not purge them.

pixelsoccupied: yeah basically if the user provides a filter to look for archived/resolved data we would simply look into the archive table...otherwise we simply check the active data. 

I think original assumption behind this is we will keep one regular table with active events and the rest goes into the archive so that active table doesnt get too large. And archive table is what we delete at an internal. 

as an optimization we can also leverage DB so that they both appear as the same table such as union view (tbh since i have looked into the filter yet)

mlguerrero12: Why do we have to overdesign this? Let's keep it simple. This endpoint does not archive anything. The purge Job will move resolved alerts to the archive table when the retention period expires. With this, Alert Record endpoints will remain simple and use the main table. No tricks with multiple DBs. The archive table can have its own rentention period and its owns mean to serve queries (outside of the specification)

What do you think? @pixelsoccupied @browsell

pixelsoccupied: deleted

internal/service/common/utils/repository.go
mlguerrero12: I didn't see that issue. Do you also have it without the refactoring of the dicts you are implementing here?

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: yeah originally thought that but its cyclical import error (would have to move all the Am conversion code there as as well)

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: updated

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +149,229 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+	}
+
+	return utils.ExecuteCollectRows[models.AlarmDefinition](ctx, ar.Db, sql, params)
+}
+
+func getGetAlertNameAndObjectTypeID(am *api.AlertmanagerNotification, clusterIDToObjectTypeID map[uuid.UUID]uuid.UUID) []bob.Expression {
+	b := make([]bob.Expression, 0, len(am.Alerts))
+	for _, alert := range am.Alerts {
+		name := alertmanager.GetAlertName(*alert.Labels)
+		ObjectTypeID := clusterIDToObjectTypeID[alertmanager.GetClusterID(*alert.Labels)]
```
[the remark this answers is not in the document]
pixelsoccupied: updated

internal/service/common/utils/repository.go
pixelsoccupied: Yeah that was the main reason for refactoring actually. Originally i thought errors were not currently bubbling up and things werent working for me. So i started splitting it up see if there any issues there...but yeah turns out that crictl was showing the logs but app was basically still.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +149,229 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+
+// GetAlarmDefinitions needed to build out aer
+func (ar *AlarmsRepository) GetAlarmDefinitions(ctx context.Context, am *api.AlertmanagerNotification, clusterIDToObjectTypeID map[uuid.UUID]uuid.UUID) ([]models.AlarmDefinition, error) {
+	m := models.AlarmDefinition{}
+	dbTags := utils.GetAllDBTagsFromStruct(m)
+	query := psql.Select(
+		sm.Columns(
+			utils.GetColumnsAsAny(utils.GetColumns(m, []string{
+				"AlarmName", "AlarmDefinitionID", "ProbableCauseID",
+				"ObjectTypeID",
```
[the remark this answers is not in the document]
pixelsoccupied: updated

internal/service/cluster/collector/collector_k8s.go
pixelsoccupied: moved to utils

internal/service/alarms/serve.go
pixelsoccupied: yeah let's leave this in one commit since i have testings everything together. review it in one shot thanks! should be very minimal changes...simply extracting out the functions and using errgroup to capture any error from the routine

internal/service/alarms/internal/alarms_server.go
pixelsoccupied: Since dropping the archive table this may not be relevant anymore since we need to upsert the resolved data now.

internal/service/alarms/internal/alertmanager/alertmanager.yaml.template
pixelsoccupied: the main purpose is actually send the notification to user. So when AM calls us that means its time to alert the user  if alarm_sequence_number updates (in our case it would update when resolved or alarm_changed_time)

internal/service/alarms/internal/db/repo/alarms_repository.go
mlguerrero12: I don't see the changes

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +145,194 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+		columns = append(utils.GetColumns(records[0], []string{"AlarmEventRecordID"}), columns...)
+	}
+	query.Expression.Columns = columns
+
+	// Build values using functional approach
+	values := buildValues(records, hasID)
+	query.Apply(values...)
+
+	// Add upsert constraints for main table
+	if isMainTable(tableName) {
```
mlguerrero12, line 194: there is only one table now

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -39,21 +38,23 @@ CREATE TABLE IF NOT EXISTS alarm_event_record (
...
 
--- Function to update the alarm_sequence_number on specific status or time changes
-CREATE OR REPLACE FUNCTION update_alarm_event_sequence()
+CREATE OR REPLACE FUNCTION update_alarm_event_record()
     RETURNS TRIGGER AS $$
 BEGIN
-    -- Update sequence if status changes to 'resolved' or if alarm_changed_time is updated
-    IF (NEW.alarm_status = 'resolved' AND OLD.alarm_status IS DISTINCT FROM 'resolved')
-       OR (NEW.alarm_changed_time IS DISTINCT FROM OLD.alarm_changed_time) THEN
+    -- Update sequence and fields for resolved status
```
mlguerrero12, line 44: Please do this in code. We had some discussions with Allain about this and the conclusion was that triggers are a pain to maintain. Let's avoid them and do stuff in code

internal/service/alarms/api/server.go
```diff
@@ -363,13 +370,37 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+
+	// Get the definition data based on current set of Alert names and managed cluster ID
+	alarmDefinitions, err := a.AlarmsRepository.GetAlarmDefinitions(ctx, request.Body, a.ClusterServer.ClusterIDToResourceTypeID)
+	if err != nil {
+		msg := "failed to get AlarmDefinitions"
+		slog.Error(msg, "error", err)
+		return nil, fmt.Errorf("%s: %w", msg, err)
 	}
 
+	// Combine possible definitions with events
```
mlguerrero12, line 391: Please update cleared time, changed time and perceived severity of resolved alarms in code. Discussed this with @alegacy  on previous prs where I tried to add some triggers. The conclusion was that triggers are difficult to maintain and we should avoid them. It makes the logic more difficult to follow.

internal/service/common/utils/utils.go
pixelsoccupied: yeah the problem is that DbTags is a map which is not ordered. This way we can make sure the tags are coming out inorder

from 
```
tags := utils.GetDBTagsFromStructFields(dbModel, "AlarmDictionaryVersion", "EntityType", "Vendor", "ObjectTypeID")
columns := []string{tags["AlarmDictionaryVersion"], tags["EntityType"], tags["Vendor"], tags["ObjectTypeID"]}
```
to
```
columns := utils.GetColumns(dbModel, []string{"AlarmDictionaryVersion", "EntityType", "Vendor", "ObjectTypeID"})
```

internal/service/common/utils/utils.go
pixelsoccupied: Using slice instead to assert order

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: removed

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
pixelsoccupied: removed and using code

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +146,150 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+func (ar *AlarmsRepository) ResolveNotificationIfNotInCurrent(ctx context.Context, am *api.AlertmanagerNotification) error {
+	m := models.AlarmEventRecord{}
+	dbTags := utils.GetAllDBTagsFromStruct(m)
+	query := psql.Update(
+		um.Table(m.TableName()),
+		um.SetCol(dbTags["AlarmStatus"]).ToArg(api.Resolved),
+		um.SetCol(dbTags["AlarmClearedTime"]).ToArg(time.Now()),
+		um.SetCol(dbTags["PerceivedSeverity"]).ToArg(api.CLEARED),
+		um.Where(
+			psql.Group(psql.Quote(dbTags["Fingerprint"]), psql.Quote(dbTags["AlarmRaisedTime"])).
```
mlguerrero12, line 267: It seems to me that this is missing a conditional in the Where to only update rows that are not Resolved/Cleared. Otherwise, it will update the AlarmClearedTime with the current time on each notification

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -63,3 +68,114 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+
+		// Anything else that's not mentioned explicitly will be handled by DB such ID generation and default values as needed.
+		records = append(records, record)
+	}
+
+	return records
+}
+
+// finalOranSeverity set severity cleared
+func finalOranSeverity(ps api.PerceivedSeverity, status api.AlertmanagerNotificationStatus) api.PerceivedSeverity {
```
mlguerrero12, line 108: I don't think you need another level of abstraction for this. Simply after record

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -63,3 +68,114 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+		slog.Info("Converting Alertmanager alert", "alert name", GetAlertName(*alert.Labels))
+		oranPS, _ := GetPerceivedSeverity(*alert.Labels)
+		record := models.AlarmEventRecord{
+			AlarmRaisedTime:   *alert.StartsAt,
+			AlarmClearedTime:  alert.EndsAt,
+			AlarmChangedTime:  alert.StartsAt, //Note: this value is not available as part of the notification (only in v2 api as updatedAt)
+			PerceivedSeverity: finalOranSeverity(oranPS, *alert.Status),
+			AlarmStatus:       string(*alert.Status),
+			Fingerprint:       *alert.Fingerprint,
+			ObjectID:          GetObjectID(*alert.Labels),
```
mlguerrero12, line 85: why not calling GetClusterID directly?

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -21,9 +21,8 @@ CREATE TABLE IF NOT EXISTS alarm_event_record (
     extensions JSONB, -- Additional data for extensibility
 
     -- O-RAN additional data to create AlarmEventNotification
-    resource_id UUID NOT NULL, -- Same as manager_cluster_id for caas alerts
-    resource_type_id UUID NOT NULL, -- Derived from manager_cluster_id
-    notification_event_type INT NOT NULL, -- Should be enum calculated from current alert
+    object_id UUID, -- Same as manager_cluster_id for caas alerts
```
mlguerrero12, line 24: these are not nullable according to the spec and our api

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -63,3 +68,114 @@ func Setup(ctx context.Context, cl client.Client) error {
...
+		record.Extensions = getExtensions(*alert.Labels, *alert.Annotations)
+
+		// See if possible to pick up additional info from its definition
+		for _, def := range aDefinitionRecords {
+			if def.AlarmName == GetAlertName(*alert.Labels) && def.ObjectTypeID == record.ObjectTypeID && severityToPerceivedSeverity(def.Severity) == record.PerceivedSeverity {
+				record.AlarmDefinitionID = def.AlarmDefinitionID
+				record.ProbableCauseID = def.ProbableCauseID
+			}
+		}
+
```
mlguerrero12, line 99: Just to double check, if we don't find a clusterID, AlarmDefinitionID, ProbableCauseID, the record will have zeros on these fields, right?

internal/service/alarms/api/server.go
mlguerrero12: I'm not being pushing. If you don't want to add it, it's fine. But please just for my understanding because it seems I'm missing something.

The `change_time` is (copying from the spec) a Date/Time stamp value when any value of the AlarmEventRecord has been modified. The AlarmEventRecord after being added the first time, can only be modified when it is acknowledged or cleared. The acknowledge operation comes from the API. There, we update the change time. For Manual clearing type alarms, the operation also comes from the API. There, we also update the change time. Now, for this part, Automatic clearing type alarms, we are not setting it when we clear the alarm. Why? What is preventing us to do it? Again, it is just for my understanding not trying to impose anything.

And I don't think this value should come from AlertManager, the spec clearly says that it is only for values of the alarm record. In fact, as far as I know, the alertManager only updates an alert when it changes the firing status to Resolved. The EndsAt is also updated.

c.c. @browsell in case he has more info about this constraint.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -137,3 +146,150 @@ func (ar *AlarmsRepository) UpsertAlarmDefinitions(ctx context.Context, records
...
+	// Set upsert constraints
+	query.Apply(buildUpsertConstraints())
+
+	return query.Build() //nolint:wrapcheck
+}
+
+// buildUpsertConstraints creates the conflict resolution modifiers
+func buildUpsertConstraints() bob.Mod[*dialect.InsertQuery] {
+	dbTags := utils.GetAllDBTagsFromStruct(models.AlarmDefinition{})
+	return im.OnConflictOnConstraint("unique_fingerprint_alarm_raised_time").DoUpdate(
```
mlguerrero12, line 202: the db/model has a OnConflict method for this

internal/service/common/utils/repository.go
mlguerrero12: and what was the issue?

internal/service/alarms/api/server.go
```diff
@@ -363,13 +370,37 @@ func (a *AlarmsServer) GetProbableCause(ctx context.Context, request api.GetProb
...
+	aerModels := alertmanager.ConvertAmToAlarmEventRecordModels(request.Body, alarmDefinitions, a.ClusterServer.ClusterIDToResourceTypeID)
+
+	// Insert and update AlarmEventRecord
+	if err := a.AlarmsRepository.UpsertAlarmEventRecord(ctx, aerModels); err != nil {
```
mlguerrero12, line 395: Why the upsert? Say there is a large number of alert firing at a given time. We will have to update a large number of rows when a single alert is resolved or added. If you loop through the list, you can identify which alerts are Resolved and update them separately. And for new alerts, is the upsert more efficient than querying the DB to identify which are not present?

As followup question about upsert, how can we identify which alerts are New? The sequence on the table is only updated when the alert is resolved or change time has been modified (which should only happen when it is resolved). So, I don't see how we can identify which alerts were inserted by the upsert and not updated. This is important to notifiy (or not notify based on the filter) the users.

pixelsoccupied: Sequence is updated when a new row is added + as well as when the conditions are met. So need to insert+update the whole list to allow it auto update as needed. Along with status (like you mentioned)...some of the firing alerts may also new "changed" time or maybe somehow we added the ObjectTypeID that maybe missing previously. Taking one these one at a time is far simpler than to just let upsert do its thing.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: yeah

internal/service/alarms/api/server.go
mlguerrero12: ok, help me to understand. The sequence starts with 1, when it is added, it goes to the next value, which is 2. But again, how can you identify later that this is a new row? Other rows will also have value 2 when they are not updated. The update happens only when it is resolved because you are not changing alarm_changed_time anywhere.

pixelsoccupied: So basically the sequence `alarm_sequence_seq` is global. Meaning every time `nextval('alarm_sequence_seq')` is called (during insertion or update) no one row will have the same value. And then finally this value is used with subscriber's `event_cursor` to determine if that subscriber already been notified or not and update the subscriber cursor. I think the design doc has all the details + example code to explain it better. 

yeah alarm_changed_time is not updated since AM is not sending that value.

mlguerrero12: First of all, thank you for the explanation of how the sequence works. It makes sense now.

Only question that I have now is, why do you expect the AM to send the alarm_changed_time value? The spec clearly says that this value must be set when any value of the AlarmEventRecord has been modified. The AM does not update anything. All is control by us. We modify the record when we acknowledge and/or clear the alert. And now I understand that with the upsert, you want to add another layer of complexity for firing Alerts. You want to introduce for these the possibility to update the AlarmDefinitionID, ResourceID, ResourceTypeID. I don't know if we should add this new level of complexity. But fine. However for all this to work, we need the alarm_changed_time to be modified whenever we touch any of those fields. Otherwise, the sequence won't be updated. It won't be possible to detect a change.

And please, add more comments in the code, especially for the upsert. That you want to be possible to update some fields for current firing alerts.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: updated to check AlarmClearedTime is null or not. So only the null fields will have the timestamp.

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: removed

internal/service/alarms/internal/alertmanager/alertmanager.go
pixelsoccupied: updated

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
pixelsoccupied: so this is there to simply capture all the events regardless if managedcluster id is present. the idea is that we capture these issue now and then let ACM folks know about it instead of dropping any alerts.

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: updated

internal/service/alarms/api/server.go
pixelsoccupied: Ok let me add a trigger to keep track of the rows and use that as alarm_changed_time

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
```diff
@@ -0,0 +1,460 @@
...
+// deleteOutdatedDictionaries remove any dictionary that may not be available anymore
+func (r *AlarmDictionaryDefinition) deleteOutdatedDictionaries(ctx context.Context, nodeClusterTypes []clusterserver.NodeClusterType) error {
+	// Delete Dictionaries that do not have a corresponding resource type
+	ids := make([]any, 0, len(nodeClusterTypes))
+	for _, nodeClusterType := range nodeClusterTypes {
+		ids = append(ids, nodeClusterType.NodeClusterTypeId)
+	}
+
+	if err := r.AlarmsRepository.DeleteAlarmDictionariesNotIn(ctx, ids); err != nil {
+		return fmt.Errorf("failed to delete alarm dictionaries: %w", err)
```
mlguerrero12, line 223: even if for some reason we cannot delete outdated dictionaries, we can still sync the ones present - We had a warning here before

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
```diff
@@ -0,0 +1,460 @@
...
+		})
+	}
+	slog.Info("Waiting for all nodeClusterTypes to processed")
+	return g.Wait() //nolint:wrapcheck
+}
+
+// validateRulesExist assert we have rules to work with that will converted to alarm defs
+func (r *AlarmDictionaryDefinition) validateRulesExist(nodeClusterTypes []clusterserver.NodeClusterType) error {
+	var missingRules []uuid.UUID
+	for _, nct := range nodeClusterTypes {
```
mlguerrero12, line 253: This is different to what we had before. If for any reason, we fail to collect rules for a resource type we can simply warn and continue collecting the rules for the other resource types. We don't need to stop operation for a single one. Please add it back

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
```diff
@@ -0,0 +1,460 @@
...
+		Severity string
+	}
+
+	var filteredRules []monitoringv1.Rule
+	exist := make(map[uniqueAlarm]bool)
+
+	for _, rule := range r.RulesMap[nodeClusterTypeID] {
+		severity, ok := rule.Labels["severity"]
+		if !ok {
+			slog.Warn("rule missing severity label", "alert", rule.Alert, "nodeClusterTypeID", nodeClusterTypeID)
```
mlguerrero12, line 344: we shouldn't restrict this. Our application supports alerts without this as well.

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
```diff
@@ -0,0 +1,460 @@
...
+	defer cancel()
+
+	// First clean up outdated dictionaries
+	if err := r.deleteOutdatedDictionaries(ctx, nodeClusterTypes); err != nil {
+		return fmt.Errorf("failed to delete outdated dictionaries: %w", err)
+	}
+
+	// Then process all NodeClusterTypes concurrently
+	if err := r.processNodeClusterTypes(ctx, nodeClusterTypes); err != nil {
+		return fmt.Errorf("failed to process resource types: %w", err)
```
mlguerrero12, line 207: I would say that some node cluster types were not processed correctly since wait returns the first error but other routines could have successfully passed.

Another I made sure to change all reference (code and comment) of Resource Type to Node Cluster Type. It seems you made changes before that PR got merged. Could you please align them again?

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
pixelsoccupied: only warning now

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
pixelsoccupied: only warning now

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
pixelsoccupied: removed continue

internal/service/alarms/internal/dictionary_definition/dictionary_definition.go
pixelsoccupied: updated

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -57,3 +56,40 @@ CREATE OR REPLACE TRIGGER update_alarm_event_sequence
     BEFORE UPDATE ON alarm_event_record
     FOR EACH ROW
     EXECUTE FUNCTION update_alarm_event_sequence();
+
+-- Function to track alarm_changed_time
+CREATE OR REPLACE FUNCTION track_alarm_event_record_change_time()
```
mlguerrero12, line 61: As far as I know the triggers are executed in alphabetical order, so this works fine. But I usually see that a single function is created where these functions are called inside. To not depend on that. Perhaps, later

mlguerrero12 (commented):
Good work. Thanks @pixelsoccupied 
/lgtm
/approve

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
pixelsoccupied: yeah that's true. So in this case we would like `track_alarm_event_record_change_time` to execute before `update_alarm_event_sequence`. To make it prod friendly we would need to do something `CREATE TRIGGER ... AFTER TRIGGER another_trigger_name`. Will this change later;. Keeping it unresolved till then.

PR #360, opened 2024-11-28
Align postgres image var name with midstream process

alegacy:
This updates the name of the postgres image to have the RELATED_IMAGE_ prefix required by the midstream process.


bundle/manifests/oran-o2ims.clusterserviceversion.yaml
```diff
@@ -1558,5 +1558,8 @@ spec:
   minKubeVersion: 1.28.0
   provider:
     name: Red Hat
+  relatedImages:
```
rauhersu, line 1561: Just remove this 'relatedImages' config; our midstream overlay will generate all the 'relatedImages' we need; we just need the env var renamed as you did a couple of lines above. Thanks !

alegacy: @rauhersu  This is automatically added by the bundle generation.

PR #362, opened 2024-11-29
Fix alertmanager authorization failure

mlguerrero12:
CA certificate and Bearer token were missing in
alertmanager configuration

### DIFF
--- a/internal/controllers/inventory_controller.go
+++ b/internal/controllers/inventory_controller.go
@@ -69,6 +69,7 @@ import (
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="internal.open-cluster-management.io",resources=managedclusterinfos,verbs=get;list;watch
 //+kubebuilder:rbac:groups="config.openshift.io",resources=clusterversions,verbs=get;list;watch
+//+kubebuilder:rbac:urls="/internal/v1/caas-alerts/alertmanager",verbs=create;post
 
 // Reconciler reconciles a Inventory object
 type Reconciler struct {
@@ -352,6 +353,17 @@ func (t *reconcilerTask) setupAlarmServerConfig(ctx context.Context, defaultResu
 		t.logger.ErrorContext(
 			ctx,
 			"Failed to create alarm server cluster role binding",
+			slog.String("error", err.Error()),
+		)
+		return
+	}
+
+	// Needed for the Alertmanager to be able to send alerts to the alarm server
+	err = t.createAlertmanagerClusterRoleAndBinding(ctx)
+	if err != nil {
+		t.logger.ErrorContext(
+			ctx,
+			"Failed to create alertmanager cluster role and binding",
 			slog.String("error", err.Error()),
 		)
 		return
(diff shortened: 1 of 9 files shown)


internal/controllers/inventory_controller.go
```diff
@@ -1082,6 +1094,69 @@ func (t *reconcilerTask) createAlarmServerClusterRoleBinding(ctx context.Context
 	return nil
 }
 
+// createAlertmanagerClusterRoleAndBinding creates the cluster role and binding needed for the Alertmanager to be able to
+// send alerts to the alarm server.
+func (t *reconcilerTask) createAlertmanagerClusterRoleAndBinding(ctx context.Context) error {
+	name := "alertmanager"
+
+	// Cluster Role
+	role := &rbacv1.ClusterRole{
```
alegacy, line 1103: Since this is scoped to our namespace we should make this a `Role` instead of a `ClusterRole`

internal/controllers/inventory_controller.go
```diff
@@ -1082,6 +1094,69 @@ func (t *reconcilerTask) createAlarmServerClusterRoleBinding(ctx context.Context
...
+			},
+		},
+	}
+
+	if err := utils.CreateK8sCR(ctx, t.client, role, t.object, utils.UPDATE); err != nil {
+		return fmt.Errorf("failed to create Alertmanager cluster role: %w", err)
+	}
+
+	// Cluster Role Binding
+	binding := &rbacv1.ClusterRoleBinding{
```
alegacy, line 1129: Similar to above... this can be a `RoleBinding` since it is scoped to our namespace.

internal/controllers/inventory_controller.go
```diff
@@ -1082,6 +1094,69 @@ func (t *reconcilerTask) createAlarmServerClusterRoleBinding(ctx context.Context
...
+			Kind:     "ClusterRole",
+			Name: fmt.Sprintf(
+				"%s-%s",
+				t.object.Namespace, name,
+			),
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      rbacv1.ServiceAccountKind,
+				Namespace: "open-cluster-management-observability",
```
alegacy, line 1147: We should use a constant for this namespace name so it is not repeated in multiple places. ...same for the `alertmanager` name.

internal/controllers/inventory_controller.go
mlguerrero12: All examples I've found use clusterrole and clusterrolebinding. Non resource endpoints are not namescoped.
The official k8s rbac docs has:

Allow GET and POST requests to the non-resource endpoint /healthz and all subpaths (must be in a ClusterRole bound with a ClusterRoleBinding to be effective):

internal/controllers/inventory_controller.go
mlguerrero12: updated, thanks

internal/service/alarms/internal/alertmanager/alertmanager_test.go
```diff
@@ -53,6 +53,9 @@ var _ = Describe("Alertmanager", func() {
 			err = c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: secretName}, secret)
 			Expect(err).NotTo(HaveOccurred())
 			data := strings.Replace(string(alertManagerConfig), "{{ .url }}", utils.GetServiceURL(utils.InventoryAlarmServerName), 1)
+			data = strings.Replace(data, "{{ .caFile }}", utils.DefaultServiceCAFile, 1)
```
pixelsoccupied, line 56: I'm a little confused about this technique (please fill in my knowledge gap!). From what i understand...generally content under `/kubernetes.io/serviceaccount/serviceaccount/service-ca.crt ` is scoped to Kubernetes API server right?

And instead apps' certs are managed either the kubernetes native way e.g using [cert-manager](https://cert-manager.io/) or we create our own with `openssl req -x509...` (with subj being api) and mount it using secret and link the secret in tls under Ingres?

While works i wonder if it's the "right" approach? Curious to read more on this!

alegacy: Each service account gets a pair of cert/key (if they request one).  For every service account they are all signed by the same CA.  Usually that's a default self-signed internal CA, but the end user can override that if they want (that's not relevant here because the mechanism is the same).  When the deployment (with a service account) is launched k8s auto-injects the internal root CA into a volume at the location specified by that constant.  The pod can then use that file to validate any certificate of any service account in the system.   Since we are configuring the alert manager to invoke another service account's API endpoints (us) then here we are telling it to go and use its auto-injected CA file to verify our identity. 

If we weren't another service account running on the same system then we'd either need to have our cert signed by a public CA that the alert manager already knew about or we'd have to tell the alert manager about our private CA.

pixelsoccupied: > we'd have to tell the alert manager about our private CA. 

Ah yeah that's exactly what i was thinking. ACM's observability does allow for 
https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.12/html-single/observability/index#mount-secrets-alertmanager

So we with this no need for a key file. Also does that mean MultiClusterObservability config don't need to be changed? 

```
advanced:
 alertmanager:
   secrets: ['tls']
 ```

PR #364, opened 2024-11-29
CNF-15750: Add DB and API handling for new resource endpoints

alegacy:
This adds the initial DB and API handling for the new resource server implementation.  This covers the Subscription and Deployment Manager API endpoints.  More to follow.

### DIFF
--- a/internal/service/common/utils/repository.go
+++ b/internal/service/common/utils/repository.go
@@ -1,0 +1,98 @@
+package utils
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stephenafamo/bob/dialect/psql"
+	"github.com/stephenafamo/bob/dialect/psql/dm"
+	"github.com/stephenafamo/bob/dialect/psql/sm"
+
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/db"
+)
+
+// Find retrieves a specific tuple from the database table specified.  If no record is found an empty array is returned.
+func Find[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) ([]T, error) {
+	// Build sql query
+	var record T
+	tags := GetAllDBTagsFromStruct(record)
+
+	query, args, err := psql.Select(
+		sm.Columns(tags.Columns()...),
+		sm.From(record.TableName()),
+		sm.Where(psql.Quote(record.PrimaryKey()).EQ(psql.Arg(uuid))),
+	).Build()
+	if err != nil {
+		return []T{}, fmt.Errorf("failed to build query: %w", err)
+	}
+
+	// Run query
+	rows, _ := db.Query(ctx, query, args...) // note: err is passed on to Collect* func so we can ignore this
+	record, err = pgx.CollectExactlyOneRow(rows, pgx.RowToStructByNameLax[T])
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			slog.Info("No Entity found", "uuid", uuid, "table", record.TableName())
+			return []T{}, nil
+		}
+		return []T{}, fmt.Errorf("failed to call database: %w", err)
+	}
+
+	slog.Info("record found", "table", record.TableName(), "uuid", uuid)
+	return []T{record}, nil
+}
+
+// FindAll retrieves all tuples from the database table specified.  If no records are found then an empty array is
+// returned.
+func FindAll[T db.Model](ctx context.Context, db *pgxpool.Pool) ([]T, error) {
+	// Build sql query
+	var record T
+	var records []T
+	tags := GetAllDBTagsFromStruct(record)
+
+	query, args, err := psql.Select(
+		sm.Columns(tags.Columns()...),
+		sm.From(record.TableName()),
+	).Build()
+	if err != nil {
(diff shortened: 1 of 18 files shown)


mlguerrero12 (commented):
some initial comments

internal/service/common/utils/repository.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/stephenafamo/bob/dialect/psql"
+	"github.com/stephenafamo/bob/dialect/psql/dm"
+	"github.com/stephenafamo/bob/dialect/psql/sm"
+
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/db"
+)
+
+// Find retrieves a specific tuple from the database table specified.  If no record is found an empty array is returned.
+func Find[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) ([]T, error) {
```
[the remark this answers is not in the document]
alegacy: Ya, that's how I set it up initially but it causes a linter error to `return nil, nil` when no data is found.

internal/service/resources/db/models/converters.go
```diff
@@ -0,0 +1,82 @@
...
+		Filter:                 record.Filter,
+		SubscriptionId:         record.SubscriptionID,
+	}
+
+	return object, nil
+}
+
+// SubscriptionFromModel converts a API model to a DB tuple
+func SubscriptionFromModel(object *generated.Subscription) (*Subscription, error) {
+	id, err := uuid.NewRandom()
```
mlguerrero12, line 68: the database can generate this for us

internal/service/resources/api/server.go
```diff
@@ -3,69 +3,355 @@ package api
...
+			AdditionalAttributes: &map[string]string{
+				"consumerSubscriptionId": consumerSubscriptionId,
+			},
+			Detail: err.Error(),
+			Status: http.StatusInternalServerError,
+		}, nil
+	}
+
+	// Set internal fields
+	record.CreatedAt = time.Now()
```
mlguerrero12, line 159: also this one can be created by the DB

internal/service/resources/api/server.go
```diff
@@ -3,69 +3,355 @@ package api
...
+		return api.CreateSubscription500ApplicationProblemPlusJSONResponse{
+			AdditionalAttributes: &map[string]string{
+				"consumerSubscriptionId": consumerSubscriptionId,
+			},
+			Detail: err.Error(),
+			// TODO: map errors to 400 if possible; else 500
+			Status: http.StatusInternalServerError,
+		}, nil
+	}
+
```
mlguerrero12, line 174: you can use returning in the INSERT query to get what you just inserted

internal/service/resources/api/server.go
```diff
@@ -3,69 +3,355 @@ package api
...
+	return api.GetSubscription200JSONResponse(object), nil
 }
 
-func (r ResourceServer) GetResourcePool(ctx context.Context, request api.GetResourcePoolRequestObject) (api.GetResourcePoolResponseObject, error) {
+// DeleteSubscription receives the API request to this endpoint, executes the request, and responds appropriately
+func (r *ResourceServer) DeleteSubscription(ctx context.Context, request api.DeleteSubscriptionRequestObject) (api.DeleteSubscriptionResponseObject, error) {
+	records, err := r.Repo.GetSubscription(ctx, request.SubscriptionId)
```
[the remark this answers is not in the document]
alegacy: Because if a client requests an action on a resource that doesn't exist the typical response is a 404.

internal/service/resources/db/repo/repository.go
```diff
@@ -1,8 +1,100 @@
...
+	if subscription.Filter != nil {
+		columns = append(columns, "filter")
+		values = append(values, *subscription.Filter)
+	}
+
+	// Add columns and corresponding values
+	query.Expression.Columns = columns
+	query.Apply(im.Values(psql.Arg(values...)))
+
+	sql, args, err := query.Build()
```
mlguerrero12, line 68: you can use returning in the query to get what you just inserted

internal/service/resources/db/models/converters.go
alegacy: Yes, I now, but there are use cases later where I will be creating this based on named UUID values so I prefer to keep it here.

internal/service/resources/api/server.go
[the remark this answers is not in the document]
mlguerrero12: I've had some discussions about this in the past. But fine, if we decide to go for 404, you could imply that from the delete if the SQL DELETE return 0 or if you use returning in the Query to get the ID for example

internal/service/resources/db/models/converters.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+
+	"github.com/openshift-kni/oran-o2ims/internal/service/resources/api/generated"
+)
+
+// DeploymentManagerToModel converts a DB tuple to an API Model
+func DeploymentManagerToModel(record *DeploymentManager) (generated.DeploymentManager, error) {
```
mlguerrero12, line 12: I would delete the error here and in all the XToY functions. It adds unneeded code (checking error that is always nil) in server.go

alegacy: I can remove them for now, but i expect that some of them will need the error to convert the enums and handled unexpected value.  ...for now I agree this is unnecessary.  ...i'll deal with it in the next PR.

mlguerrero12: this will be addressed in a future PR (discussed internally)

PR #366, opened 2024-12-02
CNF-15750: Add additional API and DB resource models

alegacy:
This adds the remaining set of API and DB models for the resource related entities.

### DIFF
--- a/internal/service/alarms/internal/alarms_server.go
+++ b/internal/service/alarms/internal/alarms_server.go
@@ -2,11 +2,14 @@ package internal
 
 import (
 	"context"
+	"errors"
 
 	"log/slog"
 
 	"fmt"
 	"net/http"
+
+	"github.com/jackc/pgx/v5"
 
 	api "github.com/openshift-kni/oran-o2ims/internal/service/alarms/api/generated"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/db/models"
@@ -57,12 +60,10 @@ func (a *AlarmsServer) GetAlarms(ctx context.Context, request api.GetAlarmsReque
 // GetAlarm returns an AlarmEventRecord with a given ID
 func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequestObject) (api.GetAlarmResponseObject, error) {
 	aerModel, err := a.AlarmsRepository.GetAlarmEventRecordWithUuid(ctx, request.AlarmEventRecordId)
-	if err != nil {
+	if !errors.Is(err, pgx.ErrNoRows) {
 		return nil, fmt.Errorf("failed to get AlarmEventRecord due to issues with DB conn: %w", err)
-	}
-
-	// Nothing found
-	if len(aerModel) == 0 {
+	} else if err != nil {
+		// Nothing found
 		return api.GetAlarm404ApplicationProblemPlusJSONResponse(common.ProblemDetails{
 			AdditionalAttributes: &map[string]string{
 				"UUID": request.AlarmEventRecordId.String(),
@@ -72,7 +73,7 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
 		}), nil
 	}
 
-	return api.GetAlarm200JSONResponse(convertAerModelToApi(aerModel[0])), nil
+	return api.GetAlarm200JSONResponse(convertAerModelToApi(*aerModel)), nil
 }
 
 func (a *AlarmsServer) AckAlarm(ctx context.Context, request api.AckAlarmRequestObject) (api.AckAlarmResponseObject, error) {
--- a/internal/service/alarms/internal/alarms_server.go
+++ b/internal/service/alarms/internal/alarms_server.go
@@ -9,13 +9,12 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/jackc/pgx/v5"
-
 	api "github.com/openshift-kni/oran-o2ims/internal/service/alarms/api/generated"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/db/models"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/db/repo"
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/resourceserver"
 	common "github.com/openshift-kni/oran-o2ims/internal/service/common/api/generated"
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/utils"
 )
 
 type AlarmsServer struct {
@@ -60,9 +59,7 @@ func (a *AlarmsServer) GetAlarms(ctx context.Context, request api.GetAlarmsReque
 // GetAlarm returns an AlarmEventRecord with a given ID
 func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequestObject) (api.GetAlarmResponseObject, error) {
 	aerModel, err := a.AlarmsRepository.GetAlarmEventRecordWithUuid(ctx, request.AlarmEventRecordId)
-	if !errors.Is(err, pgx.ErrNoRows) {
-		return nil, fmt.Errorf("failed to get AlarmEventRecord due to issues with DB conn: %w", err)
-	} else if err != nil {
+	if errors.Is(err, utils.ErrNotFound) {
 		// Nothing found
 		return api.GetAlarm404ApplicationProblemPlusJSONResponse(common.ProblemDetails{
 			AdditionalAttributes: &map[string]string{
@@ -71,6 +68,8 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
 			Detail: "Could not find AlarmEventRecord for given UUID",
 			Status: http.StatusNotFound,
 		}), nil
+	} else if err != nil {
+		return nil, fmt.Errorf("failed to get AlarmEventRecord due to issues with DB conn: %w", err)
 	}
 
 	return api.GetAlarm200JSONResponse(convertAerModelToApi(*aerModel)), nil
--- a/internal/service/common/utils/repository.go
+++ b/internal/service/common/utils/repository.go
@@ -40,7 +40,7 @@ func Find[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID, col
 	record, err = pgx.CollectExactlyOneRow(rows, pgx.RowToStructByNameLax[T])
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
-			slog.Info("No Entity found", "uuid", uuid, "table", record.TableName())
+			slog.Info("no record found", "uuid", uuid, "table", record.TableName())
 			return []T{}, nil
 		}
 		return []T{}, fmt.Errorf("failed to call database: %w", err)
@@ -74,6 +74,7 @@ func FindAll[T db.Model](ctx context.Context, db *pgxpool.Pool, columns []any) (
 	records, err = pgx.CollectRows(rows, pgx.RowToStructByNameLax[T])
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
+			slog.Info("no records found", "table", record.TableName())
 			return []T{}, nil
 		}
 		return []T{}, fmt.Errorf("failed to call database: %w", err)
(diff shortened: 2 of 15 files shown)


internal/service/common/utils/repository.go
```diff
@@ -102,3 +103,33 @@ func Delete[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) (
...
+	record, err = pgx.CollectExactlyOneRow(rows, pgx.RowToStructByNameLax[T])
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			slog.Info("No Entity found", "expression", expression, "table", record.TableName())
+			return []T{}, nil
+		}
+		return []T{}, fmt.Errorf("failed to call database: %w", err)
+	}
+
+	slog.Info("records found", "table", record.TableName(), "expression", expression)
```
mlguerrero12, line 133: just realized that this log message and the ones for the functions above are misleading. Record can be an empty string.

internal/service/resources/api/server.go
```diff
@@ -256,32 +212,158 @@ func (r *ResourceServer) DeleteSubscription(ctx context.Context, request api.Del
...
+	object := models.ResourcePoolToModel(&records[0])
+	return api.GetResourcePool200JSONResponse(object), nil
 }
 
 // GetResources receives the API request to this endpoint, executes the request, and responds appropriately
 func (r *ResourceServer) GetResources(ctx context.Context, request api.GetResourcesRequestObject) (api.GetResourcesResponseObject, error) {
-	// TODO implement me
-	return nil, fmt.Errorf("not implemented")
+	// First, find the pool
+	pools, err := r.Repo.GetResourcePool(ctx, request.ResourcePoolId)
```
mlguerrero12, line 258: what about using select exists for this? We just need a quick way to check that it is there
https://github.com/jackc/pgx/discussions/1427

internal/service/common/utils/repository.go
alegacy: I feel as though returning an error when no result is found is a bit heavy handed.  Doing a select and having an empty result set is not an error.  It is simply an indication that there is no data or that the search criteria was too narrow.   This is how a normal "select * from table where ...;' works in any DB implementation.

internal/service/resources/api/server.go
alegacy: Not sure I follow the intent of your comment.  Are you proposing that we add a different repo API that essentially is a `exists, err := ResourceExists(ctx, request.ResourcePoolId)`

internal/service/common/utils/repository.go
alegacy: > just realized that this log message and the ones for the functions above are misleading. Record can be an empty string.

I've added a change to adjust some of the logs.   Not sure I completely understand your concern here though.

internal/service/resources/api/server.go
alegacy: If so, I'd rather tackle that at another time.

internal/service/common/utils/repository.go
mlguerrero12: The thing is that when I saw `Search` I started to think that `Find` is not ok. 

So, there are 2 functions from pgx, `CollectExactlyOneRow` which returns an error when the `SELECT` does not return rows and `CollectRows` which does not return an error when nothing is returned (I don't know why we are checking `pgx.ErrNoRows` here in `Search`and `FindAll`. Only returned by `CollectExactlyOneRow`),

For `Find`, we decided to use `CollectExactlyOneRow`. When no rows are returned, we ignore the error and return an empty slice. If there are no errors, that means that it is guaranteed to have a `Record`. However, we put it in a slice which forces us to check again outside if the slice returned by `Find` has length zero. We have to check if `Find` returns and error and then if the length of the slice is zero. And this is very odd to me. The verb `Find` should return an error when no rows are found, in my opinion.

But fine, if we insists on having this on `Find`, then we could use `CollectRows` instead of `CollectExactlyOneRow` because we are checking the length of the slice outside anyway. And if we do this, `Find` is very similar to `Search`. So, `Find` could define a `Quote` to search for the primary key and then call `Search` to avoid duplicating code.

internal/service/resources/api/server.go
mlguerrero12: No, that's not what I meant. Here and in the function below, you are checking if the `pool` exists. You do that by using `Search` which fetches all the columns. Nothing is done with the data of the pool.

So, what I try to say is that there is a more efficient way to do this. We could create an utility function which checks with `SELECT EXISTS` if an object exists

internal/service/common/utils/repository.go
alegacy: fixed

internal/service/resources/api/server.go
alegacy: fixed

internal/service/alarms/internal/alarms_server.go
```diff
@@ -57,12 +60,10 @@ func (a *AlarmsServer) GetAlarms(ctx context.Context, request api.GetAlarmsReque
 // GetAlarm returns an AlarmEventRecord with a given ID
 func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequestObject) (api.GetAlarmResponseObject, error) {
 	aerModel, err := a.AlarmsRepository.GetAlarmEventRecordWithUuid(ctx, request.AlarmEventRecordId)
-	if err != nil {
```
mlguerrero12, line 60: I think it is more natural to check first if the error is different than nil and then check if it is a specific error

internal/service/alarms/internal/alarms_server.go
```diff
@@ -57,12 +60,10 @@ func (a *AlarmsServer) GetAlarms(ctx context.Context, request api.GetAlarmsReque
...
+	if !errors.Is(err, pgx.ErrNoRows) {
```
mlguerrero12, line 63: pgx.ErrNoRows is not returned but the custom error you created in common/utils/repository

internal/service/alarms/internal/alarms_server.go
mlguerrero12: Yeah, I've had discussions about this in the past. It's a matter of personal taste I would say. Personally, I like to check nil first and then the specific error inside. This is fine as well of  course(although, I've usually seen it without the else because it doesn't add anything, just if != nil below).

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
 // FindAll retrieves all tuples from the database table specified.  If no records are found then an empty array is
 // returned.
 func FindAll[T db.Model](ctx context.Context, db *pgxpool.Pool, columns []any) ([]T, error) {
+	var record T
+	e := psql.Quote(record.PrimaryKey()).IsNotNull()
+	return Search[T](ctx, db, e, columns)
+}
+
+// Delete deletes a specific tuple from the database table specified.  If no matching tuples are found an error will be
+// returned therefore the caller is responsible for checking for existing records.
```
mlguerrero12, line 71: this comment is not ok - I'm fixing it in my PR, so we can leave it as it is for now

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
-	return []T{record}, nil
+	slog.Debug("record found", "table", record.TableName(), "uuid", uuid)
+	return &record, nil
 }
 
 // FindAll retrieves all tuples from the database table specified.  If no records are found then an empty array is
 // returned.
 func FindAll[T db.Model](ctx context.Context, db *pgxpool.Pool, columns []any) ([]T, error) {
+	var record T
+	e := psql.Quote(record.PrimaryKey()).IsNotNull()
```
mlguerrero12, line 66: this is redudant - primary keys always have not null and unique constraints

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
 
+	slog.Debug("executing statement", "sql", sql, "args", args)
+
 	// Run query
-	rows, _ := db.Query(ctx, query, args...) // note: err is passed on to Collect* func so we can ignore this
-	records, err = pgx.CollectRows(rows, pgx.RowToStructByNameLax[T])
+	rows, _ := db.Query(ctx, sql, args...) // note: err is passed on to Collect* func so we can ignore this
+	records, err := pgx.CollectRows(rows, pgx.RowToStructByNameLax[T])
 	if err != nil {
 		if errors.Is(err, pgx.ErrNoRows) {
```
mlguerrero12, line 118: CollectRows does not return pgx.ErrNoRows when there are no rows

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
-// Delete deletes a specific tuple from the database table specified.  If no matching tuples are found an error will be
-// returned therefore the caller is responsible for checking for existing records.
-func Delete[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) (int64, error) {
+// Exists checks whether a record exists in the table with a primary key that matches uuid.
+func Exists[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) (bool, error) {
 	var record T
-	query := psql.Delete(
-		dm.From(record.TableName()),
-		dm.Where(psql.Quote(record.PrimaryKey()).EQ(psql.Arg(uuid))))
+	query := fmt.Sprintf("SELECT EXISTS(SELECT 1 FROM %s WHERE %s=%s)",
```
mlguerrero12, line 132: bob has psql.RawQuery

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
 	}
 
-	result, err := db.Exec(ctx, sql, params...)
+	return result, nil
+}
+
+// Create creates a record of the requested model type.  The stored record is returned on success; otherwise an error
+// is returned.
+func Create[T db.Model](ctx context.Context, db *pgxpool.Pool, record T) (*T, error) {
+	nonNilTags := GetAllDBTagsFromStruct(record, ExcludeNilValues)
```
mlguerrero12, line 149: I find this very confusing. Besides, it is only used in Create. It is polluting everything else. Why don't we simply accept columns and values as arguments? or if you insist on keeping this, why don't you create another function to get the tags whose values are not nil instead of adding an argument to GetAllDBTagsFromStruct? It's much cleaner in my opinion.

internal/service/common/utils/repository.go
```diff
@@ -9,96 +9,212 @@ import (
...
+	tags := GetAllDBTagsFromStruct(record, IncludeNilValues)
```
mlguerrero12, line 150: I see that tags is only used for the Returning, you could use * instead

internal/service/common/utils/utils.go
```diff
@@ -7,6 +7,10 @@ import (
 )
 
 type DBTag map[string]string
+type DBValue map[string]interface{}
```
mlguerrero12, line 10: I don't see DBValue used anywhere?

internal/service/common/utils/utils.go
```diff
@@ -7,6 +7,10 @@ import (
...
+
+const IncludeNilValues = false
```
mlguerrero12, line 12: these are confusing, better to create a specific function

internal/service/common/utils/utils.go
```diff
@@ -18,22 +22,77 @@ func (r DBTag) Columns() []any {
...
+			}
+		}
 	}
 
 	return tags
 }
 
+// GetColumnsAndValues returns the list of values associated to the field names specified in the tags parameter.  Both the
+// columns and the values are returned together to ensure that they are aligned.
+func GetColumnsAndValues[T db.Model](s T, tags DBTag) ([]string, []any) {
```
mlguerrero12, line 67: I like this function

internal/service/common/utils/repository.go
alegacy: I'll rework this so that there are 2 public functions that don't have this flag, but internally they will re-use the same code rather than duplicate it since it is mostly the same.

alegacy: fixed

internal/service/common/utils/utils.go
alegacy: removed

internal/service/common/utils/utils.go
alegacy: restructured this a bit so still reuse a common function but to not expose these options.

PR #374, opened 2024-12-04
CNF-15570: Add new resource server args to config

alegacy:
Adds the command line arguments required by the resource server to achieve parity with the previous implementation of the resource server.


alegacy:
closed as combined with https://github.com/openshift-kni/oran-o2ims/pull/379

PR #375, opened 2024-12-05
CNF-15570: Refactor pool fetcher and handler

alegacy:
This moves certain parts of the resource pool fetcher into the handler so that it comparable to the structure/relationship between between the resource fetcher/handler.  This is to facilitate using both handlers directly from the new resource server implementation without needing to add an adaptation layer over them.

### DIFF
--- a/internal/service/resource_pool_fetcher.go
+++ b/internal/service/resource_pool_fetcher.go
@@ -23,17 +23,11 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
-	"strings"
 
 	"github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 	"github.com/openshift-kni/oran-o2ims/internal/data"
-	"github.com/openshift-kni/oran-o2ims/internal/graphql"
-	"github.com/openshift-kni/oran-o2ims/internal/jq"
 	"github.com/openshift-kni/oran-o2ims/internal/k8s"
 	"github.com/openshift-kni/oran-o2ims/internal/model"
-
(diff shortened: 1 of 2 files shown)


internal/service/resource_pool_fetcher.go
```diff
@@ -192,33 +160,29 @@ func (b *ResourcePoolFetcherBuilder) Build() (
...
-	clusters, err := k8s.NewStream().
+	clusters, err = k8s.NewStream().
 		SetLogger(r.logger).
 		SetReader(response).
 		Build()
 	if err != nil {
 		return
 	}
 
-	// Transform Clusters to ResourcePools
```
mlguerrero12, line 219: why deleting this and adding them outside? You are changing the whole purpose of this method.

alegacy: I am changing this because it was done differently in the resource pool fetcher vs the resource fetcher.  The resource fetcher did a good job of separating the concerns of collecting the data (fetcher) and mapping the data to the API model (handler).  The resource pool fetch did not separate those concerns and did both tasks in the same piece of code.  I intend on re-using the fetchers to pull the data into the new resource server persisted storage and the mapping to the old API model is an unnecessary step so I am moving it into the handler where it belongs.   Once we switch over to the new resource server then the handlers will all be deleted.

mlguerrero12: I see, so all this will be deleted then. No need to modify name of method or go docs to align to what it is doing now.

PR #376, opened 2024-12-05
CNF-15570: Add support for updating only some fields

alegacy:
This makes changes to the Update utility so that it is possible to request that only some fields be set in the update operation.  This is to facilitate changes required by the upcoming resource data collector functionality that will be synchronizing data between ACM and our persistent storage.

### DIFF
--- a/internal/service/common/utils/repository.go
+++ b/internal/service/common/utils/repository.go
@@ -151,9 +151,10 @@ func Exists[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) (
 // is returned.
 func Create[T db.Model](ctx context.Context, db *pgxpool.Pool, record T) (*T, error) {
 	nonNilTags := GetNonNilDBTagsFromStruct(record)
+	tags := GetAllDBTagsFromStruct(record)
 
 	// Return all columns to get any defaulted values that the DB may set
-	query := psql.Insert(im.Into(record.TableName()), im.Returning("*"))
+	query := psql.Insert(im.Into(record.TableName()), im.Returning(tags.Columns()...))
 
 	// Add columns to the expression.  Maintain the order here so that it coincides with the order of the values
 	columns, values := GetColumnsAndValues(record, nonNilTags)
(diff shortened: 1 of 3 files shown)


internal/service/common/utils/repository.go
```diff
@@ -151,9 +151,10 @@ func Exists[T db.Model](ctx context.Context, db *pgxpool.Pool, uuid uuid.UUID) (
 // is returned.
 func Create[T db.Model](ctx context.Context, db *pgxpool.Pool, record T) (*T, error) {
 	nonNilTags := GetNonNilDBTagsFromStruct(record)
+	tags := GetAllDBTagsFromStruct(record)
 
 	// Return all columns to get any defaulted values that the DB may set
-	query := psql.Insert(im.Into(record.TableName()), im.Returning("*"))
+	query := psql.Insert(im.Into(record.TableName()), im.Returning(tags.Columns()...))
```
alegacy, line 157: Putting this back in because it makes the code too fragile.  If we add columns to the table, but forget to add them to the model it causes an error.   Until we have a better unit test strategy we should avoid making this more fragile than necessary.

internal/service/common/utils/repository.go
```diff
@@ -182,16 +183,18 @@ func Create[T db.Model](ctx context.Context, db *pgxpool.Pool, record T) (*T, er
...
 
 // Update attempts to update a record with a matching primary key.  The stored record is returned on success; otherwise
-// an error is returned.
-func Update[T db.Model](ctx context.Context, db *pgxpool.Pool, record T, uuid uuid.UUID) (*T, error) {
-	tags := GetAllDBTagsFromStruct(record)
+// an error is returned.  If the `fields` argument is set then only those columns are updated but the returned object
+// will contain all columns.
+func Update[T db.Model](ctx context.Context, db *pgxpool.Pool, record T, uuid uuid.UUID, fields ...string) (*T, error) {
+	all := GetAllDBTagsFromStruct(record)
+	tags := GetDBTagsFromStructFields(record, fields...)
```
mlguerrero12, line 190: this is how I imaged some specific columns to be selected in Find, FindAll and Search, instead of passing []any for columns

alegacy: ok, I'll rework those other ones in a future PR.

internal/service/common/utils/utils.go
```diff
@@ -21,6 +21,16 @@ func (r DBTag) Columns() []any {
 	return columns
 }
```
mlguerrero12, line 23: I don't see all this being used anywhere. Let's keep this pr clean and introduce `CompareObjects` later when is used

alegacy: I did this intentionally to reduce the size of the upcoming PRs.  I wanted to avoid one huge PR with everything in it since it makes it difficult to review.

mlguerrero12: I understand that and I'm not trying to be picky. It is just a good practice. These parts are not too big. Let's introduce them later.

PR #379, opened 2024-12-05
CNF-15570: Add new resource servers args and update some utils

alegacy:
This adds command line argument handling for the new resource server, and updates some utilities to simplify how they are invoked by converting optional arguments to variadic arguments.

The PR is broken down into a few commits to separate chunks for functionality.


internal/service/common/utils/utils.go
```diff
@@ -115,3 +125,52 @@ func GetDBTagsFromStructFields[T db.Model](s T, fields ...string) DBTag {
 
 	return tags
 }
+
+// CompareObjects compares two objects and returns the tags for fields having differing values.  Any field names listed
+// in `excluded` are ignored.
+func CompareObjects[T db.Model](a, b T, excluded ...string) DBTag {
```
mlguerrero12, line 131: I can see where this function is used in subsequent commits. I have many concerns/questions there.
I think we are abusing of reflect and generics. I see that you use it to update only some fields but I wonder if it is worth the trouble. As far as I know, postgres is a mvcc db which means an update is an insert-delete underneath. I know, there is some gain by not sending everything, but is it really too much?  How often do you expect to have changes.

And if performance is a concern here, then I see you are using the slowest approach, updating an object at a time. CopyFrom is the fastest, then single query with multiple rows, after that batch (multiiple queries inside) and finally one row at a time. https://github.com/jackc/pgx/discussions/1545

Also, postgres has an upsert verb that can be used instead of checking in code if an insert or update is needed.

I'm almost done reviewing the first 3 commits of this PR. I suggest you split this pr in 2, one with the first 3 commits so that we can merge them today. The last 3 will require more time. If you do that, please remove this function from the "CNF-15570: Add support for updating only some fields" commit

alegacy: I don't anticipate that the volume of updates will be high.  I was following best practices of only updating necessary fields.  This is what most ORMs do to avoid overloading the database with updating unnecessary fields.   We can certainly remove that, but from a debug point of view it makes it difficult to tell why something is being updated if everything is being written on all updates.

alegacy: I should qualify my comment about the volume of updates.  As the code is written now every record will get updated to set the generation id to track stale records.   I could switch to a different approach to manage that and we wouldn't need as many updates.

alegacy: @mlguerrero12 as per your request I have split the PR.  The other one with the top 3 commits (plus a new one for the subscription notification handling) is over [here](https://github.com/openshift-kni/oran-o2ims/pull/384).

alegacy:
/hold

would like to merge this manually without squashing

internal/service/resources/cmd/serve.go
```diff
@@ -1,25 +1,98 @@
...
+	flags.StringVar(
+		&config.GlobalCloudID,
+		server.GlobalCloudIDFlagName,
+		utils.DefaultOCloudID,
+		"The global O-Cloud identifier.",
+	)
+
+	// The O-Cloud ID and External address arguments are mandatory while all other arguments are optional.  The Global
+	// O-Cloud ID is special in that it is not strictly mandatory to start the server but it is mandatory to enable
+	// some API endpoints (e.g., subscriptions).
```
mlguerrero12, line 85: why is it needed to enable some API endpoints?

internal/service/resources/cmd/serve.go
```diff
@@ -1,25 +1,98 @@
...
+	flags.StringArrayVar(
+		&config.Extensions,
+		server.ExtensionsFlagName,
+		[]string{},
+		"Extensions to add to resources and resource pools.",
+	)
+	flags.StringVar(
+		&config.ExternalAddress,
+		server.ExternalAddressFlagName,
+		"string",
```
mlguerrero12, line 73: shouldn't be empty?

internal/service/resources/api/openapi.yaml
```diff
@@ -758,7 +758,6 @@ components:
             country: "ES"
       required:
       - oCloudId
-      - globalCloudId
```
mlguerrero12, line 761: The specification says that this field is mandatory, not nullable, shouldn't be make it mandatory in the flags as well?

internal/service/resources/cmd/serve.go
alegacy: Because the global cloud id is a required piece of data to send notification events to subscribers and so I don't believe we should allow a new subscription to be created until the end user has provided us with our global cloud id.   Our servers need to be up and running before the user will have provided this so we can't make it mandatory.  The best we can do is block some API endpoints if it is not provided -- with a 400 bad request with a suitable message. 

It appears in both the alarm event and the cloud available notification events.   I think I originally confused the cloud available event with the regular inventory change notifications so it isn't in fact needed on the resource server API endpoints, but it is necessary on the alarm server subscription create API.  ...but I'm sort of the view that if we block the alarm subscriptions we should block the inventory subscriptions as well to create consistent behaviour.

internal/service/resources/cmd/serve.go
```diff
@@ -1,25 +1,98 @@
...
+// setServerFlags creates the flag instances for the server
+func setServerFlags(cmd *cobra.Command) {
+	flags := cmd.Flags()
+	flags.StringVar(
+		&config.Listener.Address,
+		server.APIListenerAddressFlagName,
+		"127.0.0.1",
+		"API listener address.",
+	)
+	flags.StringVar(
```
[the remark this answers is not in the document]
alegacy: Correct, it is currently not needed because we terminate our TLS at the RBAC proxy, but given the recent announcement about the proxy and kubebuilder is quite likely that we'll have to move that back to the server directly.  Still a bit TBD.

internal/service/resources/api/openapi.yaml
alegacy: actually since the type is a uuid the best we could do here to allow it to be null or to return a zero value uuid object since we can't force it to be an "".

internal/service/resources/cmd/serve.go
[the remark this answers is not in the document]
alegacy: removed

internal/service/resources/cmd/serve.go
alegacy: fixed

internal/service/resources/api/openapi.yaml
alegacy: I've made it mandatory and left it as a zero uuid.

alegacy:
> Just minor comments, looks good. Please change the title and description of the PR

done

internal/service/resources/cmd/serve.go
mlguerrero12: Why do resource and alarms servers need to up and running before the user has provided the global cloud id? Even after they do, we will need to restart the servers, or not?

How is this (global cloud id) provided? (haven't read the code for that part yet) Shouldn't be mandatory by the operator?

PR #381, opened 2024-12-06
Implement alarm subscription endpoints

mlguerrero12:


internal/service/alarms/internal/alarms_server.go
```diff
@@ -41,8 +41,21 @@ func (a *AlarmsServer) DeleteSubscription(ctx context.Context, request api.Delet
...
 
 func (a *AlarmsServer) GetSubscription(ctx context.Context, request api.GetSubscriptionRequestObject) (api.GetSubscriptionResponseObject, error) {
-	// TODO implement me
-	return nil, fmt.Errorf("not implemented")
+	subsModel, err := a.AlarmsRepository.GetAlarmSubscriptionWithUuid(ctx, request.AlarmSubscriptionId)
+	if errors.Is(err, utils.ErrNotFound) {
+		// Nothing found
+		return api.GetSubscription404ApplicationProblemPlusJSONResponse(common.ProblemDetails{
+			AdditionalAttributes: &map[string]string{
+				"UUID": request.AlarmSubscriptionId.String(),
```
alegacy, line 49: nit:  since our yaml defines this variable as `subscriptionId` which should refer to it by name here rather than "UUID"

internal/service/alarms/internal/db/migrations/000005_create_alarm_subscription_info_table.up.sql
```diff
@@ -1,15 +1,22 @@
...
+    filter VARCHAR(20) NULL, -- Can be [new, change, clear, acknowledge], NULL means all
     callback TEXT NOT NULL, -- URL or endpoint for sending notifications
 
     -- Internal
     event_cursor BIGINT NOT NULL DEFAULT 0, -- Tracks the latest event for the subscriber. This used with alarm_event_record.alarm_sequence_number
     created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP, -- Record creation timestamp
     updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP -- Record last update timestamp
+
+    -- check does not fail with NULL
+    CONSTRAINT chk_filter CHECK (filter IN ('new', 'change', 'clear', 'acknowledge'))
```
alegacy, line 19: It would be best if this check was done in our rest handler so that we can provide a properly formatted error message rather than rely on the message returned by the DB.

alegacy (commented):
a few minor things, but otherwise looks good

internal/service/alarms/internal/db/migrations/000005_create_alarm_subscription_info_table.up.sql
mlguerrero12: rest handler will also have a check

internal/service/alarms/internal/alarms_server.go
```diff
@@ -18,31 +18,98 @@ import (
...
 }
 
+// CreateSubscription handles an API request to create an Alarm Subscription
 func (a *AlarmsServer) CreateSubscription(ctx context.Context, request api.CreateSubscriptionRequestObject) (api.CreateSubscriptionResponseObject, error) {
-	// TODO implement me
-	return nil, fmt.Errorf("not implemented")
+	// Block API if GlobalCloudID is not set
+	if a.GlobalCloudID == uuid.Nil {
+		return api.CreateSubscription409ApplicationProblemPlusJSONResponse(common.ProblemDetails{
+			Detail: "provisioning of Alarm Subscriptions is blocked until the global O-Cloud identifier is set",
```
alegacy, line 52: We should probably generalize this to:

```
provisioning of Alarm Subscriptions is blocked until the SMO attributes are configured.
```

The SMO attributes would encompass not only the global cloud id but also the OAuth attributes which would also be required (eventually) for us to get a token and send a notification.

mlguerrero12:
/hold

I'm having some issues with list and create

internal/service/alarms/internal/alarms_server.go
mlguerrero12: Indeed, I will change it and add a TODO for Reachability check.

 O-RAN.WG6.ORCH-USE-CASES-R003-v11.00 suggest to perform a reachability check before creating a subscription. I saw that inventory has CABundleName but I couldn't find the token. I decided to leave it for later.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -18,31 +18,98 @@ import (
...
+			Status: http.StatusConflict,
```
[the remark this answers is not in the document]
mlguerrero12: I don't know. I tend to associate 400 more like an error with the request itself. Another option is 422 Unprocessable Entity. To be honest, no of them really fit. The key here is to have a good message.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -18,31 +18,98 @@ import (
...
+		return api.GetSubscription404ApplicationProblemPlusJSONResponse(common.ProblemDetails{
+			AdditionalAttributes: &map[string]string{
+				"alarmSubscriptionId": request.AlarmSubscriptionId.String(),
+			},
+			Detail: "requested Alarm Subscription not found",
+			Status: http.StatusNotFound,
+		}), nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Alarm Subscription due to issues with DB conn: %w", err)
```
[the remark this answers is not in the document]
mlguerrero12: Ack

internal/service/alarms/internal/alarms_server.go
alegacy: ya, i still have to push those parameters down to the server.  Hopefully later this week or next.

internal/service/alarms/internal/alarms_server.go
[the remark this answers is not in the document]
alegacy: ok, 409

internal/service/alarms/internal/alarms_server.go
alegacy: i like the reachability check idea

PR #382, opened 2024-12-06
Refactor DB generics

mlguerrero12:
Simplify code by moving the creating of query
directly in the db repositories and use generics
only to execute the queries

### DIFF
--- a/internal/service/alarms/internal/db/repo/alarms_repository.go
+++ b/internal/service/alarms/internal/db/repo/alarms_repository.go
@@ -8,6 +8,10 @@ import (
 	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/stephenafamo/bob"
 	"github.com/stephenafamo/bob/dialect/psql"
+	"github.com/stephenafamo/bob/dialect/psql/dialect"
+	"github.com/stephenafamo/bob/dialect/psql/dm"
+	"github.com/stephenafamo/bob/dialect/psql/im"
+	"github.com/stephenafamo/bob/dialect/psql/sm"
 
 	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/db/models"
 	"github.com/openshift-kni/oran-o2ims/internal/service/common/utils"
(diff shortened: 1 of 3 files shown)


mlguerrero12:
@alegacy, I believe this will simplify a lot the code. By moving the build of the query in the db repositories, we can easily construct more complex queries with many records. It looks cleaner. Please have a look

alegacy:
> @alegacy, I believe this will simplify a lot the code. By moving the build of the query in the db repositories, we can easily construct more complex queries with many records. It looks cleaner. Please have a look

Let's discuss next week.  I must be missing something because this seems to have moved what was a relatively manageable set of generic functions back out to the repository and duplicates the logical in dozens of places.

If any particular use cases needed a more complex query there was nothing stopping us from doing that on a case-by-case basis.  Having the typical Find/Create/Delete cases handled generically seemed like a better approach short of using a proper ORM to do the heavy lifting.

mlguerrero12:
> > @alegacy, I believe this will simplify a lot the code. By moving the build of the query in the db repositories, we can easily construct more complex queries with many records. It looks cleaner. Please have a look
> 
> Let's discuss next week. I must be missing something because this seems to have moved what was a relatively manageable set of generic functions back out to the repository and duplicates the logical in dozens of places.
> 
> If any particular use cases needed a more complex query there was nothing stopping us from doing that on a case-by-case basis. Having the typical Find/Create/Delete cases handled generically seemed like a better approach short of using a proper ORM to do the heavy lifting.

As we discussed internally, my main concerned was that it seemed we were maintaining a library here. There was no clear convention of what was a basic generic operation. I didn't want us in the future to continue adding arguments to fulfill all needs. For instance, if someone needed more flexibility with the upsert operations (which might be very complex itself) or add support for bulk operations with Insert or Update.

I do understand your point about repeating code. Although, I believe it was ok in this case.

So, as discussed, I left the basic crud operations and emphasized that Insert and Update are for a single record. I cleaned up a bit. I removed the generic functions for upsert. The queries for upsert are now built in the repo file. I also introduced 2 generic functions to execute queries which are used by the existing crud operations and can be used outside to run more complex queries.

Please have a look again, @alegacy

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -45,35 +47,58 @@ func (ar *AlarmsRepository) DeleteAlarmDefinitionsNotIn(ctx context.Context, ids
...
 func (ar *AlarmsRepository) UpsertAlarmDictionary(ctx context.Context, record models.AlarmDictionary) ([]models.AlarmDictionary, error) {
-	tags := utils.GetDBTagsFromStructFields(record, "AlarmDictionaryVersion", "EntityType", "Vendor", "ResourceTypeID")
+	dbModel := models.AlarmDictionary{}
 
-	// Important to keep the order. The order of tags.Columns().. is not deterministic
-	values := []bob.Expression{psql.Arg(record.AlarmDictionaryVersion, record.EntityType, record.Vendor, record.ResourceTypeID)}
-	records, err := utils.UpsertOnConflict[models.AlarmDictionary](ctx, ar.Db, []string{tags["AlarmDictionaryVersion"], tags["EntityType"], tags["Vendor"], tags["ResourceTypeID"]}, values)
+	tags := utils.GetDBTagsFromStructFields(dbModel, "AlarmDictionaryVersion", "EntityType", "Vendor", "ResourceTypeID")
+
+	columns := []string{tags["AlarmDictionaryVersion"], tags["EntityType"], tags["Vendor"], tags["ResourceTypeID"]}
```
alegacy, line 54: It's really unfortunately that they sometimes accept `any` and other times they accept `string` because here we could have used `tags.Columns()`

alegacy (approved):
/lgtm
/approve

PR #384, opened 2024-12-07
CNF-15570: Add inventory collector/persistence (part 2)

alegacy:
This is a continuation of PR https://github.com/openshift-kni/oran-o2ims/pull/379.  At @mlguerrero12 's request I have split it so that the first 3 commits are in the other PR, and the last 4 commits are in this PR.  Since the first PR isn't yet merged those commits appear here as well so just ignore those and review the last 4 only. 

These are relevant here:
+ [CNF-15570](https://issues.redhat.com//browse/CNF-15570): Add notification support
+ [CNF-15570](https://issues.redhat.com//browse/CNF-15570): Add support for persisting deployment managers
+ [CNF-15570](https://issues.redhat.com//browse/CNF-15570): Add inventory data collector
+ [CNF-15570](https://issues.redhat.com//browse/CNF-15570): Add changes needed for upcoming collector


internal/service/resources/api/openapi.yaml
```diff
@@ -1067,3 +1066,42 @@ components:
...
+      type: object
+      properties:
+        notificationId:
+          type: string
+          format: uuid
+          description: |
+            A unique identifier to represent this notification event
+        consumerSubscriptionId:
+          type: string
+          description: |
```
mlguerrero12, line 1082: format is UUID

internal/service/resources/api/openapi.yaml
```diff
@@ -1067,3 +1066,42 @@ components:
...
+            The value provided by the consumer in the subscription
+        notificationEventType:
+          type: integer
+          description: |
```
mlguerrero12, line 1086: why don´t you add an enum here so that the check is performed by the middleware?

internal/service/resources/serve.go
```diff
@@ -151,6 +167,8 @@ func Serve(config *api.ResourceServerConfig) error {
 	select {
 	case err := <-serverErrors:
 		return fmt.Errorf("error starting server: %w", err)
+	case err := <-collectorErrors:
+		return fmt.Errorf("error starting collector: %w", err)
```
mlguerrero12, line 186: shutdown the server

internal/service/resources/serve.go
```diff
@@ -151,6 +167,8 @@ func Serve(config *api.ResourceServerConfig) error {
 	select {
 	case err := <-serverErrors:
 		return fmt.Errorf("error starting server: %w", err)
```
mlguerrero12, line 184: likewise, if the server is stopped, we need to stop the collector

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+	dataSource.IncrGenerationID()
+
+	// Get the list of resource pools for this data source
+	pools, err := c.collectResourcePools(ctx, dataSource)
+	if err != nil {
+		return fmt.Errorf("failed to collect resource pools: %w", err)
+	}
+
+	// Get the list of resources for this data source
+	_, err = c.collectResources(ctx, dataSource, pools)
```
mlguerrero12, line 144: dataSource.GetResources doesn't need pools, so we could delete it. If we do that, is it not safe to run collectResources even if collectResourcesPools fails. We could attempt to sync Resources

internal/service/resources/api/openapi.yaml
alegacy: I could add an enum here, but it is not needed for validation as this object is outbound only.  The user never sends us one of these.   I'll change it.

internal/service/resources/collector/collector.go
alegacy: The reason I set this up this way is that it will have to be this way in the future for the h/w manager because as of now there is no way to get just a full list of resources, they need to be scoped by resource pool... so for the ACM one it just gets ignored, but other data source types may need this information.

internal/service/resources/serve.go
alegacy: thanks

internal/service/resources/serve.go
alegacy: reply

internal/service/resources/collector/utils.go
```diff
@@ -0,0 +1,225 @@
...
+// model representation of the object has changed then a data change event is stored.  Persisting
+// of the object and its change event are captured under the same transaction to ensure we never
+// lose any change events.
+func persistObjectWithChangeEvent[T db.Model](ctx context.Context, db *pgxpool.Pool, record T,
+	uuid uuid.UUID, parentUUID *uuid.UUID,
+	converter GenericModelConverter) (*models.DataChangeEvent, error) {
+	var dataChangeEvent *models.DataChangeEvent
+	txCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	tx, err := db.Begin(txCtx)
```
mlguerrero12, line 123: this seems not to be ok. You are not doing anything with the transaction structure (tx).

the defer tx.Rollback() is also missing. For queries to be part of the transaction, they should be called from methods of the tx structure. https://pkg.go.dev/github.com/jackc/pgx#Tx There are Exec and QueryRow

alegacy: I'm not sure that's true.  The `BeginTx` just stores the current `conn` into the `Tx` structure and those delegated methods on `Tx` just call the underlying connection.  I'll take a closer look tomorrow as I missed this comment when I fixed the others from today.  

As for the tx.Rollback() it is not necessary if you do a `defer cancel()` on the context passed to `Begin()`... it does nothing if the transaction was already commited, and rolls it back otherwise.

alegacy: Also, when I tested this it seemed to be working as expected... but i'll double-check.

mlguerrero12: That is not true. We are using a pool here. `Begin` calls `BeginTx` which acquires a connection and then stores that connection into Tx. https://github.com/jackc/pgx/blob/v5.5.5/pgxpool/pool.go#L653

The transaction is relevant for that connection. From our utils/repository, we call `Query` which acquires another connection.

And Rollback is needed, the context only applies for the Begin method. See comment

```
// Begin acquires a connection from the Pool and starts a transaction. Unlike database/sql, the context only affects the begin command. i.e. there is no
// auto-rollback on context cancellation. Begin initiates a transaction block without explicitly setting a transaction mode for the block (see BeginTx with TxOptions if transaction mode is required).
// *pgxpool.Tx is returned, which implements the pgx.Tx interface.
// Commit or Rollback must be called on the returned transaction to finalize the transaction block.
```

alegacy: thanks, I'll rework this a bit

internal/service/resources/serve.go
alegacy: Fixed

internal/service/resources/api/openapi.yaml
alegacy: Fixed

internal/service/resources/collector/utils.go
alegacy: Fixed

internal/service/common/utils/repository.go
```diff
@@ -26,10 +26,26 @@ var ErrNotFound = errors.New("record not found")
 // Following functions are meant to fulfill basic CRUD operations on the database. More complex queries or bulk operations
 // for Insert or Update should be built in the repository files of the specific service and called one of the Execute helper functions.
 
+// QueryFunc is an abstraction to allow passing either a pool or a transaction query function to any of the utilities.
```
mlguerrero12, line 29: this could be much cleaner if you create an interface

```
type DBQuery interface {
   QueryFunc func(ctx context.Context, sql string, args ...any) (pgx.Rows, error)
}

```
and then replace db *pgxpool.Pool by DBQuery

No need to create additional functions and have internal ones (first letter lowercase functions)

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Loop over the set of resource pools and insert (or update) as needed
+	for _, pool := range pools {
+		dataChangeEvent, err := persistObjectWithChangeEvent(
+			ctx, c.repository.Db, pool, pool.ResourcePoolID, nil, func(object interface{}) any {
+				record, _ := object.(models.ResourcePool)
+				return models.ResourcePoolToModel(&record)
+			})
+		if err != nil {
+			return nil, fmt.Errorf("failed to persist resource pool: %w", err)
```
mlguerrero12, line 234: I think we should try persist as many pools as we can. Instead of returning an error, we could log (similar to what you do in GetResourcePools) and continue. We can remove that pool from pools (again similar to GetResourcePools).

internal/service/resources/collector/utils.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	return result, nil
+}
+
+// persistObjectWithChangeEvent persists an object to its database table and if the external API
+// model representation of the object has changed then a data change event is stored.  Persisting
+// of the object and its change event are captured under the same transaction to ensure we never
+// lose any change events.
+func persistObjectWithChangeEvent[T db.Model](ctx context.Context, db *pgxpool.Pool, record T,
```
mlguerrero12, line 127: What do you think about creating methods in dbModel to return the primarykey and parentId values? It could simplify the signature of these methods. T.GetprimaryKey()

internal/service/resources/collector/utils.go
```diff
@@ -0,0 +1,225 @@
...
+
+// GenericModelConverter converts a db model object into an API object and returns it as an `any`
+// so that it can be used generically since the API model doesn't implement a base interface.
+type GenericModelConverter func(object interface{}) any
+
+// persistObject persists an object to its database table.  If the object does not already have a
+// persisted representation then it is created; otherwise any modified fields are updated in the
+// database tuple.  The function returns both the before and after versions of the object.
+func persistObject[T db.Model](ctx context.Context, db *pgxpool.Pool,
+	object T, uuid uuid.UUID) (*T, *T, error) {
```
mlguerrero12, line 32: nit: I'm so used to record now that object confused me for a moment. Perhaps for consistency,  consider changing object by record

internal/service/common/utils/utils.go
```diff
@@ -115,3 +125,52 @@ func GetDBTagsFromStructFields[T db.Model](s T, fields ...string) DBTag {
...
+		case field.Type.Kind() != reflect.Ptr:
+			// Non-pointer value compare them directly
+			if !reflect.DeepEqual(aFieldValue.Interface(), bFieldValue.Interface()) {
+				tags[fieldName] = tagValue
+			}
+		case aFieldValue.IsNil() != bFieldValue.IsNil():
+			// Pointer nil-ness are different so include this field
+			tags[fieldName] = tagValue
+		case !aFieldValue.IsNil():
+			// Pointer nil-ness are the same, so compare their values
```
mlguerrero12, line 168: Something seems wrong with this comment - you're saying something like "compare values of non nil pointers", right?

internal/service/resources/collector/utils.go
```diff
@@ -0,0 +1,225 @@
...
+	var beforeModel any = nil
+	if before != nil {
+		value := converter(*before)
+		beforeModel = value
+	}
+
+	if beforeModel == nil || !reflect.DeepEqual(beforeModel, afterModel) {
+		// Capture a change event if the data actually changed
+		dataChangeEvent, err = persistDataChangeEvent(
+			ctx, db, record.TableName(), uuid, parentUUID, beforeModel, afterModel)
```
mlguerrero12: So, this is what makes difficult using an upsert and/or run a bulk operation because as I commented previously, updating one object at a time is the lowest approach in pgx.

The way I see it is, we only compare the fields that are exposed in the API model. I don't see you have defaults in the table for those. So, record already have those values. If find a way to determine when an upsert inserts or updates, then we could replace everything inside `persistObject` by an upsert. 

I found this https://stackoverflow.com/questions/39058213/differentiate-inserted-and-updated-rows-in-upsert-using-system-columns. But of course, this needs further investigation.

This is a good start. We could improve it later

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Loop over the set of resources and create the associated resource types.
+	seen := make(map[uuid.UUID]bool)
+	for _, resource := range resources {
+		resourceType, err := dataSource.MakeResourceType(&resource)
+		if err != nil {
+			return nil, fmt.Errorf("failed to make resource type from '%v': %w", resource, err)
+		}
+
+		if _, ok := seen[resourceType.ResourceTypeID]; ok {
```
mlguerrero12, line 179: if seen[resourceType.ResourceTypeID] is enough. Not need to use ok

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Loop over the set of resources and insert (or update) as needed
+	for _, resource := range resources {
+		dataChangeEvent, err := persistObjectWithChangeEvent(
+			ctx, c.repository.Db, resource, resource.ResourceID, &resource.ResourcePoolID, func(object interface{}) any {
+				record, _ := object.(models.Resource)
+				return models.ResourceToModel(&record, nil)
+			})
+		if err != nil {
```
mlguerrero12, line 206: Same comment as in the resource pool for errors.

Let's try to keep the DB as consistent as possible.

Say one of these fails. We will have commited resource pools, resource types and some or at the worst none resources (if the first fails). The whole data willl be inconsistent until the next sync.

Am I right?

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Get the list of resources for this data source
+	_, err = c.collectResources(ctx, dataSource, pools)
+	if err != nil {
+		return fmt.Errorf("failed to collect resources: %w", err)
+	}
+
+	// TODO: purge stale record
+
+	// TODO: persist data source info
```
mlguerrero12, line 157: Another reason not to fail with errors in the for loops. We don´t want the generation id to be greater than the generation id of the data source, in the DB I mean

internal/service/common/utils/repository.go
alegacy: great idea.

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Run the initial data collection
+	c.execute(ctx)
+
+	for {
+		select {
+		// TODO: Add hook for new data sources from watch events
+		// TODO: Add hook for kick to run collection based on watch events on individual data sources
+		case <-time.After(time.Minute * 1):
```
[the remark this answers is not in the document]
alegacy: ya, i kept it short for development.  I'll change it to 10 for now and then later make it configurable.

internal/service/resources/api/openapi.yaml
```diff
@@ -667,6 +667,7 @@ components:
             List of globalLocationIDs that were assigned to the OCloud Site(s) which this Deployment Manager supports.
           items:
             type: string
+            format: uuid
```
mlguerrero12, line 670: the spec doesn't say this is a uuid - it explicitly says string, not identifier

internal/service/resources/collector/utils.go
alegacy: I agree, this should be an area to revisit later.   As discussed the other day, having a proper CDC mechanism would solve these issues since we could blindly do an upsert and handle the change event separately to compare the fields.  Bringing it a full CDC mechanism is too much complexity for what we actually need.  Implemented them with a simpler DB trigger would be easier but would pose portability issues, as well as issues scaling the number of servers so we'll have to consider those issues.   For now I think this will serve us well enough.

internal/service/resources/collector/collector.go
```diff
@@ -227,3 +234,30 @@ func (c *Collector) collectResourcePools(ctx context.Context, dataSource DataSou
...
+	dms, err := dataSource.GetDeploymentManagers(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get deployment managers: %w", err)
+	}
+
+	// Loop over the set of deployment managers and insert (or update) as needed
+	for _, dm := range dms {
+		dataChangeEvent, err := persistObjectWithChangeEvent(
+			ctx, c.repository.Db, dm, dm.DeploymentManagerID, nil, func(object interface{}) any {
+				record, _ := object.(models.ResourcePool)
```
mlguerrero12, line 257: ResourcePool?

internal/service/resources/collector/collector.go
alegacy: I'll give this some thought based on how things behave in testing at larger scale.  Like I said in the previous comment there shouldn't be any reason for one record to pass and another to fail so that leaves db connection issues which should be quite rare.  There is too much data to try and do everything under one single transaction so it would be difficult to ensure the data is always consistent.

internal/service/resources/collector/collector.go
alegacy: Having a generation id greater than the data source doesn't have to be a problem.  If it is less than the data source than it will be purged.  

I'll think about this a bit and revisit in another PR when I implement the purging.

internal/service/resources/collector/collector_acm.go
```diff
@@ -414,3 +449,110 @@ func (d *ACMDataSource) convertClusterToResourcePool(from data.Object) (to model
...
+		"profileName":          "k8s",
+		"admin_client_cert":    adminCert,
+		"admin_client_key":     adminKey,
+		"admin_user":           adminUser,
+		"cluster_ca_cert":      caData,
+		"cluster_api_endpoint": url,
+	}, nil
+}
+
+// convertManagedClusterToDeploymentManager converts a ManagedCluster to a DeploymentManager object
```
mlguerrero12, line 516: Please remind me the relationships again

DeploymentManager from ManagedCluster
ResourcePool also from ManagedCluster
Resource from Node which for SNO is only one
ResourceType for each Resource, meaning a Node

is that right?

mlguerrero12: My point is, it seems you already have info enough to build the DeploymentManager, no?

internal/service/resources/collector/collector_acm.go
```diff
@@ -414,3 +449,110 @@ func (d *ACMDataSource) convertClusterToResourcePool(from data.Object) (to model
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal cluster config: %w", err)
+	}
+
+	return config, nil
+}
+
+// getDeploymentManagerExtensions builds the extensions required for the deployment manager
+func (d *ACMDataSource) getDeploymentManagerExtensions(ctx context.Context, clusterName string) (map[string]string, error) {
+	// Fetch the kubeconfig for this cluster and provide the info as extensions to the returned object. This
```
mlguerrero12, line 485: Good idea to add it to the extensions from now

internal/service/resources/api/openapi.yaml
alegacy: Yes, I know but it sort of sounds like this is supposed to be a collection of globalLocationId values from the pools which isn't well defined in the spec and sounds like it is a uuid...  the conformance tests seem to treat these as non-uuid strings so i'll line them both up the same way for now.

internal/service/resources/collector/collector_acm.go
alegacy: In this PR, yes that's correct.  That will get us parity with how the API currently works in the older implementation.  We need to change the handling of cluster so that it also has a resourceType so that we can raise alarms against it.   Need to get clarification from the standards folks on how they want that done.

internal/service/resources/collector/collector_acm.go
alegacy: Not my idea... that's how it was currently being done because of how it is defined in the conformance tests.  IMHO, the fact that the conformance tests are expecting specific things in extensions is wrong, but...

internal/service/resources/collector/collector.go
alegacy: fixed

internal/service/resources/api/openapi.yaml
alegacy: Fixed

internal/service/resources/collector/collector.go
alegacy: Fixed

internal/service/resources/collector/utils.go
alegacy: Maybe in a future PR.  I was hoping to avoid adding too much to the dbModel.  For now this two extra parameters are used in many places.

internal/service/resources/collector/collector.go
alegacy: I'll revisit when we add more functionality.

internal/service/common/utils/repository.go
alegacy: fixed

internal/service/resources/collector/collector.go
mlguerrero12: I know, but it is not for us. Support people will have to interact with this. At some point, something could happen. A support person will look at the DB and get very confused (based on training and experience) at seeing rows with generation id greater than the one for the data source.

But yeah, let's discuss this in another pr

internal/service/resources/collector/collector_acm.go
mlguerrero12: yeah, weird

internal/service/resources/collector/collector.go
```diff
@@ -0,0 +1,265 @@
...
+		} else {
+			slog.Debug("collected data from data source", "source", d)
+		}
+	}
+	slog.Debug("collector loop complete", "sources", len(c.dataSources))
+}
+
+// executeOneDataSource runs a single iteration of the main loop for a specific data source instance.
+func (c *Collector) executeOneDataSource(ctx context.Context, dataSource DataSource) (err error) {
+	// Increment the generation id
```
mlguerrero12, line 129: I think we are missing debug logs here at the start of  executeOneDataSource with the new generation ID and inside of collectResourcePools, collectResources and collectDeploymentManagers

mlguerrero12 (commented):
/approve

Just a minor comment

internal/service/resources/collector/collector.go
alegacy: added

PR #390, opened 2024-12-09
CNF-15750: Add field and record filtering to new API servers

alegacy:
This ports the existing include/exclude field filter as well as the record filter from the existing API implementation to the new API implementation for both the alarm and resource servers.  See the common openapi.yaml file for more information on the functioning of the query parameters for 'fields', 'exclude_fields', and 'filter' parameters.

The 'filter' functionality should eventually be integrated directly into the database queries to limit the number of results directly at the source rather than post-processing the result which is far more expensive. The current approach is a temporary solution to achieve parity with the existing API implementation with the least amount of cost.

### DIFF
--- a/internal/service/alarms/serve.go
+++ b/internal/service/alarms/serve.go
@@ -127,6 +127,18 @@ func Serve(config *AlarmsServerConfig) error {
 
 	r := http.NewServeMux()
 
+	// Create a new logger to be passed to things that need a logger
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		AddSource: true,
+		Level:     slog.LevelDebug, // TODO: set with server args
+	}))
+
+	// Create a response filter filterAdapter that can support the 'filter' and '*fields' query parameters
+	filterAdapter, err := common.NewFilterAdapter(logger)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


mlguerrero12:
I'm not sure how to approach this PR. If there is a need to have feature parity as soon as possible and if this works, then, I'm ok with it. We could work on integrating the DB later and also to use nethttp-middleware to build the middleware instead of Gorilla Mux. In other words, rewrite it so that it looks more to how we are handling stuff in the new implementation.

/lgtm
/assign @pixelsoccupied

pixelsoccupied:
/approve

yep let's get this merged to unblock tests! 

But yeah generally i was under the impression that each API will handle there own request which will contain their specific filter/fields/exclude_fields etc and the we will then convert that into specific DB query. 

e.g For alerts events we will allow users to only query these [attributes](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql#L12-L21) and throw `bad request` if want to query based say `fingerprint` (basically not allowed to query internal stuff) 

Doing this this after (in the middleware) as a catch all may not be the right approach and each API should handle it in its own way.

PR #412, opened 2024-12-16
NO-ISSUE: Update go modules

donpenney:
Update k8s.io/client-go, which has a `replace` directive, to v0.31.4 to align with other k8s.io modules.

Updating go.mongodb.org/mongo-driver to latest (v1.17.1) from v1.10.0, as that version has been retracted by the module author:
```
go: warning: go.mongodb.org/mongo-driver@v1.10.0: retracted by module author: Contains a possible data corruption bug in RewrapManyDataKey when using libmongocrypt versions less than 1.5.2.
```


donpenney:
/override ci/prow/markdownlint ci/prow/ci-bundle-operator-bundle

PR #414, opened 2024-12-16
Implement List and Patch for Alarm Event Record

mlguerrero12:
This implements the missing API endpoints for Alarm Event Record


mlguerrero12:
@pixelsoccupied, @alegacy, PTAL

internal/service/alarms/internal/alarms_server.go
```diff
@@ -191,8 +195,58 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
...
+	if !alarmAcknowledged {
+		// Bad request - Acknowledged status is expected to be true
+		return api.AckAlarm400ApplicationProblemPlusJSONResponse(common.ProblemDetails{
+			AdditionalAttributes: &map[string]string{
+				"alarmEventRecordId": request.AlarmEventRecordId.String(),
+			},
+			Detail: "Alarm Event Record Acknowledged status is expected to be true",
+			Status: http.StatusBadRequest,
+		}), nil
+	}
```
alegacy, line 226: We may want to consider allowing un-acknowledging an alarm so that a client can 'undo' a mistake of acknowledging the wrong alarm.  The interface document doesn't describe how to handle this so a more flexible approach may be better.

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -39,21 +39,33 @@ CREATE TABLE IF NOT EXISTS alarm_event_record (
 -- Set ownership of the alarm_sequence_seq sequence to alarm_event_record.alarm_sequence_number
 ALTER SEQUENCE alarm_sequence_seq OWNED BY alarm_event_record.alarm_sequence_number;
 
--- Function to update the alarm_sequence_number on specific status or time changes
-CREATE OR REPLACE FUNCTION update_alarm_event_sequence()
+-- Function to execute before updating alarm_event_record
+CREATE OR REPLACE FUNCTION update_alarm_event_record()
     RETURNS TRIGGER AS $$
 BEGIN
+    -- Update alarm_acknowledged_time if alarm_acknowledged is set to TRUE
```
alegacy, line 46: Is there a specific reason for doing this work in a trigger?   This can easily be handled while setting the acknowledged attribute by the server.   Triggers are a pain to maintain and not portable so I'd prefer to limit our use of them to a minimum.

alegacy:
Could you update the test-client RBAC and test that the new API endpoints are accessible with a token from this client. Here: config/testing/client-service-account-rbac.yaml

```bash
oc apply -f config/testing/client-service-account-rbac.yaml
export MY_TOKEN=$(oc create token -n oran-o2ims test-client --duration=24h)
```

internal/service/alarms/internal/alarms_server.go
mlguerrero12: That was my initial thought as well, but then I realized that the specifications refer to this operation in only one direction. Either acknowledging or clearing the alarm but no the reverse operation. The use case in O-RAN.WG6.ORCH-USE-CASES-R003-v11.00 too. In fact, for clearing the alarm, perceivedSeverity is restricted to 5, so once you set it, there is no way back. I think it is ok to not allow the reverse operation for acknowledging. Functionality is not affected, it is only a way to mark that an alarm has been seen. Also, alarms will be eventually deleted. So, this is not a persistent object.

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
mlguerrero12: Yeah, let's keep it in code. Thanks!

internal/service/alarms/internal/alarms_server.go
```diff
@@ -190,9 +195,110 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
...
+			Status: http.StatusBadRequest,
+		}), nil
+
+	}
+
+	// Patch perceivedSeverity
+	if request.Body.PerceivedSeverity != nil {
+		perceivedSeverity := *request.Body.PerceivedSeverity
+		// Only the value "5" for "CLEARED" is permitted in a request message content
+		if perceivedSeverity != 5 {
```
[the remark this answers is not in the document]
mlguerrero12: fixed

internal/service/alarms/internal/alarms_server.go
alegacy: ok

mlguerrero12:
> Could you update the test-client RBAC and test that the new API endpoints are accessible with a token from this client. Here: config/testing/client-service-account-rbac.yaml
> 
> ```shell
> oc apply -f config/testing/client-service-account-rbac.yaml
> export MY_TOKEN=$(oc create token -n oran-o2ims test-client --duration=24h)
> ```

Done

internal/service/alarms/internal/alarms_server.go
```diff
@@ -190,9 +195,110 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
 	return api.GetAlarm200JSONResponse(models.ConvertAlarmEventRecordModelToApi(*record)), nil
 }
 
-func (a *AlarmsServer) AckAlarm(ctx context.Context, request api.AckAlarmRequestObject) (api.AckAlarmResponseObject, error) {
-	// TODO implement me
-	return nil, fmt.Errorf("not implemented")
+// PatchAlarm handles an API request to patch an Alarm Event Record
+func (a *AlarmsServer) PatchAlarm(ctx context.Context, request api.PatchAlarmRequestObject) (api.PatchAlarmResponseObject, error) {
```
pixelsoccupied, line 199: As discussed previously and noted in the design doc...we are only supporting Ack. i,e user may only supply AlarmEventRecordModifications with a alarmAcknowledged bool. Let's revert the spec and follow the design as agreed upon.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -159,15 +160,19 @@ func (a *AlarmsServer) GetSubscription(ctx context.Context, request api.GetSubsc
 	return api.GetSubscription200JSONResponse(models.ConvertSubscriptionModelToApi(*record)), nil
 }
 
-func (a *AlarmsServer) GetAlarms(ctx context.Context, request api.GetAlarmsRequestObject) (api.GetAlarmsResponseObject, error) {
-	// TODO implement me
+// GetAlarms handles an API request to fetch Alarm Event Records
+func (a *AlarmsServer) GetAlarms(ctx context.Context, _ api.GetAlarmsRequestObject) (api.GetAlarmsResponseObject, error) {
+	records, err := a.AlarmsRepository.GetAlarmEventRecords(ctx)
```
pixelsoccupied, line 165: Though filter support is currently not there, please log the input and log so that user is aware of this and that we are simply returning all the data .

internal/service/alarms/internal/alarms_server.go
mlguerrero12: I don´t mind reverting this part. However, I thought this is a good opportunity to put this into discussion again. So far, we've been making an effort to follow what the specification defines. Not only what is defined in O-RAN.WG6.O2IMS-INTERFACE-R004-v07.00 but also properties of fields defined in O-RAN.WG6.O2IMS-INTERFACE-R004-v07.00 (format, nullable, etc).

The current version (supporting only ACK) does not follow at all what is defined in table 3.3.6.2.4-1 (O-RAN.WG6.O2IMS-INTERFACE-R004-v07.00). Not only the lack of the `perceivedSeverity` field but also that these are nullable fields. @alegacy has been running conformance tests. It seems there are none for this, but eventually they will. I don't see why we cannot follow the specification.

Another thing, the initial design is not set in stone (not even merged). Luckily, we are still in the initial phases. We could change stuff as we progress with the implementation.

As said, I don't mind changing it back, but I think it is wrong. Let's see what @alegacy and/or @browsell thinks.

internal/service/alarms/internal/alarms_server.go
mlguerrero12: This is how we've handling list for alarm and resource endpoints. If we are gonna modify something, let's do it for all at once in a different pr. Here, I'm following the same format.

internal/service/alarms/internal/alarms_server.go
pixelsoccupied: The reason why we landed on only ack (for now at least) is that we cant modify the caas alerts data. The PerceivedSeverity is essentially PromRule label `severity` which is hardcoded.

alegacy: @pixelsoccupied I must have missed the conversation about why we wouldn't support the clear use case.  AFAIK, our goal is to try to support as much of the spec as technically possible so unless there is a strong reason why we shouldn't I think we should proceed with both the ack and clear.   Of course, we'll have to be mindful of the manually cleared state when syncing alarms in from downstream sources so that we don't un-clear it again when we sync them in.

internal/service/alarms/internal/alarms_server.go
alegacy: Filtering is coming in PR #390 to get us back to parity with how the API works in the current/old implementation.  The filtering is not optimal as it looks at the data after it has already been fetched from the db, but we can rework this as a later step... with a goal of getting more functionality in place before we circle back and rework too many things.

Logging input can be handled when we rework the filtering at a later time... and/or we can throw an additional debug log in PR #390

internal/service/alarms/internal/alarms_server.go
alegacy: @browsell @pixelsoccupied  You are correct that we only have read-only access to the source data, but given that we are persisting an intermediate copy there are some things that we can do to support the API requirements. ... we just have to be careful not to overwrite any user modifications if we re-sync the data.  Of course, if there is a technical reason why we can't do that then this is a problem, but is that the case?

mlguerrero12: @alegacy @pixelsoccupied @browsell 

This is from the use case for reference to support my previous message about Manual and Automatic type of alarms

```
This use case is used for a management entity to acknowledge or request for alarm(s) to be cleared. It is also used for
the IMS to clear alarm(s) autonomously. It is expected that some resources could have some alarms that can be
manually cleared while other alarms would be automatically cleared. 
```

mlguerrero12: Discussed internally. We will follow proposed approach for clearing alarms. Distinction between Manual and Automatic ones.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -190,9 +196,149 @@ func (a *AlarmsServer) GetAlarm(ctx context.Context, request api.GetAlarmRequest
...
+
+		if alarmDefinition.ClearingType != string(apiresources.MANUAL) {
+			return api.PatchAlarm409ApplicationProblemPlusJSONResponse(common.ProblemDetails{
+				AdditionalAttributes: &map[string]string{
+					"alarmEventRecordId": request.AlarmEventRecordId.String(),
+				},
+				Detail: "cannot clear an alarm with clearing type other than MANUAL",
+				Status: http.StatusConflict,
+			}), nil
+		}
```
alegacy, line 288: We should check the clearing type as a first step so that we don't proceed further if the type is not "manual".  I would also argue that since we don't currently have any alarms of type "manual" that we probably don't need the other code at all... just this code to always reject because the clearing type is "manual".

mlguerrero12: It seems more natural to me to check the clearing status of the alarm first and then check if it can be cleared. It also saves a DB round trip. But if you have a strong request for the other way. I won't mind changing it.

Regarding the other part, I think for correctness, we should handle clearing of Manual. We don't know what we will have later. Besides, it is just a few lines of code.

alegacy: I don't feel strongly about it, but it seems pointless to validate a request that is not allowed....  also, it gives a client a false impression about what needs to change for their request to succeed.

mlguerrero12: done

PR #415, opened 2024-12-16
CNF-15570: Set default 404 handler

alegacy:
The default 404 handler on the router generates a text response, but the ORAN spec requires a JSON response.  This adapts the output to be a JSON response with the appropriate content type header.


alegacy:
@pixelsoccupied  had to push to fix the ci-job.

PR #419, opened 2024-12-18
CNF-15570: Fix issues with notifier

alegacy:
This addresses two issues with the notifier.

+ On a restart it blocks on running a subscription worker rather than dispatching it to a go routing and continuing.

+ Any events received when idle results in 2 notifications being sent to the subscriber.


Missxiaoguo:
CNF-15570 doesn't seem to be the correct JIRA 🤔

PR #420, opened 2024-12-19
CNF-16177: Implement Alarm Service Configuration endpoints

mlguerrero12:
This implements the alarm service configuration endpoints. The specification defines GET, PUT (only update, not create) and PATCH operations. Therefore, we create a default service configuration at startup. The spec also mentions that the retention period has a minimum value which is not explicitly defined. We assume 10 and 1 as default and minimum retention periods respectively. The values represent days.

### DIFF
--- a/internal/service/alarms/api/openapi.yaml
+++ b/internal/service/alarms/api/openapi.yaml
@@ -18,7 +18,8 @@ servers:
 tags:
   - name: alarms
     description: Alarm management
-  - name: subscriptions
+  - name: service configuration
+  - name: Alarm Service Configuration
     description: Alarm subscription management
   - name: probableCauses
     description: Probable cause information
@@ -204,6 +205,106 @@ paths:
                 $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
         '409':
           description: Conflict.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+        '412':
+          description: Precondition failed.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+        '500':
+          description: Internal server error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+
+  /o2ims-infrastructureMonitoring/v1/alarmServiceConfiguration:
+    get:
+      operationId: GetServiceConfiguration
+      summary: Retrieve the alarm service configuration
+      tags:
+        - service configuration
+      responses:
+        '200':
+          description: Successful response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AlarmServiceConfiguration'
+        '400':
+          description: Bad request
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+        '500':
+          description: Internal server error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+
+    put:
+      operationId: UpdateAlarmServiceConfiguration
+      summary: Modify all fields of the Alarm Service Configuration.
+      tags:
+        - service configuration
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AlarmServiceConfiguration'
+      responses:
+        '200':
+          description: Successful response.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AlarmServiceConfiguration'
+        '400':
+          description: Bad request.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+        '412':
+          description: Precondition failed.
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+        '500':
+          description: Internal server error
+          content:
+            application/problem+json:
+              schema:
+                $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
+
+    patch:
+      operationId: PatchAlarmServiceConfiguration
+      summary: Modify individual fields of the Alarm Service Configuration.
+      tags:
+        - service configuration
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AlarmServiceConfiguration'
+      responses:
+        '200':
+          description: Successful response.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AlarmServiceConfiguration'
+        '400':
+          description: Bad request.
           content:
             application/problem+json:
               schema:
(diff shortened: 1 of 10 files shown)


mlguerrero12:
/hold 
Testing and missing put permission for test-client

internal/service/alarms/api/openapi.yaml
```diff
@@ -576,6 +677,23 @@ components:
...
             $ref: '#/components/schemas/PerceivedSeverity'
 
+    AlarmServiceConfiguration:
+      type: object
+      properties:
+        retentionPeriod:
+          type: integer
+          description: | 
+            Number of days for alarm history to be retained.
+            This value has a minimum limit such that it can’t be set to a value lower than the limit.
```
alegacy, line 687: We should state what the minimum is.

internal/service/alarms/internal/alarms_server.go
```diff
@@ -341,6 +346,107 @@ func (a *AlarmsServer) PatchAlarm(ctx context.Context, request api.PatchAlarmReq
...
+
+	// There must always be a single record
+	if len(records) != 1 {
+		return nil, fmt.Errorf("expected a single Alarm Service Configuration record, but got %d", len(records))
+	}
+
+	serviceConfigRecord := records[0]
+
+	// Patch the Alarm Service Configuration
+	if request.Body.RetentionPeriod != 0 {
```
alegacy, line 382: Perhaps we should make this a non-required field so that we can distinguish between it not being set at all by the client and being set below the minimum.

internal/service/alarms/internal/db/migrations/000006_create_alarm_service_configuration_table.up.sql
```diff
@@ -0,0 +1,26 @@
...
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(), -- Unique identifier for each alarm service configuration
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP, -- Record creation timestamp
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP -- Record last update timestamp
+);
+
+-- Trigger function to update updated_at timestamp for alarm_service_configuration
+CREATE OR REPLACE FUNCTION update_alarm_service_configuration_timestamp()
+    RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = CURRENT_TIMESTAMP;
```
alegacy, line 17: Can't we do this simply in code rather than rely on a trigger?

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -35,6 +36,55 @@ func (ar *AlarmsRepository) GetAlarmEventRecord(ctx context.Context, id uuid.UUI
...
+}
+
+// GetServiceConfigurations grabs all rows of alarm_service_configuration
+func (ar *AlarmsRepository) GetServiceConfigurations(ctx context.Context) ([]models.ServiceConfiguration, error) {
+	return utils.FindAll[models.ServiceConfiguration](ctx, ar.Db)
+}
+
+// UpdateServiceConfiguration updates a row of alarm_service_configuration using a primary key
+func (ar *AlarmsRepository) UpdateServiceConfiguration(ctx context.Context, id uuid.UUID, record *models.ServiceConfiguration) (*models.ServiceConfiguration, error) {
+	return utils.Update[models.ServiceConfiguration](ctx, ar.Db, id, *record, " RetentionPeriod", "Extensions")
```
alegacy, line 85: Extra space in `" RetentionPeriod"`

internal/service/alarms/serve.go
```diff
@@ -110,6 +110,13 @@ func Serve(config *AlarmsServerConfig) error {
 		}
 	}
 
+	// Add Alarm Service Configuration to the database
+	serviceConfig, err := alarmRepository.CreateServiceConfiguration(ctx, internal.DefaultRetentionPeriod)
+	if err != nil {
+		return fmt.Errorf("failed to create alarmservice configuration: %w", err)
```
alegacy, line 116: missing space in `alarmservice`

internal/service/alarms/internal/db/migrations/000006_create_alarm_service_configuration_table.down.sql
```diff
@@ -0,0 +1,8 @@
+-- Drop the trigger that updates updated_at for alarm_service_configuration
```
alegacy, line 1: I've had this discussion with @pixelsoccupied before... IMO, we shouldn't be adding any down commands for this set of initial scripts.  These initial scripts essentially represent our "baseline" schema and later once we've reached a more productized state we can start adding incremental updates with related down scripts if necessary.

internal/service/alarms/api/openapi.yaml
mlguerrero12: yeah, I know, but I didn´t want to have to maintain the constant in code and here. But yeah, will add it

internal/service/alarms/internal/alarms_server.go
mlguerrero12: It makes sense that it was nil, but I'm following the spec. The field is expected.

internal/service/alarms/internal/db/migrations/000006_create_alarm_service_configuration_table.up.sql
mlguerrero12: We took this convention for alarm server. At least, for the update_at field. If we are gonna change it, then we will need another pr to modify all the tables

internal/service/alarms/internal/db/migrations/000006_create_alarm_service_configuration_table.down.sql
alegacy: ok

internal/service/alarms/api/openapi.yaml
```diff
@@ -221,6 +222,106 @@ paths:
               schema:
                 $ref: '../../common/api/openapi.yaml#/components/schemas/ProblemDetails'
 
+  /o2ims-infrastructureMonitoring/v1/alarmServiceConfiguration:
```
pixelsoccupied, line 225: sorry think i missed the discussion...but did we decide this already? basically for phase 1 we were going to hardcode the retention period to 24h. And make it configureable later on. Alternatively instead of exposing this in api we could pass this info in the innovatory CR or configMap that we can we reload right?

if already decided lgtm but im totally out of the loop! fyi @browsell

internal/service/alarms/api/openapi.yaml
```diff
@@ -576,6 +677,23 @@ components:
...
 
+    AlarmServiceConfiguration:
+      type: object
+      properties:
+        retentionPeriod:
+          type: integer
+          description: | 
+            Number of days for alarm history to be retained.
+            This value has cannot be set lower than 1 (day).
+          example: 30
```
pixelsoccupied, line 688: Not sure how flexible we want to be...consider something similar to to say Prometheuse retention period. So allow user to put something like `30d` or `24h`.

But if settling with int..try hour instead of day...that we can simply use the time.ParseDuration to get the interval.

internal/service/alarms/api/openapi.yaml
mlguerrero12: This is coming from the specification.

3.3.4.6 REST resource: Alarm Service Configuration (O2IMS-INTERFACE-R004-v07.00)
Table 4.2.1.4.13.2-1 Attributes Properties for AlarmList (O-CLOUD-IM.0-R004-v02.00)
3.7.13 Alarm Purge Use Case (ORCH-USE-CASES-R003-v11.00)

internal/service/alarms/api/openapi.yaml
mlguerrero12: This is set by the specification. I mean, that the value represent days. They do not mention what the minimum is, just that there is a minimum. I'm proposing 1 (day)

pixelsoccupied: ah then all good

mlguerrero12:
@alegacy, I fixed the ci issue

mlguerrero12:
@pixelsoccupied, can we merge this?

PR #431, opened 2025-01-02
CNF-15902: notify alarms subscriptions

pixelsoccupied:
- After capturing all the alerts we notify any subscriber using alarm sequence, event cursor and filter to only send the latest applicable. 
- Triggers are combined to managed the events during insert or update. 
- Some cleanup to use existing http client.


internal/service/alarms/api/server.go
```diff
@@ -502,12 +514,120 @@ func (a *AlarmsServer) AmNotification(ctx context.Context, request api.AmNotific
 		return nil, fmt.Errorf("%s: %w", msg, err)
 	}
 
-	//TODO: Get subscriber
+	// Return 200 before subscription processing to free up AM conn
+	a.Wg.Add(1)
```
alegacy, line 518: IMO, it would be cleaner if this part was decoupled and moved to a separate component of the alarm server to behave more like the way it was setup for the resource and cluster servers.   There is some duplication here with what the `Notifier` was intended to do.   

This could be aligned easily if the code above materialized each of the alarm changes into a pending notifications table and sent an event over to the `Notifier`.  If the proper hooks were implemented then the subscription filtering and cleanup of completed notifications would be handled in the same way that this is done for the resource/cluster data change notifications...including the cursor handling.

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -357,3 +357,78 @@ func getGetAlertFingerPrintAndStartAt(am *api.AlertmanagerNotification) []bob.Ex
...
+	}
+
+	record, err := utils.ExecuteCollectExactlyOneRow[models.AlarmSubscription](ctx, ar.Db, sql, params)
+	if err != nil {
+		return fmt.Errorf("failed to execute UpdateSubscriptionEventCursor query: %w", err)
+	}
+
+	slog.Info("Successfully updated subscription with new event cursor", "subscription", subscription.SubscriptionID, "from", subscription.EventCursor, "to", record.EventCursor)
+	return nil
+}
```
alegacy, line 434: I think this could be replaced with a call to `utils.Update` similar to this:

```
return utils.Update(ctx, ar.Db, subscription.SubscriptionID, subscription, []string{'EventCursor"}...)
```

internal/service/common/notifier/event.go
```diff
@@ -46,26 +50,45 @@ func processEvent(ctx context.Context, logger *slog.Logger, client *http.Client,
...
+	_ = CallUrl(ctx, logger, client, url, event)
+
+	completionChannel <- &SubscriptionJobComplete{
+		subscriptionID: subscriptionID,
+		notificationID: event.NotificationID,
+		sequenceID:     event.SequenceID,
+	}
+}
+
+func CallUrl(ctx context.Context, logger *slog.Logger, client *http.Client, url string, event Notification) error {
```
alegacy, line 62: As mentioned in another comment, rather than exposing this and using it directly from places other than directly within the `Notifier` I'd prefer if we align the alarm functionality to how the resource/cluster components work so that the `Notifier` handles the subscriber jobs and sending of the notifications.   One reason for this is that I plan on refactoring this to allow for sending to different types of subscribers based on the callback URL (i.e., special handling for SMO notifications vs internal notifications between our servers that wouldn't require (or support) OAuth tokens).

internal/service/alarms/serve.go
```diff
@@ -91,6 +89,9 @@ func Serve(config *api.AlarmsServerConfig) error {
 		if err != nil {
 			return fmt.Errorf("failed to parse global cloud id: %w", err)
 		}
+	} else {
+		slog.Debug("No globalCloudID set by the operator, generating one instead")
```
alegacy, line 93: We shouldn't be generating our own here.   The user must set this for the system to work as designed.   We should leave this as the nil UUID if not set... and continue to block the creation of new subscription until the user sets it.

internal/service/alarms/api/server.go
pixelsoccupied: /hold

looking into it!

internal/service/alarms/serve.go
pixelsoccupied: Yeah i added this just to quickly test things end to end. removing this.

internal/service/alarms/internal/db/migrations/000003_create_alarm_event_record_table.up.sql
```diff
@@ -32,68 +33,102 @@ CREATE TABLE IF NOT EXISTS alarm_event_record (
...
+                NEW.alarm_definition_id IS DISTINCT FROM OLD.alarm_definition_id OR
+                NEW.probable_cause_id IS DISTINCT FROM OLD.probable_cause_id)
+            THEN
+                NEW.notification_event_type := 'CHANGE';
+                NEW.alarm_changed_time := CURRENT_TIMESTAMP;
+                NEW.alarm_sequence_number := nextval('alarm_sequence_seq');
+            END IF;
+        END IF;
+
+        RETURN NEW;
```
alegacy, line 121: We've discussed this briefly in the past, but I feel like it is worth mentioning again ... I'm a bit concerned that we're using the AlarmEventRecord table to do double-duty in also storing event/notification type information.   If two or more changes occur to any given alarm in a short period of time then we may not be able to send off both notifications and the client would only get the last event.   Clients may want to see all transitions as they may want to take action on each transition.

pixelsoccupied: (just noting it here what we discussed) 

we do track all the changes back to back...partially. e.g if we have a subscriber that has a filter "CHANGE" (so notify me for everything that's NOT change), and then we have a notification that was firing, and the resolved and then firing again and now the subscriber is added, the subscriber will receive a payload with two events - one for the resolved and one for firing. The first one resolved (updated from firing to resolved) and then finally the second (firing with the new start time but same fingerprint, resulting in a new row/event) 

But its true 3 things happened but the client is getting 2. But given that its currently resolved..it somewhat implied that originally it was firing?

internal/service/alarms/internal/notifier_provider/notification.go
```diff
@@ -0,0 +1,92 @@
...
+	} else {
+		slog.Info("Found subscriptions to notify", "subscription count", len(subscriptions))
+		// Find the min event cursor, this will decide the total events that needs to be there ignoring the filter
+		alarms, err := n.repository.GetAlarmsForSubscription(ctx, *getMinSubscription(subscriptions))
+		if err != nil {
+			return notifications, fmt.Errorf("failed to get alarms for subscription: %w", err)
+		}
+
+		// Convert alert to notification and using notificationID as subscriberID
+		for _, sub := range subscriptions {
```
alegacy, line 50: Could you maybe add a `TODO` here about what we discussed yesterday regarding needing to refactor this a bit so that we aren't directly cloning the notifications for each subscription and instead have this handled a bit more generically directly in the notifier/worker via an interface callback to insert the consumer subscription id directly into the payload before sending.

As I said, I'm quite happy/willing to do that refactoring since I'm the one that missed this issue with the consumer subscription id!!

pixelsoccupied: Yeah i got one todo under Match! Will add a note there too!

internal/service/alarms/api/server.go
```diff
@@ -41,6 +44,14 @@ type AlarmsServer struct {
 	AlarmsRepository *repo.AlarmsRepository
 	// ClusterServer contains the cluster server client and fetched objects
 	ClusterServer *clusterserver.ClusterServer
+	// Wg to allow alarm server level background tasks to finish before graceful exit
+	Wg sync.WaitGroup
+	// NotificationProvider to handle new events
+	NotificationProvider notifier.NotificationProvider
```
alegacy, line 50: Are the last 3 necessary in the `AlarmsServer` object or can they live as locals in the start function?

internal/service/alarms/internal/notifier_provider/notification.go
pixelsoccupied: updated

internal/service/alarms/api/server.go
pixelsoccupied: Removed Wg since now we also check the notification during restart...so no need to wait. 
Removed SubscriptionProvider since its not used. 

But keeping Notifier to call Notify + NotificationProvider since its used to grab the notifications with  GetNotifications (before Notify)

pixelsoccupied: On second thought...brought back the server Wg. Hopefully its going to be basically noop - but if we ever have a something stuck in notify during server exit will exit with error. wdyt?

internal/service/alarms/api/server.go
```diff
@@ -503,9 +526,23 @@ func (a *AlarmsServer) AmNotification(ctx context.Context, request api.AmNotific
 		return nil, fmt.Errorf("%s: %w", msg, err)
 	}
 
-	//TODO: Get subscriber
+	// Return 200 before subscription processing to free up AM conn
+	a.Wg.Add(1)
+	go func() {
```
alegacy, line 531: This is ok for now, but I've had a chance to think about this since you raised in our weekly call and I am no longer sure this is the correct way to go.  If the client is creating a new subscription then it has the ability to use the GET API to get the current state of the alarms once its subscription is created.  That will give it the full picture without us having to replay all of the notifications that may have occurred prior to it creating its subscription.   Essentially the cluster and resource server subscription work that way as well... ie. create a subscription and get new stuff.  ... want to know what happened before then use the GET to go get those objects.

So.. in the interest of time I think we merge this as is and then open another PR to remove this functionality.  

cc:  @browsell

internal/service/alarms/internal/notifier_provider/notification.go
```diff
@@ -0,0 +1,96 @@
...
+	var notifications []notifier.Notification
+	// Get all subscriptions
+	subscriptions, err := n.repository.GetAlarmSubscriptions(ctx)
+	if err != nil {
+		return notifications, fmt.Errorf("failed to get all subscriptions for alarms: %w", err)
+	}
+	if len(subscriptions) == 0 {
+		slog.Info("No subscriptions to notify")
+		return notifications, nil
+	} else {
```
alegacy, line 41: `else` is redundant here since the `if` ends in a `return`... can save some indent by removing it.

internal/service/alarms/internal/notifier_provider/subscription.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (s *SubscriptionStorageProvider) UpdateSubscription(ctx context.Context, subscription *notifier.SubscriptionInfo) error {
+	if err := s.repository.UpdateSubscriptionEventCursor(ctx, models.AlarmSubscription{
+		SubscriptionID: subscription.SubscriptionID,
+		EventCursor:    int64(subscription.EventCursor),
+	}); err != nil {
+		return fmt.Errorf("update subscription failed for %s: %w", subscription.SubscriptionID, err)
+	}
+
+	slog.Info("Subscription courser updated", "to", subscription.EventCursor)
```
alegacy, line 60: typo `courser` -> `cursor`

internal/controllers/utils/oauth.go
```diff
@@ -71,12 +77,25 @@ func SetupOAuthClient(ctx context.Context, config *OAuthClientConfig) (*http.Cli
...
+
+		// Verify token acquisition works
+		if src, ok := oauthClient.Transport.(*oauth2.Transport); ok {
+			// Force token acquisition now to verify it works
+			_, err := src.Source.Token()
+			if err != nil {
+				return nil, fmt.Errorf("failed to acquire initial OAuth token: %w", err)
+			}
+			slog.Info("Successfully acquired initial OAuth token")
+		}
```
alegacy, line 91: I don't think that failing here is a reasonable approach.  The OAuth server may be momentarily unreachable... i'd rather defer this type of failure when the token is actually needed.

internal/service/alarms/api/server.go
pixelsoccupied: +1

internal/service/alarms/internal/notifier_provider/notification.go
pixelsoccupied: removed!

internal/service/alarms/internal/notifier_provider/subscription.go
pixelsoccupied: updated

internal/controllers/utils/oauth.go
pixelsoccupied: removed

PR #432, opened 2025-01-02
CNF-16193: Templates REST API internal server - GET ALL

irinamihai:
Description:
- Implement the artifacts server and the code for getting all the `ClusterTemplates` from a hub cluster
- The ClusterTemplates are converted to the `ManagedInfrastructureTemplate` format by following the rules:
```
clustertemplate.name                    -> managedInfrastructureTemplate.name
clustertemplate.version                 -> managedInfrastructureTemplate.version
clustertemplate.description             -> managedInfrastructureTemplate.description 
clustertemplate.templateParameterSchema	-> managedInfrastructureTemplate.parameterSchema
clustertemplate.templateId              -> managedInfrastructureTemplate.artifactResourceId
```
- For local runs (the only supported method so far):
```console
./bin/oran-o2ims artifacts-server serve
curl http://127.0.0.1:8000/o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates | jq
```
- Authentication has not been implemented yet.
- Filtering & running as a pod will be addressed in a future PR.

### DIFF
--- a/internal/service/artifacts/serve.go
+++ b/internal/service/artifacts/serve.go
@@ -1,0 +1,121 @@
+package resources
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/openshift-kni/oran-o2ims/internal/service/artifacts/api"
+	"github.com/openshift-kni/oran-o2ims/internal/service/artifacts/api/generated"
+	common "github.com/openshift-kni/oran-o2ims/internal/service/common/api"
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/clients/k8s"
+)
+
+// Resource server config values
+const (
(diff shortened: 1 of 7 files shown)


internal/service/artifacts/serve.go
```diff
@@ -0,0 +1,119 @@
...
+
+	serverStrictHandler := generated.NewStrictHandlerWithOptions(&server, nil,
+		generated.StrictHTTPServerOptions{
+			RequestErrorHandlerFunc:  common.GetOranReqErrFunc(),
+			ResponseErrorHandlerFunc: common.GetOranRespErrFunc(),
+		},
+	)
+
+	router := http.NewServeMux()
+
```
pixelsoccupied, line 65: one thing that you might need is to add a default route - something like this 
https://github.com/openshift-kni/oran-o2ims/blob/main/internal/service/alarms/serve.go#L123-L125

irinamihai: Thank you. Added.

PR #434, opened 2025-01-03
Handle alarm_change_time only on trigger for all operations

mlguerrero12:
This addresses the following:

- When alarm is cleared, the changed time is equal to the cleared time to properly show that the record was updated due to this and not give the impression that a subsequent change was made.

- When an acknowledged alarm is resolved, the change time is not updated.

- Handle alarm_change_time column only in the trigger. Remove setting this on the API.

- Have a consistent behavior. When an alarm is created, the changed time is equal to the raised time. When it is resolved, the change time is equal to the cleared time. When it is acknowledged, the change time is equal to the change time. Any other updates have CURRENT_TIMESTAMP set by the trigger


mlguerrero12:
@pixelsoccupied PTAL

PR #435, opened 2025-01-03
Update default VERSION to 4.18.0

donpenney:


donpenney:
It looks like this update will require https://github.com/openshift/release/pull/60306 to be merged for there to be a  quay.io/openshift-kni/oran-o2ims-operator:4.18.0 image in the repo, to be pulled during the scorecard test.

PR #443, opened 2025-01-07
CNF-16211: Implement Webhook validation for ProvisioningRequest creation and update

Missxiaoguo:
- Move neccessary functions to provisioning api so they can be shared between both webhook and controller
- Implement ValidateCreate function
   * reject the CR if required keys are missing (e.g., nodeClusterName, oCloudSiteId, clusterInstanceParameters, clusterInstanceParameters) 
   * reject the CR if ClusterInstance input doesn't match the schema
- Implement ValidateUpdate function
   * reject the CR if required keys are missing
   * reject the CR if ClusterInstance input doesn't match schema
   * reject the CR for immutable ClusterInstance fields changes

Example outputs
```
error when creating "sampleprovisioningrequests/provisioningrequest-cnfdf20.yaml": admission webhook "provisioningrequests.o2ims.provisioning.oran.org" denied the request: spec.templateParameters does not match the schema defined in ClusterTemplate (sno-ran-du.v4-16-21-v1) spec.templateParameterSchema: invalid input: (root): oCloudSiteId is required; (root): clusterInstanceParameters is required
 
error when patching "sampleprovisioningrequests/provisioningrequest-cnfdf20.yaml": admission webhook "provisioningrequests.o2ims.provisioning.oran.org" denied the request: only "extraAnnotations" and/or "extraLabels" changes in spec.TemplateParameters.ClusterInstanceParameters are allowed once cluster installation has started or reached to Completed/Failed state, detected changes in immutable fields: clusterName
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,7 +363,7 @@ go-generate:
 .PHONY: test tests
 test tests:
 	@echo "Run ginkgo"
-	HWMGR_PLUGIN_NAMESPACE=hwmgr ginkgo run -r $(ginkgo_flags)
+	HWMGR_PLUGIN_NAMESPACE=hwmgr ginkgo run -r ./internal ./api $(ginkgo_flags)
 
 .PHONY: fmt
 fmt:


Missxiaoguo:
/hold
to debug OOMKilled

api/provisioning/v1alpha1/provisioningrequest_webhook.go
```diff
@@ -64,3 +95,59 @@ func (r *ProvisioningRequest) ValidateDelete() (admission.Warnings, error) {
...
+		return nil
+	}
+
+	// Check for updates to immutable fields in the ClusterInstance input.
+	// Once provisioning has started or reached a final state (Completed or Failed),
+	// updates to immutable fields in the ClusterInstance input are disallowed,
+	// with the exception of scaling up/down when Cluster provisioning is completed.
+	// TODO: consider exposing the conditions in the API.
+	crProvisionedCond := meta.FindStatusCondition(r.Status.Conditions, "ClusterProvisioned")
+	if crProvisionedCond != nil && crProvisionedCond.Reason != "Unknown" {
```
irinamihai, line 127: I think we should make these conditions constants and add a comment that they should be kept in sync with those under [internal/controllers/utils/conditions.go.](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/utils/conditions.go#L22-L46)
It would be easy to get the conditions mismatched.
It would actually be even better to move these conditions to completely into the `api/provisioning` directory and have them referenced in both the webhook and the controller code under `internal/`.

Missxiaoguo: > It would actually be even better to move these conditions completely into the api/provisioning directory and have them referenced in both the webhook and the controller code under internal/.

Yep, I was thinking the same, since the movement may require a lot of places to be updated so I added the TODO for that. Maybe it's not clear enough.

irinamihai: Ok, I see it now. Sounds good to me.

PR #444, opened 2025-01-07
CNF-16246: Add list of cluster resources to node cluster

alegacy:
This adds the ability to return the list of cluster resource id values as part of an API query for node cluster(s).  These values are not persisted to the database, but are queried on an as-needed basis in response to an API query.

### DIFF
--- a/internal/service/cluster/api/server.go
+++ b/internal/service/cluster/api/server.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/google/uuid"
+
 	api "github.com/openshift-kni/oran-o2ims/internal/service/cluster/api/generated"
 	"github.com/openshift-kni/oran-o2ims/internal/service/cluster/db/models"
 	"github.com/openshift-kni/oran-o2ims/internal/service/cluster/db/repo"
@@ -165,9 +167,26 @@ func (r *ClusterServer) GetNodeClusters(ctx context.Context, request api.GetNode
 		return nil, fmt.Errorf("failed to get node clusters: %w", err)
 	}
 
+	// Retrieve the list of ClusterResourceID values per NodeCluster in one query so that we don't have to look it up
+	// on a per NodeCluster basis.
+	nodeClusterResourceIDs, err := r.Repo.GetNodeClusterResourceIDs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get node cluster resource ids: %w", err)
+	}
+
+	// Store them in a map so they are easier to access for the next operation.
+	mapper := make(map[uuid.UUID][]uuid.UUID)
+	for _, entry := range nodeClusterResourceIDs {
+		mapper[entry.NodeClusterID] = entry.ClusterResourceIDs
+	}
+
 	objects := make([]api.NodeCluster, len(records))
 	for i, record := range records {
-		objects[i] = models.NodeClusterToModel(&record)
+		clusterResourceIDs, found := mapper[record.NodeClusterID]
+		if !found {
+			clusterResourceIDs = make([]uuid.UUID, 0)
+		}
+		objects[i] = models.NodeClusterToModel(&record, clusterResourceIDs)
 	}
 
 	return api.GetNodeClusters200JSONResponse(objects), nil
@@ -194,7 +213,17 @@ func (r *ClusterServer) GetNodeCluster(ctx context.Context, request api.GetNodeC
 		}, nil
 	}
 
-	object := models.NodeClusterToModel(record)
+	resources, err := r.Repo.GetNodeClusterResources(ctx, record.NodeClusterID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get node cluster resources: %w", err)
+	}
+
+	ids := make([]uuid.UUID, len(resources))
+	for i, resource := range resources {
+		ids[i] = resource.ClusterResourceID
+	}
+
+	object := models.NodeClusterToModel(record, ids)
 	return api.GetNodeCluster200JSONResponse(object), nil
 }
 
--- a/internal/service/cluster/api/server.go
+++ b/internal/service/cluster/api/server.go
@@ -182,11 +182,7 @@ func (r *ClusterServer) GetNodeClusters(ctx context.Context, request api.GetNode
 
 	objects := make([]api.NodeCluster, len(records))
 	for i, record := range records {
-		clusterResourceIDs, found := mapper[record.NodeClusterID]
-		if !found {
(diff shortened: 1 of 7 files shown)


internal/service/cluster/api/server.go
```diff
@@ -165,9 +167,26 @@ func (r *ClusterServer) GetNodeClusters(ctx context.Context, request api.GetNode
...
+	}
+
 	objects := make([]api.NodeCluster, len(records))
 	for i, record := range records {
-		objects[i] = models.NodeClusterToModel(&record)
+		clusterResourceIDs, found := mapper[record.NodeClusterID]
+		if !found {
+			clusterResourceIDs = make([]uuid.UUID, 0)
+		}
+		objects[i] = models.NodeClusterToModel(&record, clusterResourceIDs)
```
mlguerrero12, line 189: objects[i] = models.NodeClusterToModel(&record, mapper[record.NodeClusterID]) since you are checkin nil inside NodeClusterToModel

internal/service/cluster/api/server.go
```diff
@@ -194,7 +213,17 @@ func (r *ClusterServer) GetNodeCluster(ctx context.Context, request api.GetNodeC
 		}, nil
 	}
 
-	object := models.NodeClusterToModel(record)
+	resources, err := r.Repo.GetNodeClusterResources(ctx, record.NodeClusterID)
```
mlguerrero12, line 216: is it not possible to do the same trick as in the List? Perhaps adding a where in the query with the nodeclusterid?

internal/service/cluster/api/server.go
alegacy: Ok, i'll push this down into the conversion function.

internal/service/cluster/db/models/converters.go
```diff
@@ -37,19 +37,24 @@ func ClusterResourceTypeToModel(record *ClusterResourceType) generated.ClusterRe
...
 		ArtifactResourceId:             record.ArtifactResourceID,
 		ClientNodeClusterId:            record.ClientNodeClusterID,
 		ClusterDistributionDescription: record.ClusterDistributionDescription,
 		ClusterResourceGroups:          record.ClusterResourceGroups,
-		ClusterResourceIds:             []uuid.UUID{}, // TODO: caller should pass in this list
 		Description:                    record.Description,
 		Extensions:                     record.Extensions,
 		Name:                           record.Name,
 	}
+
```
mlguerrero12, line 52: I'm a bit confused, the field is required by the api, but if you don't set it, it will be nil. Will not the api complain about this? In other words, this function should never be called clusterResourceIDs as nil. Perhaps setting clusterResourceIDs = make([]uuid.UUID, 0) if nil is safer?

internal/service/cluster/api/server.go
alegacy: Just to be clear, you are referring to a custom query like this one?
```sql
SELECT nc.*, array_agg(cluster_resource_id) as cluster_resource_ids 
  FROM node_cluster AS nc 
    JOIN cluster_resource AS cr ON (cr.node_cluster_id = nc.node_cluster_id)
  WHERE nc.node_cluster_id = $1
  GROUP BY nc.node_cluster_id
```

internal/service/cluster/db/models/converters.go
alegacy: yes, you are correct.  I overlooked that it was mandatory.

internal/service/cluster/collector/collector.go
```diff
@@ -18,6 +18,10 @@ import (
 
 const pollingDelay = 10 * time.Minute
 
+// clusterNameExtension represents a mandatory extension that data sources must add to ClusterResource objects to
+// identify their parent NodeCluster.
```
mlguerrero12, line 22: it seems to be something missing in the spec (draft) - perhaps it will be added later. But is it not most likely that they add the id instead of the name?

internal/service/cluster/collector/collector.go
```diff
@@ -195,6 +205,13 @@ func (c *Collector) collectClusterResources(ctx context.Context, dataSource Data
 
 	// Loop over the set of cluster resources and insert (or update) as needed
 	for _, resource := range resources {
+		// Set the cluster ID since the data source doesn't have access to this info
+		if resource.Extensions != nil {
+			clusterName := (*resource.Extensions)[clusterNameExtension].(string)
+			resource.NodeClusterID = clusterNameToID[clusterName]
+			delete(*resource.Extensions, clusterNameExtension)
```
mlguerrero12, line 212: here you add the id, I see. Why deleting the name?

internal/service/cluster/api/server.go
mlguerrero12: is the join needed? I meant adding a where to the raw query you have in `GetNodeClusterResourceIDs`

alegacy: Oh, I see ... yes, I suppose we could do that.  Given that it would still only return a few tuples at most there wouldn't be much advantage.   I thought what you meant was to combine both queries (the node cluster, and the resources) into a single query.

internal/service/cluster/collector/collector.go
alegacy: I can add a comment to clarify what I am doing here.  The problem is that the `ClusterResource` needs to have its `NodeClusterID` value set, but there's no way to get that value from inside of `GetClusterResources` so I'm adding the node cluster name into the extensions, and then having the caller convert that to a NodeClusterID and then removing the name from the extensions since it was merely used as a way to pass the cluster name down to the caller.  

A bit of a hack, but there are limited options here.  I considered passing the cluster name to id map into `GetClusterResources` but as I have been looking into handling the async change notifications recently I realized that it would be a problem in that scenario so to avoid adding it now, and then having to remove/rewrite it later I opted for this simpler approach.

internal/service/cluster/collector/collector.go
alegacy: see explanation above.  i'll add a better comment.

internal/service/cluster/collector/collector.go
alegacy: Added a comment.

internal/service/cluster/db/models/converters.go
alegacy: fixed

internal/service/cluster/api/server.go
alegacy: Added a `where` clause to the original and re-used it.

internal/service/cluster/api/server.go
alegacy: fixed

PR #445, opened 2025-01-07
Fix kustomize layering for debug mode

alegacy:
The introduction of the webhook validation affected the ability to build with DEBUG=yes so this commit re-arranges how the lists are defined and patched to work around the issue.


donpenney (commented):
/lgtm

PR #448, opened 2025-01-08
CNF-16243: Transition to new resource server

alegacy:
This commit transitions to using the new implementation of the resource
server.  This enables the use of notifications to subscribers when
inventory data changes.  For now this is limited to polling of inventory
at 10 minute intervals.  In the future this will evolve to an
edge-driven approach depending on whether the hardware manager plugin
supports this type of update mechanism.


internal/controllers/utils/utils.go
```diff
@@ -462,45 +454,8 @@ func GetServerArgs(inventory *inventoryv1alpha1.Inventory, serverName string) (r
 			result,
 			fmt.Sprintf("--cloud-id=%s", inventory.Status.ClusterID))
```
alegacy, line 456: I accidentally removed the SMO args from here.

/hold

alegacy: fixed

alegacy: /cancel

irinamihai:
/lgtm
I suppose cleaning up the deprecated commands & parameters is planned for a future commit?

alegacy:
> /lgtm I suppose cleaning up the deprecated commands & parameters is planned for a future commit?

Absolutely!  For this first pass I wanted to cleanup the operator and make sure we didn't make this difficult for adding new servers.

PR #453, opened 2025-01-09
Fix notifier deadlock on new events

alegacy:
The design of the notifier and notifier worker interactions was flawed in that it allowed the possibility of a deadlock whenever both the worker's work channel and notifier's completion channel buffers were full.  The notifier cannot send new work to the worker, and the worker gets stuck waiting for the notifier to acknowledge receipt of the completion event.

This is being mitigated by having the notifier enqueue work onto the worker's work queue directly, and only signaling the work channel upon adding the first event onto the queue.  The worker will independently process the queue and report completion down to the notifier.  This should eliminate the possibility that the completion channel fills up and the worker blocking indefinitely.

### DIFF
--- a/internal/service/common/notifier/notifier.go
+++ b/internal/service/common/notifier/notifier.go
@@ -12,7 +12,12 @@ import (
 )
 
 // DefaultBufferedChannelSize defines the default size for buffered channels used across the notifier.
-const DefaultBufferedChannelSize = 10
+const DefaultBufferedChannelSize = 100
+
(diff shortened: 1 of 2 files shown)


alegacy:
/hold

need to discuss with @pixelsoccupied  and retest a couple of scenarios.

pixelsoccupied:
looks good tried with 3500 alerts (previously was showing issues around 10+)! 

/lgtm 
/approve

PR #456, opened 2025-01-09
Add support to override imagePullPolicy for designer build

donpenney:


alegacy:
Is this just a concern for the operator or should we try to propagate this to the server configs that are dynamically created?  The current approach of not specifying the policy seems to be working fine.

donpenney:
> Is this just a concern for the operator or should we try to propagate this to the server configs that are dynamically created? The current approach of not specifying the policy seems to be working fine.

The easiest approach here would be to extend this deployment patch to also set an env variable in the manager container which could then be looked up when creating these additional deployments. Other option would be to query the running pod CR to check the spec and reuse the imagePullPolicy from there.

If imagePullPolicy is not set, the default behaviour is to use "Always" if the image tag is "latest" or unset, and "IfNotPresent" if it's a  SHA reference or non-latest tag. The intent of this update is to use "Always" for designer builds, to avoid inadvertent use of stale images if using a versioned tag. When building a bundle to be installed from catalog with a subscription, you can't use VERSION=latest.

This update also provides a means for the designer to specify IMAGE_PULL_POLICY when building, to override this default.

PR #458, opened 2025-01-10
Integrate the artifacts server in the inventory controller

irinamihai:
Description:
- Integrate the Artifacts server in the inventory controller:
  - add the new server in the default `Inventory` CR
  - create the ClusterRoles & ClusterRoleBinding associated
  - add the new path (`/o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates`) to the `api` Ingress
  - add the new URLs in the test `oran-o2ims-test-client-role` ClusterRole
  - add unit tests
- Command example:
```console
curl -kq \
    'https://o2ims.apps.${MY_CLUSTER}/o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates?filter=(eq,version,v4-17-3-2)%3b(eq,name,sno-ran-du)' \
    -H "Authorization: Bearer ${MY_TOKEN}"  | jq
```


bundle/manifests/o2ims.oran.openshift.io_inventories.yaml
```diff
@@ -67,6 +67,19 @@ spec:
                 required:
                 - enabled
                 type: object
+              artifactsServerConfig:
+                default:
```
browsell, line 71: Outside of perhaps lab debugging is there a scenario where we need this switch  vs just unconditionally starting the server ?

irinamihai: I followed the model of the other servers that have this `enabled` switch which is by default set to `true`. I did add the server in the `default` Inventory CR, so when the O-Cloud manager is installed, the artifacts server will automatically start.

config/testing/client-service-account-rbac.yaml
```diff
@@ -16,6 +16,9 @@ rules:
   - /o2ims-infrastructureMonitoring/api_versions
   - /o2ims-infrastructureCluster/*
   - /o2ims-infrastructureCluster/api_versions
+  - /o2ims-infrastructureArtifacts/*
+  - /o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates
+  - /o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates/*
```
alegacy, line 21: See my comment in inventory_controller.go regarding the new for us to support the `/api_versions` endpoint.  We can add that in a separate PR, but we should probably just setup the rbac in anticipation of that change now to save the hassle of doing it later.  

```
- /o2ims-infrastructureArtifacts/*
- /o2ims-infrastructureArtifacts/api_versions
```

bundle/manifests/o2ims.oran.openshift.io_inventories.yaml
alegacy: Ya, I agree... which should work towards removing these "enabled" boolean flags.  No idea outside of debugging.

alegacy: i would leave it as-is for now and remove them all in one shot later.

internal/controllers/inventory_controller.go
```diff
@@ -1438,6 +1585,21 @@ func (t *reconcilerTask) createIngress(ctx context.Context) error {
 									},
 								},
 							},
+							{
+								Path: "/o2ims-infrastructureArtifacts/v1/managedInfrastructureTemplates",
```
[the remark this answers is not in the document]
irinamihai: Agreed. Done.

config/testing/client-service-account-rbac.yaml
irinamihai: I've modified the test rbac for now and will include the support for the `/api_versions` support in a future PR.

PR #460, opened 2025-01-13
CNF-16245: Add edge driven cluster updates with Reflector

alegacy:
This adds the ability to watch for changes to ManagedCluster and Agent objects so that we can dynamically update our persistent storage rather than have to poll at fixed intervals.  The watch is implemented using a Reflector and supplying an adapter layer that receives cache update events and feeds them to the Collector which then updates our storage and emits async notification events to subscribers as needed.

### DIFF
--- a/internal/service/common/async/store.go
+++ b/internal/service/common/async/store.go
@@ -66,6 +66,8 @@ type ReflectorStore struct {
 	mutex sync.Mutex
 	// ready is used to signal that a new operation has been received from the Reflector
 	ready chan struct{}
+	// hwm is the high watermark of the queue length
+	hwm int
 }
 
 // NewReflectorStore creates a new ReflectorStore
@@ -86,8 +88,14 @@ func (c *ReflectorStore) enqueue(operation operation) {
 	if operation.eventType == SyncComplete {
 		c.hasSynced = true
 	}
-	if len(c.queue) == 1 {
+	count := len(c.queue)
+	if count == 1 {
 		c.ready <- struct{}{}
+	}
+	if count >= c.hwm+10 {
+		// We don't need a log every single time the high watermark is surpassed by 1, so log it when it surpasses by 10
+		slog.Debug("New reflector store queue high watermark", "new", count, "old", c.hwm)
+		c.hwm = count
 	}
 }
 
@@ -132,6 +140,7 @@ func (c *ReflectorStore) Receive(ctx context.Context, handler AsyncEventHandler)
 		select {
 		case <-ctx.Done():
 			slog.Info("stopping store adapter; context canceled")
+			close(c.ready)
 		case <-c.ready:
 			c.handleOperations(ctx, handler)
 		}
--- a/internal/service/common/async/store.go
+++ b/internal/service/common/async/store.go
@@ -1,0 +1,216 @@
+package async
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"sync"
+
+	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/db"
+)
+
+// AsyncEventType defines the types of async events that are supported between the Collector and Reflector.
+type AsyncEventType int
+
+const (
+	// Updated indicates that the object is inserted or updated
+	Updated AsyncEventType = iota
+	// Deleted indicates that the object is deleted
+	Deleted
+	// SyncComplete indicates that the Reflector has re-listed all objects from the API server
+	SyncComplete
+)
+
+// AsyncChangeEvent defines the data received by a data source to signal that an async handleWatchEvent has been received for a
+// given object type.
+type AsyncChangeEvent struct {
+	DataSourceID uuid.UUID
+	EventType    AsyncEventType
+	Object       db.Model
+	Keys         []uuid.UUID
+}
+
+// AsyncEventHandler is intended to be implemented by the Collector so that it can receive data from the Reflector via
+// the Store interface.
+type AsyncEventHandler interface {
+	// HandleSyncComplete is used to indicate that the full list of objects has been received.  Any stale objects should now
+	// be deleted.
+	HandleSyncComplete(ctx context.Context, objectType runtime.Object, keys []uuid.UUID) error
+	// HandleAsyncEvent is used to pass objects received from the Reflector to the Collector
+	HandleAsyncEvent(ctx context.Context, obj interface{}, eventType AsyncEventType) (uuid.UUID, error)
+}
+
+// operation defines an internal structure used to pass information around about pending operations.
+type operation struct {
+	eventType AsyncEventType
+	// objects is used to store the subject(s) of the operation.  If eventType is Updated or Deleted,
+	// then this slice will contain only a single element.  If eventType is SyncComplete, then it
+	// may contain 0 or more elements.
+	objects []interface{}
+}
+
+// ReflectorStore defines an adaptation layer between a Reflector and our Collector.  This is an alternative to using an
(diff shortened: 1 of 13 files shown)


alegacy:
I got the Jira number wrong... it is CNF-16245.  I'll amend the commit.

internal/service/common/async/store.go
```diff
@@ -0,0 +1,216 @@
...
+type operation struct {
+	eventType AsyncEventType
+	// objects is used to store the subject(s) of the operation.  If eventType is Updated or Deleted,
+	// then this slice will contain only a single element.  If eventType is SyncComplete, then it
+	// may contain 0 or more elements.
+	objects []interface{}
+}
+
+// ReflectorStore defines an adaptation layer between a Reflector and our Collector.  This is an alternative to using an
+// Informer which eliminates the need for caching objects in memory.  On a system engineered to capacity, this can save
```
mlguerrero12, line 56: Why is caching bad? All OCP operators I've worked in the past use caching/informers, even core payload ones as the cluster-network-operator. These servers will run in the hub so I don't think we will have memory constraints. ACM has many operators running in the hub as well.

Reading about informers and reflectors (I was not familiar with this one). The informer is a high level abstraction that besides caching offers built-in event handling, built-in periodic resyncs and error handling/recovery. Where as, the reflector is a low level abstraction lacking these things.

So, is it worth giving up the advantages of informers/caching for saving some hundreds of MB of memory? What am I missing?

alegacy: Informers are built on top of Reflectors.  In my view, caching hundreds of MB of objects in memory that we don't need is always an unnecessary use of resources.  I think we should do what we can to avoid such waste.

alegacy: In some use cases, the memory use could be justified, but in our use case where we are persisting objects to the database and serving our APIs based on the persisted copy there is no point is retaining all of those objects in memory.

alegacy:
rebased

internal/service/common/async/store.go
mlguerrero12: Like I said, we are not constraint by memory in the hub. I prefer to spend some hundreds of MB and use a high level abstraction resource that provides many advantages such as built-in periodic resync in case of missing events, built-in event handling, and extra error handling mechanism. Instead of us doing all that. And most importantly, that is proven to work in situations with high scalability requirements and is less complex. Informers seem to be the way to go in OCP based on what I've seen so far. In terms of maintainability, it is more likely to find people more familiar with informers instead of the reflector which is meant to use for low level applications. New people will have more difficulties understanding this than an informer. 

It seems we have different views. Let's see what other people think about this.

You calculated those hundreds of MB based on 3500 managed clusters, right?

browsell: Since we don't leverage caching in this case, it seems to be a poor use of resources to use an informer.

alegacy: @mlguerrero12  The informer is built on the reflector, so all of the advantages you listed above about scalability, fault resilience, error handling come from the reflector rather than the informer.  The informer basically just provides the memory cache part of it.

alegacy:
rebased ... again

internal/service/common/async/store.go
mlguerrero12: @alegacy With the reflector, you have to implement the handling of events. These are callbacks with informers. I also see you have to implement periodic resyncs when the watch has completed. This is also handled by the informer. And the informer implements the reflector, so any error and needed retries are handled by the informer. Here, you have to do it yourself. So, all this besides adding complexity, represent code that can fail. There are many points of failures (channels and routines). We are adding an extra layer that needs to be tested thoroughly.

This morning I was talking about this (general terms, nothing specific about this project) with some people from the controller-runtime/kubebuilder team of Red Hat. They acknowledged there is a consumption of memory by having a cache, but the benefits in most of the cases outweighs the memory load, especially in large clusters. It is very efficient. Of course, they said there is nothing wrong about using a Reflector, however, it gets more complex and dev maintainability decreases. And this reminded me of discussion we had (more than a year ago I think) for the sriov-network-operator about reducing the complexity of code. We currently have informers there and the proposal was to use the libraries of controller-runtime to simplify the code. Today I asked about this, and the PR is currently opened. Note that here NVIDIA contributors have strong scalability requirements. They have very large clusters. So, this is a better option, I would say. Use controller-runtime code instead of plain informers.

Like I said previously, we have different ways to see this. If you believe our application is different and if the memory consumption at the hub is the most important thing for us, and if you have support from other team members. Then, it's settled.

alegacy: Just to close the loop on this... the resync logic that we needed to write here would still be required for an Informer based implementation.  That's because the Informer is able to distill internal resyncs down to individual add/update/delete events because of its in-memory cache.  But, in the event of a pod restart it wouldn't be able to tell us what was added/updated/deleted relative to our persisted storage so we'd still need to implement and "end of sync" mechanism that purged stale records out of the database.

internal/service/common/utils/persist.go
```diff
@@ -149,7 +150,44 @@ func PersistObjectWithChangeEvent[T db.Model](ctx context.Context, db *pgxpool.P
...
 		if err != nil {
-			return nil, fmt.Errorf("failed to persist resource type data change object: %w", err)
+			return nil, fmt.Errorf("failed to persist data change object: %w", err)
+		}
+	}
+
+	err = tx.Commit(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to commit transaction: %w", err)
+	}
```
pixelsoccupied, line 160: There's also pgx.BeginFunc(...) helper func to basically the do the transaction. Might be a cleaner?

internal/service/common/async/store.go
```diff
@@ -0,0 +1,216 @@
...
+	// ObjectType is an instance of the object type monitored by the Reflector.
+	ObjectType runtime.Object
+	// hasSynced indicates whether the initial sync has completed (i.e., whether Replace has been called)
+	hasSynced bool
+	// queue stores the pending operations received from the Reflector
+	queue []operation
+	// mutex protects the shared instance variables from concurrency issues
+	mutex sync.Mutex
+	// ready is used to signal that a new operation has been received from the Reflector
+	ready chan struct{}
```
pixelsoccupied, line 68: ready is never closed? might be something we can do during shutdown?

internal/service/common/async/store.go
```diff
@@ -0,0 +1,216 @@
...
+		ready:      make(chan struct{}, 1),
+	}
+}
+
+// enqueue adds an operation to the queue and signals the receiver if it is the first operation
+// enqueued to the queue.
+func (c *ReflectorStore) enqueue(operation operation) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.queue = append(c.queue, operation)
```
pixelsoccupied, line 85: Wonder if it would benefit if we limit the size of the q. Like block until maxQueueSize for when operations are being added faster than they can be processed.

pixelsoccupied (commented):
very cool use of reflector. looking forward to using it!

internal/service/common/async/store.go
alegacy: I don't expect for this to get this out of hand given that there shouldn't be more than a few thousand in the worse case which is startup.   We could add a high-watermark type log here to track what the behaviour is first and then add some sort of rate limiting if we see it exceed what we think is 'normal'

alegacy: Added a log if the high watermark jumps by 10 (to avoid excessive logs)

internal/service/common/utils/persist.go
alegacy: fixed

PR #464, opened 2025-01-13
Fix notification payload

alegacy:
This change includes 3 commits to address issues with the notification payloads.

1) The ConsumerSubscriptionID wasn't being set.
2) The prior/post object states in the cluster and inventory notifications were serialized as strings rather than JSON.
3) Fix loading of OAuth config from environment variables


internal/service/alarms/internal/notifier_provider/subscription.go
```diff
@@ -63,3 +66,20 @@ func (s *SubscriptionStorageProvider) UpdateSubscription(ctx context.Context, su
...
+		return notification, nil
+	}
+
+	payload, ok := notification.Payload.(generated.AlarmEventNotification)
+	if !ok {
+		return nil, fmt.Errorf("notification payload is not of type AlarmEventNotification")
+	}
+	payload.ConsumerSubscriptionId = subscription.ConsumerSubscriptionID
+
+	result := deepcopy.Copy(*notification).(notifier.Notification)
```
alegacy, line 82: @pixelsoccupied  Is this `deepcopy.Copy` overkill?  I suppose since we're only modifying the top-level fields in the struct we can simply do a:

```
result := *notification
```

to do a shallow copy?

alegacy:
/hold

need to retest the alarm notifications... cluster and resource notifications works fine.

internal/service/alarms/internal/notifier_provider/subscription.go
pixelsoccupied: Yeah i think so too! Given its only the top level change + we don't really need any isolation from the original data...shallow should be good enough.

alegacy: thanks.  fixed.

internal/service/alarms/internal/notifier_provider/notification.go
```diff
@@ -46,33 +48,17 @@ func (n *NotificationStorageProvider) GetNotifications(ctx context.Context) ([]n
...
 
-	// TODO: Refactor to use a callback interface for subscription-specific payload modifications
-	// rather than cloning notifications for each subscription. This would improve efficiency
-	// and make the notification system more generic.
-	// As workaround for now using notificationID as subscriberID to link them with Match interface.
-	for _, sub := range subscriptions {
-		var subNotification []notifier.Notification
-		for _, alarm := range alarms {
-			if alarm.AlarmSequenceNumber > sub.EventCursor {
-				if sub.Filter != nil && alarm.NotificationEventType == *sub.Filter {
```
alegacy, line 57: Need to propagate t his to the `Matches` function

alegacy: fixed

alegacy:
> /hold
> 
> need to retest the alarm notifications... cluster and resource notifications work fine.

retested with alert manager.  notifications received at mock smo.

/hold cancel

internal/service/alarms/internal/notifier_provider/subscription.go
```diff
@@ -48,8 +49,16 @@ func (s *SubscriptionStorageProvider) GetSubscriptions(ctx context.Context) ([]n
...
+	payload, ok := notification.Payload.(generated.AlarmEventNotification)
+	if !ok {
+		slog.Warn("notification payload is not of type generated.AlarmEventNotification", "type", fmt.Sprintf("%T", notification.Payload))
+		return false
+	}
+	if subscription.Filter == nil {
+		return true
+	}
+	filter := generated.AlarmSubscriptionInfoFilter(*subscription.Filter)
+	return models.AlarmFilterToEventType(filter) == payload.NotificationEventType
```
pixelsoccupied, line 61: its such strange thing but the spec says to do the opposite. 

```
        filter:
          type: string
          enum: [ NEW, CHANGE, CLEAR, ACKNOWLEDGE ]
          description: |
            Criteria for events which do not need to be reported or will be filtered by the subscription 
            notification service. Therefore, if a filter is not provided then all events are reported.
            It can be filtered by criteria based on the type of notification of fields of the
            AlarmEventRecord.
          example: NEW
```

it matches if `!=`

alegacy: 😮  ... fixed

PR #469, opened 2025-01-15
CNF-16172: Integrate provisioning server into inventory controller

Missxiaoguo:
The provisioning server will automatically start when the default inventory CR is created.

Updates:
- Created neccessary resources to start the server. A common `createServerClusterRoleBinding` function is added to create clusterrolebinding for each server
- Included the path for provisioning in the `Ingress`.
- Added `status.usedServerConfig.provisioningServerUsedConfig`
- Included the deployment status of provisioning server, artifacts server and cluster server in the inventory `status.conditions`
- Updated test-client clusterrole to include new URLs for provisioningrequest


internal/controllers/inventory_controller.go
```diff
@@ -1278,6 +1263,41 @@ func (t *reconcilerTask) createServerRbacClusterRoleBinding(ctx context.Context,
...
+			{
+				Kind:      rbacv1.ServiceAccountKind,
+				Namespace: t.object.Namespace,
+				Name:      serverName,
+			},
+		},
+	}
+
+	if err := utils.CreateK8sCR(ctx, t.client, binding, t.object, utils.UPDATE); err != nil {
+		return fmt.Errorf("failed to create %s cluster role binding: %w", serverName, err)
```
irinamihai, line 1296: Maybe `ClusterRoleBinding` instead of `cluster role binding`?

PR #470, opened 2025-01-15
Resource server: add secrets to clusterrole and expose error

irinamihai:
Description:
- Add `secrets` to the resource-server clusterrole
- Expose errors coming from getting managed clusters

### DIFF
--- a/internal/controllers/inventory_controller.go
+++ b/internal/controllers/inventory_controller.go
@@ -986,6 +986,19 @@ func (t *reconcilerTask) createResourceServerClusterRole(ctx context.Context) er
 			},
 			{
 				APIGroups: []string{
+					"",
+				},
+				Resources: []string{
+					"secrets",
+				},
+				Verbs: []string{
+					"get",
+					"list",
+					"watch",
+				},
+			},
+			{
+				APIGroups: []string{
 					"internal.open-cluster-management.io",
 				},
 				Resources: []string{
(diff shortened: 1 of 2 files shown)


irinamihai:
/hold
Still looking into why this change is not needed on some setups.
The expected behavior is to need this permission added in order the obtain the kubeconfig of managed clusters through the deploymentManagers extensions.

internal/service/resources/collector/collector_acm.go
```diff
@@ -535,7 +535,7 @@ func (d *ACMDataSource) convertManagedClusterToDeploymentManager(ctx context.Con
 
 	extensions, err := d.getDeploymentManagerExtensions(ctx, cluster.Name)
 	if err != nil {
-		return models.DeploymentManager{}, fmt.Errorf("failed to get extensions for cluster %s", cluster.Name)
+		return models.DeploymentManager{}, fmt.Errorf("failed to get extensions for cluster %s: %s", cluster.Name, err.Error())
```
pixelsoccupied, line 538: consider using `"%w", err` for error to preserve the error chain and allow error inspection. Unless you specifically wanted this?

irinamihai:
This PR can now merge as the reason for the inconsistent behavior has been found.
https://github.com/stolostron/ocm/pull/157: this MCE fix has been backported, but the images have not yet been released for all the versions, which is why some environments were able to query secrets without the extra permission in the resource-server ClusterRole.

PR #473, opened 2025-01-16
Intercept and convert plain text errors to JSON

alegacy:
This implements a workaround for a limitation of the http.ServeMux to convert plain text error responses to JSON.  The limitation is that it does not provide a means to specify whether error responses should be in plain text or JSON format.  The issue is discussed here [1].

The 'solution' is to wrap the ServeMux so that we supply a custom ResponseWriter that is capable of intercepting "text/plain" errors and re-format them to "application/problem+json" to comply with the ORAN interface specifications.

The previous approach attempted to better handle this scenario for a 404 error using a default "/" path, but overlooked that 405 errors are never returned as long as the "/" path is in the mux.

[1] https://github.com/golang/go/issues/65648

### DIFF
--- a/internal/service/alarms/serve.go
+++ b/internal/service/alarms/serve.go
@@ -124,9 +124,7 @@ func Serve(config *api.AlarmsServerConfig) error {
 		},
 	)
 
-	r := http.NewServeMux()
-	// Register a default handler that replies with 404 so that we can override the response format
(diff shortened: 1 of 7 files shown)


alegacy:
FYI, will need to prop this over to the oran-hwmgr-plugin repo since we just finished adding an API for the inventory data over there.

PR #474, opened 2025-01-16
CNF-14864: cronjob to cleanup resolved alarms based on retention period

pixelsoccupied:
- manage cronjob and a configMap from alarms server to cleanup alarm event rows
- only retention period is used to create the cleanup sql (extension TBD)

### DIFF
--- a/internal/service/alarms/internal/serviceconfig/serviceconfig.go
+++ b/internal/service/alarms/internal/serviceconfig/serviceconfig.go
@@ -1,0 +1,232 @@
+package serviceconfig
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strconv"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
+	api "github.com/openshift-kni/oran-o2ims/internal/service/alarms/api/generated"
+	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/db/models"
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/clients/k8s"
+	serverutils "github.com/openshift-kni/oran-o2ims/internal/service/common/utils"
+	"github.com/stephenafamo/bob/dialect/psql"
+	"github.com/stephenafamo/bob/dialect/psql/dm"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	cleanupCronJobName   = "alarms-server-events-cleanup"
+	cleanupConfigMapName = "alarms-server-events-cleanup-sql"
+	cleanScriptDir       = "/scripts"
+	cleanScriptName      = "cleanup.sql"
+)
+
+var (
+	cronSchedule          = "0 * * * *" // run at the start of every hour e.g 2:00, 3:00 etc
+	jobBackoffLimit int32 = 3
+	successLimit    int32 = 3
+	failureLimit    int32 = 1
+	commonLabels          = map[string]string{
+		"app": utils.InventoryAlarmServerName,
+	}
+)
+
+// Config defines the configuration for serviceconfig
+type Config struct {
+	PostgresImage string        `envconfig:"POSTGRES_IMAGE" required:"true"` // PG image to use psql from
+	PodNamespace  string        `envconfig:"POD_NAMESPACE" required:"true"`  // Dynamically check the current ns
+	PgClient      *pgxpool.Pool // Postgres client to get current PG config
+	HubClient     client.Client // HubClient to manage cronjob resources
+}
+
+func LoadEnvConfig() (Config, error) {
+	var config Config
+	if err := envconfig.Process("", &config); err != nil {
+		return config, err //nolint:wrapcheck
+	}
+
+	return config, nil
+}
+
+// EnsureCleanupCronJob starts (or updates) a cronjob with all the required resources to do alarms events cleanup.
+func (c *Config) EnsureCleanupCronJob(ctx context.Context, sc *models.ServiceConfiguration) error {
+	// Get the deployment alarms deployment and take ownership of resources
+	deployment := &appsv1.Deployment{}
+	if err := c.HubClient.Get(ctx, client.ObjectKey{
+		Namespace: c.PodNamespace,
+		Name:      utils.InventoryAlarmServerName,
+	}, deployment); err != nil {
+		return fmt.Errorf("failed to get alarms-server deployment: %w", err)
+	}
+
+	// Create CM with sql
+	configMap, err := c.generateConfigMapWithSql(sc)
+	if err != nil {
+		return fmt.Errorf("generate sql configmap: %w", err)
+	}
+	if err := controllerutil.SetControllerReference(deployment, &configMap, c.HubClient.Scheme(),
+		controllerutil.WithBlockOwnerDeletion(false)); err != nil {
+		return fmt.Errorf("failed to set owner reference for configMap: %w", err)
+	}
+	if err := k8s.CreateOrUpdate(ctx, c.HubClient, &configMap); err != nil {
+		return fmt.Errorf("failed to apply configmap: %w", err)
+	}
+
+	// Create CronJob that reads the CM
+	cronJob := c.generateCronJob(configMap)
+	if err := controllerutil.SetControllerReference(deployment, &cronJob, c.HubClient.Scheme(),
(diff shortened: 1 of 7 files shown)


pixelsoccupied:
/hold

updating to use internal interface to create the sql query instead of directly providing the db tags

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+	cleanScriptName      = "cleanup.sql"
+)
+
+var (
+	cronSchedule    = "0 * * * *" // run at the start of every hour e.g 2:00, 3:00 etc
+	jobBackoffLimit = int32(3)
+	successLimit    = int32(3)
+	failureLimit    = int32(1)
+	commonLabels    = map[string]string{
+		"app": "alarms-server",
```
alegacy, line 42: We should use our server name constants from utils.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+func generateConfigMapWithSql(sc *models.ServiceConfiguration) (corev1.ConfigMap, error) {
+	sql, err := getCleanUpPgSQL(sc)
+	if err != nil {
+		return corev1.ConfigMap{}, fmt.Errorf("failed to generate cleanup sql for configmap: %w", err)
+	}
+
+	return corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cleanupConfigMapName,
+			Namespace: utils.DefaultNamespace,
```
alegacy, line 86: We should try to get our namespace from our env.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+						},
+					},
+				},
+			},
+		},
+	}, nil
+}
+
+// getPostgresImage dynamically collect the existing PG image to run psql from
+func getPostgresImage(ctx context.Context, c client.Client) (string, error) {
```
alegacy, line 214: We have this in an env variable in the operator.  We should consider simply passing it to the alarms-server via its env.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+	return "", fmt.Errorf("server container not found in postgres deployment %s", deployment.Name)
+}
+
+// CleanupResources call this for cleanup of k8s resources
+func CleanupResources(ctx context.Context, hc client.Client) error {
+	// Delete the cronjob and wait for deletion
+	cronJob := &batchv1.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cleanupCronJobName,
+			Namespace: utils.DefaultNamespace,
```
alegacy, line 244: Same comment as earlier... should consider getting NS from env.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+		if err != nil {
+			return false, err //nolint:wrapcheck
+		}
+		return false, nil
+	}); err != nil {
+		return fmt.Errorf("failed waiting for cronjob deletion: %w", err)
+	}
+
+	// Delete the configmap and wait for deletion
+	configMap := &corev1.ConfigMap{
```
alegacy, line 270: Should we set the ownership of these new objects so that they can auto-deleted when the alarm-server is deleted or when the operator is deleted?

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+			return container.Image, nil
+		}
+	}
+
+	return "", fmt.Errorf("server container not found in postgres deployment %s", deployment.Name)
+}
+
+// CleanupResources call this for cleanup of k8s resources
+func CleanupResources(ctx context.Context, hc client.Client) error {
```
alegacy, line 239: Do we need to "cleanup" these resources or can we just do an "upsert" type operation on it since we seem to be immediately recreating them in all cases.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,300 @@
...
+			},
+		},
+	}, nil
+}
+
+// getPostgresImage dynamically collect the existing PG image to run psql from
+func getPostgresImage(ctx context.Context, c client.Client) (string, error) {
+	var deploymentList appsv1.DeploymentList
+	if err := c.List(ctx, &deploymentList, client.InNamespace(utils.DefaultNamespace),
+		client.MatchingLabels{"app": "postgres-server"}); err != nil {
```
alegacy, line 217: We should use our constants instead of raw "postgres-server"

alegacy (commented):
Mostly minor stuff, but overall looks good!

pixelsoccupied:
/hold 

making code review updates

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: Yeah mostly it was to force cronjob to quickly read the update configmap. But thinking about it again now that you mention it....if we do update the configmap...all the subsequent pods will read the latest cm anyway...argh yeah will make this cleaner.

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: Ah yeah good point this can be improved

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: using constant

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: using env

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: using env

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: using env

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: took ownership

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: cleaned it up and dropped the delete step

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: using env and removed this

internal/service/alarms/internal/serviceconfig/serviceconfig.go
```diff
@@ -0,0 +1,235 @@
...
+	if err := envconfig.Process("", &config); err != nil {
+		return config, err //nolint:wrapcheck
+	}
+
+	return config, nil
+}
+
+// EnsureCleanupCronJob starts (or updates) a cronjob with all the required resources to do alarms events cleanup.
+func EnsureCleanupCronJob(ctx context.Context, sc *models.ServiceConfiguration, pg *pgxpool.Pool, hc client.Client) error {
+	config, err := LoadConfig()
```
alegacy, line 63: Do we want to do this every time or should we consider moving this `AlarmSeverConfig` and doing something similar to `config.LoadFromEnv()` which is run on startup?

internal/service/common/clients/k8s/k8s.go
```diff
@@ -110,3 +117,29 @@ func GetSchemeForCluster() *runtime.Scheme {
...
+			slog.Info("Creating a new resource", "gkv", obj.GetObjectKind().GroupVersionKind().String(),
+				"namespace", obj.GetNamespace(), "name", obj.GetName())
+			return c.Create(ctx, obj) //nolint:wrapcheck
+		}
+		return fmt.Errorf("failed to get existing object: %w", err)
+	}
+
+	// Update existing object
+	obj.SetResourceVersion(existing.GetResourceVersion())
+	slog.Info("Updating an existing resource", "gkv", obj.GetObjectKind().GroupVersionKind().String(),
```
alegacy, line 142: typo: `gkv` -> `gvk`

internal/service/alarms/internal/serviceconfig/serviceconfig.go
pixelsoccupied: Yeah started looking into using AlarmSeverConfig at first...but ran into the circular import. Let me see if i can move it around a bit and pull this one time!

pixelsoccupied: should be done one time now at start now. Reusing AlarmSeverConfig required a bit too much refactoring so using unique struct still. Let me know how it looks!

internal/service/common/clients/k8s/k8s.go
pixelsoccupied: updated

PR #482, opened 2025-01-18
Define and package function-specific RBAC roles

alegacy:
Defines function-specific RBAC roles for users of the O2IMS API and CRs. These roles can later be assigned to users or groups via ClusterRoleBindings.

The roles are defined for the following functions:

O2IMS API access:

+ reader: read-only access to all API endpoints
+ subscriber: read/write access to all subscription API endpoints
+ maintainer: read/write access to alarms and alarm configuration
+ provisioner: read/write access to templates and provisioning requests
+ admin: read/write access to all API endpoints

O2IMS CR access:
(note: these are mostly for developer testing since similar ones get
 automatically created when our operator bundle is installed and run.)

+ editor: read/write access to corresponding CR
+ viewer: read-only access to corresponding CR

### DIFF
--- a/config/rbac/o2ims.oran.openshift.io_inventories_editor_role.yaml
+++ b/config/rbac/o2ims.oran.openshift.io_inventories_editor_role.yaml
@@ -1,20 +1,20 @@
-# permissions for end users to edit provisioningrequests.
+# permissions for end users to edit Inventory CR.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/name: clusterrole
-    app.kubernetes.io/instance: provisioningrequest-editor-role
+    app.kubernetes.io/instance: inventory-editor-role
     app.kubernetes.io/component: rbac
     app.kubernetes.io/created-by: oran-o2ims
     app.kubernetes.io/part-of: oran-o2ims
     app.kubernetes.io/managed-by: kustomize
-  name: provisioningrequest-editor-role
+  name: inventory-editor-role
 rules:
 - apiGroups:
-  - oran.openshift.io
+  - o2ims.oran.openshift.io
   resources:
-  - provisioningrequests
+  - inventories
   verbs:
   - create
   - delete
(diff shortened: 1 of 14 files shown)


config/rbac/o2ims.oran.openshift.io_inventories_editor_role.yaml
```diff
@@ -1,20 +1,20 @@
-# permissions for end users to edit orano2ims.
+# permissions for end users to edit Inventory CR.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/name: clusterrole
-    app.kubernetes.io/instance: orano2ims-editor-role
+    app.kubernetes.io/instance: oran-o2ims-installer-role
```
mlguerrero12, line 7: shouldn't be inventories-editor-role to be consistent with the others?

config/rbac/o2ims.oran.openshift.io_inventories_viewer_role.yaml
```diff
@@ -1,27 +1,27 @@
-# permissions for end users to view orano2ims.
+# permissions for end users to view Inventory CR.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/name: clusterrole
-    app.kubernetes.io/instance: orano2ims-viewer-role
+    app.kubernetes.io/instance: oran-o2ims-viewer-role
```
mlguerrero12, line 7: inventories-viewer-role?

config/rbac/o2ims.oran.openshift.io_inventories_editor_role.yaml
alegacy: yes, that's true!

alegacy: actually...the others are using the singular form so this will be "inventory-editor-role"

alegacy: fixed

PR #487, opened 2025-01-20
CNF-16478: Add OIDC validation at front-end proxies

alegacy:
This configures the front-end kube-rbac-proxy instances to accept OIDC OAuth tokens rather than Kubernetes tokens.  Incoming tokens are verified by first validating their signature against the certificate downloaded from the configured OAuth server.  If the signature is
 valid, then the user's roles are extracted from the token and used to
map to the Kubernetes RBAC roles to authorize the request.

Since our servers do inter-server communication and since they do not have OAuth tokens to exchange between themselves (not should they) we create separate kube-rbac-proxy instances that use the Kubernetes token validation to authorize their exchanges.

Therefore, any internal communications go through the "internal" proxy, and any public communications (e.g,. coming from the ingress controller) go through the "public" proxy.

### DIFF
--- a/internal/controllers/inventory_controller.go
+++ b/internal/controllers/inventory_controller.go
@@ -128,6 +128,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (resul
 	return
 }
 
+// getInternalServicePorts calculates the port values for the internal proxy depending on whether OAuth is enabled
+func (t *reconcilerTask) getInternalServicePorts() (int32, string) {
+	internalPort := int32(0)
+	internalTargetPort := ""
+	if utils.IsOAuthEnabled(t.object) {
+		// Configure the service to point to the internal RBAC proxy which uses regular Kubernetes authentication for
+		// service to service communication.
+		internalPort = utils.InternalServicePort
+		internalTargetPort = utils.InternalServiceTargetPort
+	}
+
+	return internalPort, internalTargetPort
+}
+
 // setupResourceServerConfig creates the resources necessary to start the Resource Server.
 func (t *reconcilerTask) setupResourceServerConfig(ctx context.Context, defaultResult ctrl.Result) (nextReconcile ctrl.Result, err error) {
 	nextReconcile = defaultResult
@@ -174,16 +188,8 @@ func (t *reconcilerTask) setupResourceServerConfig(ctx context.Context, defaultR
 		return
 	}
 
-	internalPort := int32(0)
-	internalTargetPort := ""
-	if utils.IsOAuthEnabled(t.object) {
-		// Configure the service to point to the internal RBAC proxy which uses regular Kubernetes authentication for
-		// service to service communication.
-		internalPort = utils.InternalServicePort
-		internalTargetPort = utils.InternalServiceTargetPort
-	}
-
 	// Create the Service needed for the Resource server.
+	internalPort, internalTargetPort := t.getInternalServicePorts()
 	err = t.createService(ctx, utils.InventoryResourceServerName, utils.DefaultServicePort, utils.DefaultServiceTargetPort, internalPort, internalTargetPort)
 	if err != nil {
 		t.logger.ErrorContext(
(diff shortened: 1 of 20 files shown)


internal/controllers/inventory_controller.go
```diff
@@ -174,8 +174,17 @@ func (t *reconcilerTask) setupResourceServerConfig(ctx context.Context, defaultR
 		return
 	}
 
+	internalPort := int32(0)
```
irinamihai, line 177: We could consider moving these to their own function since it's used in 3 places in this file.

internal/controllers/inventory_controller.go
```diff
@@ -1541,6 +1600,47 @@ func (t *reconcilerTask) deployServer(ctx context.Context, serverName string) (u
...
+			},
+			Image:           os.Getenv(utils.KubeRbacProxyImageName),
+			ImagePullPolicy: corev1.PullPolicy(os.Getenv(utils.ImagePullPolicyEnvName)),
+			Args: []string{
+				fmt.Sprintf("--secure-listen-address=0.0.0.0:%d", utils.InternalProxyPort),
+				fmt.Sprintf("--upstream=http://127.0.0.1:%d/", utils.DefaultContainerPort),
+				"--logtostderr=true",
+				"--tls-cert-file=/secrets/tls/tls.crt",
+				"--tls-private-key-file=/secrets/tls/tls.key",
+				"--tls-min-version=VersionTLS12",
```
irinamihai, line 1620: Would it make sense to have `VersionTLS12` a constant since it's used in multiple places? Mostly for the case when we'll have to move to the next version.
Maybe even the `10` below?

internal/controllers/inventory_controller.go
```diff
@@ -1541,6 +1600,47 @@ func (t *reconcilerTask) deployServer(ctx context.Context, serverName string) (u
...
+				"--v=10",
+			},
+			Ports: []corev1.ContainerPort{
+				{
+					Name:          utils.InternalServiceTargetPort,
+					Protocol:      corev1.ProtocolTCP,
+					ContainerPort: utils.InternalProxyPort,
+				},
+			},
+			Resources: corev1.ResourceRequirements{
```
irinamihai, line 1630: Any chance we could add a small comment to explain why we have these exact values?

internal/controllers/inventory_controller.go
```diff
@@ -1439,16 +1465,55 @@ func (t *reconcilerTask) deployServer(ctx context.Context, serverName string) (u
...
+		fmt.Sprintf("--secure-listen-address=0.0.0.0:%d", utils.DefaultProxyPort),
+		fmt.Sprintf("--upstream=http://127.0.0.1:%d/", utils.DefaultContainerPort),
+		"--logtostderr=true",
+		"--tls-cert-file=/secrets/tls/tls.crt",
+		"--tls-private-key-file=/secrets/tls/tls.key",
+		"--tls-min-version=VersionTLS12",
+		"--v=10",
+	}
+
+	if utils.IsOAuthEnabled(t.object) {
```
irinamihai, line 1478: Could we move this in its own function as well? This `deployServer` is becoming quite large.

internal/controllers/inventory_controller.go
```diff
@@ -1541,6 +1600,47 @@ func (t *reconcilerTask) deployServer(ctx context.Context, serverName string) (u
 		},
 	}
 
+	if utils.IsOAuthEnabled(t.object) && utils.RequiresInternalProxy(serverName) {
```
irinamihai, line 1603: Same comment about having this in its own function.

internal/controllers/inventory_controller.go
alegacy: ok ... i'll add a note to say that they are copied from the scaffolding

internal/controllers/inventory_controller.go
alegacy: fixed

internal/controllers/inventory_controller.go
alegacy: fixed

internal/controllers/inventory_controller.go

internal/controllers/inventory_controller.go

internal/controllers/inventory_controller.go

alegacy:
... pushed again to fix linter 😢

PR #489, opened 2025-01-21
Delay SMO registration until all servers are ready

alegacy:
Registering with the SMO implies that we are ready to accept requests. That isn't true until all Pods are reporting a Ready status; therefore, we should wait before registering with the SMO.

### DIFF
--- a/internal/controllers/inventory_controller.go
+++ b/internal/controllers/inventory_controller.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	k8sptr "k8s.io/utils/ptr"
+	"k8s.io/utils/strings/slices"
 
 	"github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 
@@ -68,6 +69,7 @@ import (
 //+kubebuilder:rbac:groups="cluster.open-cluster-management.io",resources=managedclusters,verbs=get;list;watch
 //+kubebuilder:rbac:groups="",resources=secrets,verbs=get;delete;list;watch;update
 //+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 //+kubebuilder:rbac:groups="internal.open-cluster-management.io",resources=managedclusterinfos,verbs=get;list;watch
 //+kubebuilder:rbac:groups="config.openshift.io",resources=clusterversions,verbs=get;list;watch
 //+kubebuilder:rbac:urls="/internal/v1/caas-alerts/alertmanager",verbs=create;post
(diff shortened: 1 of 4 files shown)


internal/controllers/inventory_controller.go
```diff
@@ -721,6 +723,44 @@ func (t *reconcilerTask) storeSearchURL(ctx context.Context) error {
 	return nil
 }
 
+// waitForDeployments waits for all server pods to be ready.  If any pod is not yet ready, the reconciler timer is set
+// to a short value so that we can try again quickly; otherwise either an error is returned without a reconciler timer
+// to indicate that an exponential backoff is required
+func (t *reconcilerTask) waitForPods(ctx context.Context) (nextReconcile ctrl.Result, err error) {
```
mlguerrero12, line 729: You are not waiting here, so I would change the name to CheckForPodsReady and simply return a boolean, error. Outside, you can check if the boolean is false and set the nextReconciler and return. I think it's cleaner. 

Having said the above, I believe it is more efficient to actually wait and periodically check. I've done this using a function provided by k8s. If timeout expires, you could retry the reconciliation process.

https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/poll.go#L45

Here, for example,

https://github.com/openshift/cluster-network-operator/blob/a967bf84a777795ff47f199c5200341809d250b1/pkg/controller/allowlist/allowlist_controller.go#L191

mlguerrero12: Here you are not completely idle and you have control for how long you wait. You can set the timeout to 1 minute and check every second as in the example. I've seen this in many places. The example is in the cluster-network-operator which is a core payload operator in OCP. So, It'd say it's fine.

pixelsoccupied: Although polling is great in some cases e.g to making sure pods are deleted fully...in this case however i think a non-blocking call is the right approach. Waiting for pods to be created maybe super long (e.g maybe a pod is in forever crashloop) and we wont know about it until a arbitrary timeout out is reached. 

Yeah as rule of thumb controller we should always prefer reconcile...and any long running one shot thing should be transferred to a Job resource (which is then monitored with reconcile loop)

alegacy: In the example given, doing a Get every 1 second for 1 minute seems a little harsh on the API server.   I'd prefer slowing that down to at least 30 seconds, and at that point if we are waiting for 30 seconds then our reconciler is potentially missing user requests to change the Inventory CR (yes, I know that is highly unlikely).   

I tend to agree with @pixelsoccupied here... reconciling is probably better given that we'd be polling slowly.   

I've changed the name to reflect the true nature of the method.

mlguerrero12: Retrying with a timeout is also blocking. In this case, 30 seconds. The mentioned k8s function has a timeout and interval to check the condition.

mlguerrero12: That was meant for detecting a pod in a forever crashloop. True about the user trying to change something. It will have to wait 30s in an unlikely scenario. Not critical I'd say. And, regarding the api server, my understanding is that the moment you list a resource with the client of the operator runtime, a cache is put in place. So, it doesn't go hard on the api server.

Anyway, this is not a critical part. Everything is fine.
Just a nit  in the comment  checkForPodReadyStatus waits  checkForPodReadyStatus checks

alegacy: comment updated

internal/controllers/inventory_controller.go
```diff
@@ -721,6 +723,44 @@ func (t *reconcilerTask) storeSearchURL(ctx context.Context) error {
 	return nil
 }
 
+// checkForPodReadyStatus checks for all server pods to be ready.  If any pod is not yet ready, the reconciler timer is
+// set to a short value so that we can try again quickly; otherwise either an error is returned without a reconciler
+// timer to indicate that an exponential backoff is required
+func (t *reconcilerTask) checkForPodReadyStatus(ctx context.Context) (nextReconcile ctrl.Result, err error) {
```
pixelsoccupied, line 729: Consider updating ` (nextReconcile ctrl.Result, err error) {` to `(ctrl.Result, error) {`....like we are returning explicitly so might as well (i think you simply need `err := t.client.` in L738.) 

Go's "naked" return func is probably my biggest gripe with language (and widely considered a practice)! Like I find almost impossible to read when there's "return" but nothing specified especially when they go over 3 lines lol and unfortunately we have a quite bit of it in the controllers 😖

alegacy: ya, i agree... it makes it really tough to read!

PR #493, opened 2025-01-21
CNF-14956: remove all deprecated alarms server code from previous design

pixelsoccupied:
With the newly designed and implemented alarms-server we can now safely remove the deprecated version


pixelsoccupied:
ah forgot to run ci-jobs - had to update the test cases too

PR #499, opened 2025-01-23
Use dynamically loaded certificates/tokens

alegacy:
*    Use dynamically loaded service account tokens
    
    This change ensures that we are periodically re-reading our service
    account tokens so that any long-lived HTTP clients do not start failing
    when our tokens expired, therefore requiring us to restart the Pod.
    
*    Use dynamically loaded client certificates
    
    To support use cases where certificates are renewed periodically without
    needing to restart our servers we use dynamically loaded content
    providers for the key pairs.


internal/controllers/utils/oauth.go
```diff
@@ -11,12 +11,41 @@ import (
...
+
+// AddListener is a no-op for static providers.
+func (s *StaticKeyPairLoader) AddListener(_ dynamiccertificates.Listener) {
+	// NOOP
+}
+
+// Name is a simple name to identify this provider
+func (s *StaticKeyPairLoader) Name() string {
+	return "StaticKeyPairLoader"
+}
```
pixelsoccupied, line 33: Just curious....why have a `StaticKeyPairLoader` instead of using  `dynamiccertificates.NewDynamicServingContentFromFiles("StaticKeyPairLoader"...` (just like client-oauth) ? I guess we don't really have a path which the NewDynamicServingContentFromFiles expects?

PR #502, opened 2025-01-23
Add ClusterTemplate ID as label in MCL and Agent CRs

irinamihai:
Description:
- For a ManagedCluster and the corresponding Agent, add the `clustertemplates.o2ims.provisioning.oran.org/templateId` label with a value that points to `spec.TemplateId` of the ClusterTemplate used in the corresponding ProvisioningRequest.
- The above label is only added once a ProvisioningRequest becomes `fulfilled`.
- If the ProvisioningRequest goes through Day2 configuration and its status changes from `fulfilled`, the label is kept.
- The function that adds the labels is run during each reconciliation for which the ProvisioningRequest is `fulfilled` to ensure the label has the correct value, in the case it has been tempered with.
- Add tests.

### DIFF
--- a/internal/controllers/provisioningrequest_clusterconfig.go
+++ b/internal/controllers/provisioningrequest_clusterconfig.go
@@ -12,6 +12,7 @@ import (
 
 	provisioningv1alpha1 "github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1"
 	"github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
+	assistedservicev1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 )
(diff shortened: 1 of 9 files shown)


internal/controllers/provisioningrequest_clusterconfig.go
```diff
@@ -248,6 +256,72 @@ func (t *provisioningRequestReconcilerTask) finalizeProvisioningIfComplete(ctx c
...
+		client.InNamespace(mcl.Name),
+	}
+
+	err = t.client.List(ctx, agents, listOpts...)
+	if err != nil {
+		return fmt.Errorf("failed to list Agents in the %s namespace: %w", mcl.Name, err)
+	}
+
+	if len(agents.Items) > 1 {
+		return fmt.Errorf("more than one Agent in the %s namespace", mcl.Name)
```
alegacy, line 291: I may be misreading this, but isn't `> 1` a possibility in a MNO setup?  In which case we'd want to label all of them?

alegacy:
I added a question... but I'm assuming the MNO case (if I am correct about the question) isn't imminent so we can address that in a separate PR.

internal/controllers/provisioningrequest_clusterconfig.go
browsell: each agent (node instance) would be in its own NS I believe

Missxiaoguo: What I have seen before is that Agent for each node is under the cluster NS. Same for sno + worker.

irinamihai: I remember seeing the same on 3 node spoke clusters. I will check in the QE labs for 3 node and standard spoke clusters and I'll open a following PR to properly address those scenarios.

PR #503, opened 2025-01-24
Move alarm dictionary collection logic into cluster server

mlguerrero12:
Moving the alarm dictionary collection from the alert server to the cluster server avoid the need to transfer and cache the dictionaries in the cluster server. Alarm dictionaries are relevant to the infrastructure servers, not the alert server


mlguerrero12:
@alegacy , @pixelsoccupied PTAL

mlguerrero12:
rebased

internal/service/cluster/db/repo/repository.go
```diff
@@ -126,3 +128,64 @@ func (r *ClusterRepository) GetClusterResourcesNotIn(ctx context.Context, keys [
 func (r *ClusterRepository) GetClusterResource(ctx context.Context, id uuid.UUID) (*models.ClusterResource, error) {
 	return utils.Find[models.ClusterResource](ctx, r.Db, id)
 }
+
+// UpsertAlarmDefinitions inserts or updates alarm definition records
+func (r *ClusterRepository) UpsertAlarmDefinitions(ctx context.Context, records []models.AlarmDefinition) ([]models.AlarmDefinition, error) {
```
alegacy, line 133: As discussed, this is ok for now, but eventually these alarm dictionary objects will be subject to us having to publish notification events when they are added/updated/deleted so we will need to use the `Persist*` functions that were added to handle writing to the DB and capturing the data change event for notification purposes.

internal/service/cluster/db/models/alarm_definition.go
```diff
@@ -0,0 +1,40 @@
+package models
```
alegacy, line 1: Given that the API definitions are now in the `common` package perhaps we should move these to the `common/db/models` package?

internal/service/cluster/db/models/alarm_dictionary.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// PrimaryKey returns the primary key of the table
+func (r AlarmDictionary) PrimaryKey() string {
+	return "alarm_dictionary_id"
+}
+
+// OnConflict returns the column or constraint to be used in the UPSERT operation
+func (r AlarmDictionary) OnConflict() string {
+	return "node_cluster_type_id"
```
alegacy, line 37: I'm unclear on how this is meant to be used.   Maybe a comment explaining why this field in particular is the one selected... i assumed it would be the primary key field.

internal/service/cluster/db/migrations/000001_baseline.up.sql
```diff
@@ -101,35 +101,44 @@ CREATE TABLE IF NOT EXISTS cluster_resource
...
+    -- By adding this columns and a unique constraint on (alarm_name, severity), we can differentiate between them.
+    -- All the Alerts from the Core Platform Monitoring have a severity label (except alert Watchdog). Alerts without a severity label are not affected by this.
+    severity                VARCHAR(20)   NOT NULL,
+    alarm_dictionary_id     UUID          NOT NULL,
     created_at              TIMESTAMPTZ   DEFAULT CURRENT_TIMESTAMP,
-    FOREIGN KEY (alarm_dictionary_id) REFERENCES cached_alarm_dictionary (alarm_dictionary_id),
-    CONSTRAINT unique_alarm_name_last_change UNIQUE (alarm_name, alarm_last_change)
+
+    FOREIGN KEY (alarm_dictionary_id) REFERENCES alarm_dictionary (alarm_dictionary_id) ON DELETE CASCADE,
+    CONSTRAINT unique_alarm UNIQUE(alarm_dictionary_id, alarm_name, severity)
```
alegacy, line 143: Just double-checking ... There can be 2 alarms with the same name, but with different severities on the same dictionary?

internal/service/cluster/db/migrations/000001_baseline.up.sql
```diff
@@ -101,35 +101,44 @@ CREATE TABLE IF NOT EXISTS cluster_resource
...
+    alarm_definition_id     UUID          PRIMARY KEY DEFAULT gen_random_uuid(),
+    alarm_name              VARCHAR(255)  NOT NULL,
+    alarm_last_change       VARCHAR(50)   NOT NULL,
+    alarm_change_type       VARCHAR(20)   NOT NULL,
+    alarm_description       TEXT          NOT NULL,
+    proposed_repair_actions TEXT          NOT NULL,
+    clearing_type           VARCHAR(20)   NOT NULL,
     management_interface_id VARCHAR(50)[] DEFAULT ARRAY ['O2IMS']::VARCHAR[],
-    pk_notification_field   TEXT[]        DEFAULT ARRAY ['alarm_definition_id']::TEXT[],
+    pk_notification_field   TEXT[]        DEFAULT ARRAY ['alarm_dictionary_id']::TEXT[],
```
alegacy, line 132: Shouldn't this be `alarm_definitioni_id` as it is meant to indicate which field in an event points to this record?

internal/service/cluster/collector/collector_alarms.go
```diff
@@ -0,0 +1,445 @@
...
+	hubClient    client.WithWatch
+	generationID int
+}
+
+// NewAlarmsDataSource creates a new AlarmsDataSource
+func NewAlarmsDataSource() (DataSource, error) {
+	// To avoid log from eventuallyFulfillRoot controller-runtime
+	log.SetLogger(logr.Discard())
+
+	hubClient, err := k8s.NewClientForHub()
```
alegacy, line 39: Ok for now, but since the `hubClient` is going to be long-lived there is a possibility that becomes stale as its config gets changed and will need to be refreshed.   We'll have to handle that scenario dynamically without requiring a process restart.

internal/service/cluster/collector/collector_alarms.go
```diff
@@ -0,0 +1,445 @@
...
+
+		alarmDictionaryIDString := (*nodeClusterType.Extensions)[utils.ClusterAlarmDictionaryIDExtension]
+		if alarmDictionaryIDString != nil {
+			alarmDictionaryID, err := uuid.Parse(alarmDictionaryIDString.(string))
+			if err != nil {
+				slog.Error("error parsing alarm dictionary ID", "NodeClusterType ID", nodeClusterType.NodeClusterTypeID, "error", err)
+				continue
+			}
+
+			(*nodeClusterType.Extensions)[utils.ClusterAlarmDictionaryIDExtension] = alarmDictionaryID
```
alegacy, line 108: Not clear on why this is being written back into the extensions if it was already there.

internal/service/cluster/collector/collector_alarms.go
```diff
@@ -0,0 +1,445 @@
...
+			PKNotificationField:    []string{"alarm_dictionary_id"},
+			NodeClusterTypeID:      nodeClusterType.NodeClusterTypeID,
+			DataSourceID:           d.GetID(),
+			GenerationID:           d.GetGenerationID(),
+		}
+
+		alarmDictionaries = append(alarmDictionaries, alarmDictionary)
+	}
+
+	return alarmDictionaries
```
alegacy, line 409: This will return `nil` if there were no `nodeClusterTypes`.  I think that was filtered out before calling this, but if not, would returning an empty slice be better/cleaner?

alegacy (commented):
Mostly minor comments.  Otherwise looks good!

internal/service/cluster/db/migrations/000001_baseline.up.sql
mlguerrero12: Yes, good catch!

internal/service/cluster/db/migrations/000001_baseline.up.sql
mlguerrero12: yes

internal/service/cluster/db/models/alarm_definition.go
mlguerrero12: yes, it makes sense

internal/service/cluster/db/models/alarm_dictionary.go
mlguerrero12: This one (alarm dictionary) is not being used anymore. It was used in an upsert operation but now I'm using PersistWithObjectChange and DeleteWithObjectChange to report any changes. I will leave it empty and wrote a comment in the one for alarm definitions which is still being used.

internal/service/cluster/collector/collector_alarms.go
mlguerrero12: Indeed, will write a TODO

internal/service/cluster/collector/collector_alarms.go
mlguerrero12: I don't know either. It seems a left over from a previous implementation. Will delete it. Thanks

internal/service/cluster/collector/collector_alarms.go
mlguerrero12: No, I think nil is just fine

internal/service/cluster/db/models/alarm_definition.go
mlguerrero12: done

internal/service/cluster/db/migrations/000001_baseline.up.sql
mlguerrero12: fixed

internal/service/cluster/collector/collector_alarms.go
mlguerrero12: done

internal/service/cluster/collector/collector_alarms.go
mlguerrero12: fixed

PR #515, opened 2025-01-27
Require cluster template labels on clusters and agents

alegacy:
This change adds a validation step on events received for Managed
Clusters and Agents.  If they do not have a condition that indicates
they are ready, or they do not have the cluster template id label that
indicates the provisioning request has been fulfilled, then we simply
ignore them until they do.

Because Agents may complete their installation before the Managed
Cluster is fully provisioned, it is possible for ClusterResources to
appear before the NodeCluster.  Since the NodeCluster needs to track
which ClusterResources belong to it, we need to ensure the DB
relationship between them is updated regardless of the order or
arrival.


internal/service/cluster/db/repo/repository.go
```diff
@@ -190,3 +190,15 @@ func (r *ClusterRepository) FindStaleAlarmDictionaries(ctx context.Context, data
 	e := psql.Quote("data_source_id").EQ(psql.Arg(dataSourceID)).And(psql.Quote("generation_id").LT(psql.Arg(generationID)))
 	return utils.Search[commonmodels.AlarmDictionary](ctx, r.Db, e)
 }
+
+// SetNodeClusterID sets the nodeClusterID value on cluster resources that may have arrived out of order
+func (r *ClusterRepository) SetNodeClusterID(ctx context.Context, nodeClusterName string, nodeClusterID uuid.UUID) (int, error) {
+	updates := models.ClusterResource{NodeClusterID: &nodeClusterID}
+	e := psql.Quote("node_cluster_name").EQ(psql.Arg(nodeClusterName)).And(psql.Quote("node_cluster_id").IsNull())
```
pixelsoccupied, line 197: should we be using the internal stuff to extract out the DB fields e.g utils.GetAllDBTagsFromStruct? Though i noticed in common we are also doing GetColumnsAndValues. In general not sure tbh...like doing reflections a few times just get the tags 🤔 (though i dont think performance is an issue here...but rather if idiomatic)

pixelsoccupied (commented):
lgtm

internal/service/cluster/db/repo/repository.go
alegacy: We use the utilities to get the names when we need the full list, but when we need 1 or 2 (other than the primary key) we specify them by name.

internal/service/cluster/collector/collector.go
```diff
@@ -546,7 +535,16 @@ func (c *Collector) handleAsyncClusterResourceEvent(ctx context.Context, dataSou
 		return nil
 	}
 
-	err := c.persistClusterResource(ctx, dataSource, clusterResource, map[uuid.UUID]bool{})
+	nodeCluster, err := c.repository.GetNodeClusterByName(ctx, clusterResource.NodeClusterName)
+	if errors.Is(err, utils.ErrNotFound) {
+		// Agents will finish being installed before the Managed Cluster is completely provisioned therefore we have to
+		// link them after the fact so here we just skip them.
+		slog.Warn("no node cluster found", "name", clusterResource.NodeClusterName, "resource", clusterResource.Name)
```
mlguerrero12, line 542: Just to double check... this will be a row with a foreign key equal to NULL, right?

PR #523, opened 2025-01-28
Disable http2 for the IMS Deployment and webhook

irinamihai:


alegacy:
We will have to change the kube-rbac-proxy configs generated by the inventory_controller.go when launching the individual services. ... this includes both the default proxy, and the new one added for OAuth/OIDC.

irinamihai:
/unhold
I realized that we need `--http2-disable=true` as its default is `false` ([link](https://github.com/brancz/kube-rbac-proxy/blob/af1a90b60796a6d6acb3673bd58336e1a6319c2b/cmd/kube-rbac-proxy/app/options/options.go#L134)).

PR #528, opened 2025-01-29
Code & Inventory CR cleanup

alegacy:
*    Cleanup unused fields from Inventory CRD
    
    This removes the server "enabled" flags and other unused fields from
    the Inventory CRD.
    
*    Cleanup dead code
    
    Deletes code that was part of the original implementation of the
    resource, alarm, and metadata servers.


pixelsoccupied:
deep clean! 

/lgtm

PR #531, opened 2025-01-31
Fetch alarm dictionaries from cluster server

mlguerrero12:
The alarm server uses the alarm dictionary endpoints of the cluster server to fetch the alarm dictionaries. These dictionaries are cached in the same way as the nodeClusters and nodeClusterTypes objects.

There are 2 new methods in the infrastructure client to return the nodeClusterTypeID and the alarmDefinitionID. These methods return the IDs if they are present in cache or try to fetch them from the cluster server if they are missing.

There is also a resync mechanism that runs every hour.


mlguerrero12:
@alegacy, @pixelsoccupied, PTAL

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -110,11 +100,27 @@ func ConvertAmToAlarmEventRecordModels(am *api.AlertmanagerNotification, aDefini
...
-			if def.AlarmName == GetAlertName(*alert.Labels) && def.ObjectTypeID == *record.ObjectTypeID && severityToPerceivedSeverity(def.Severity) == record.PerceivedSeverity {
-				record.AlarmDefinitionID = &def.AlarmDefinitionID
-				record.ProbableCauseID = &def.ProbableCauseID
+		if record.ObjectTypeID != nil {
+			_, severity := GetPerceivedSeverity(*alert.Labels)
+			alarmDefinitionID, err := infrastructureClient.GetAlarmDefinitionID(*record.ObjectTypeID, GetAlertName(*alert.Labels), severity)
+			if err != nil {
+				slog.Warn("Could not get alarm definition ID", "objectTypeID", *record.ObjectTypeID, "name", GetAlertName(*alert.Labels), "severity", severity, "err", err.Error())
+			} else {
+				record.AlarmDefinitionID = &alarmDefinitionID
```
pixelsoccupied, line 123: given that we are intentionally not dealing with probableCauseID until further notice...what do you think about dropping it from spec? Maybe another PR can do that work. Currently i'm assuming everything works without any issues so not a big...but essentially its all dead code right?

pixelsoccupied (commented):
lgtm!

internal/service/alarms/internal/alertmanager/alertmanager.go
```diff
@@ -110,11 +100,27 @@ func ConvertAmToAlarmEventRecordModels(am *api.AlertmanagerNotification, aDefini
...
+		}
+
 		// See if possible to pick up additional info from its definition
-		for _, def := range aDefinitionRecords {
-			if def.AlarmName == GetAlertName(*alert.Labels) && def.ObjectTypeID == *record.ObjectTypeID && severityToPerceivedSeverity(def.Severity) == record.PerceivedSeverity {
-				record.AlarmDefinitionID = &def.AlarmDefinitionID
-				record.ProbableCauseID = &def.ProbableCauseID
+		if record.ObjectTypeID != nil {
+			_, severity := GetPerceivedSeverity(*alert.Labels)
+			alarmDefinitionID, err := infrastructureClient.GetAlarmDefinitionID(*record.ObjectTypeID, GetAlertName(*alert.Labels), severity)
```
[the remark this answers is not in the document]
mlguerrero12: I'm not caching the whole alarm definition but only the IDs. There is map whose key is a struct with the name and severity of the definitions and the value the ID. At the moment, we don´t need anything else.

Regarding probableCauseID, indeed we will have to add it. However, it is not clear how this will be done. There will be a list of ProbableCauses in the dictionary, similar to the list of Definitions. WG10-Information models and Data models says that the definitions will have one or more probable causes in the extension field of the definition. So, let's think about this later, but in any case, there won't be much changes when this is added. The method will be called GetAlarmDefinitionIDAndProbableCauseID instead of just GetAlarmDefinitionID.

internal/service/alarms/internal/alertmanager/alertmanager.go
mlguerrero12: You mean in the Alarm Record? If so, I don´t think we should remove that field. It is defined in the spec. So, we simply continue leaving empty

PR #534, opened 2025-02-03
CNF-15854: Fix CloudInfo attribute name for global cloud id

alegacy:
The API specification uses a lower case "C" in "globalCloudId" therefore we need to adjust our definition until they correct the inconsistency in the documentation and conformance test suite.


internal/service/resources/serve.go
```diff
@@ -140,7 +140,7 @@ func Serve(config *api.ResourceServerConfig) error {
 		Repo:   repository,
 		Info: generated.OCloudInfo{
 			Description:   "OpenShift O-Cloud Manager",
-			GlobalCloudId: globalCloudID,
+			GlobalcloudId: globalCloudID,
```
bartwensley, line 143: Probably a dumb question, but why is the leading "G" capitalized here?

alegacy: Because it is a public/exported field on the Go struct.  If you look in the `inventory.generated.go` file  you'll see the `json:"globalcloudId"` annotation which says how it will be serialized for output.

bartwensley: Cool - thanks Allain

PR #535, opened 2025-02-03
Include hwMgrId value in inventory uuid generation

alegacy:
* Link ClusterResource to Resource
    
    This links the ClusterResource object to the Cluster object via the
    ResourceID UUID value.  This depends on labels to be applied to the
    Agent object so that we know which hwMgrID and hwMgrNodeID was used
    to implement the node.   This labels will be provided in a separate
    PR.
    
* Add labels to inventory resource
    
    This adds the "labels" provided by the hardware manager as extensions
    to the inventory resource object.
    
*  Include hwMgrId value in inventory uuid generation
    
    We generate inventory UUID values based on the cloud id and the resource
    name but the resource name does not necessarily have to be unique
    across multiple hardware managers therefore we include the hwMgrId into
    the string used to generate the UUID.


alegacy:
rebased

PR #546, opened 2025-02-06
Add disable-auto-import annotation to prevent stale ManagedCluster re-imported

Missxiaoguo:
A workaround to add the `import.open-cluster-management.io/disable-auto-import annotation` to ManagedCluster via the rendered ClusterInstance's `spec.ExtraAnnotations.ManagedCluster`. The annotation temporarily disables auto-import for ManagedCluster when it gets created. This update is to prevent a race condition where stale cluster might be re-imported because timing issues. The annotation is removed from ManagedCluster once the cluster installation process begins.

This also includes a fix to update `provisioningStatus.updateTime` only when `provisioningStatus.provisioningPhase` or `provisioningStatus.provisioningDetails` is changed.


Missxiaoguo:
This issue is hard to reproduce, but I confirmed from the managedcluster-import-controller logs that auto-import reconcilation is not triggered after adding the annotation, and the provisioning works as expected.

PR #547, opened 2025-02-06
update siteconfig operator namespace to open-cluster-management in the sample CRs

borball:
If the prefered method to enable siteconfig operator is through the command 
`oc patch multiclusterhubs.operator.open-cluster-management.io multiclusterhub -n <ACM_NAMESPACE> --type json --patch '[{"op": "add", "path":"/spec/overrides/components/-", "value": {"name":"siteconfig","enabled": true}}]'`
, then the default namespace of siteconfig operator will be open-cluster-management. 

This PR is to update the namespace to open-cluster-management in the samples in order to avoid confusion.


Missxiaoguo:
/ok-to-test

I realized this yesterday too and was planning to update these examples, but I haven’t gotten to it yet. Thank you for this commit!!!

PR #549, opened 2025-02-07
fix cluster server assertion error

pixelsoccupied:
{"time":"2025-02-07T14:59:31.862941086Z","level":"ERROR","source":{"function":"github.com/openshift-kni/oran-o2ims/internal/service/cluster/collector.(*AlarmsDataSource).buildAlarmDictionaryIDToAlarmDefinitions","file":"/workspace/internal/service/cluster/collector/collector_alarms.go","line":292},"msg":"no alarm dictionary ID found for node cluster type","NodeClusterType ID":"14ec7a85-8304-5455-a597-0770e46dcb20"}
{"time":"2025-02-07T14:59:31.863104628Z","level":"ERROR","source":{"function":"github.com/openshift-kni/oran-o2ims/internal/service/cluster/collector.(*AlarmsDataSource).makeAlarmDictionaries","file":"/workspace/internal/service/cluster/collector/collector_alarms.go","line":393},"msg":"no alarm dictionary ID found for node cluster type","NodeClusterType ID":"14ec7a85-8304-5455-a597-0770e46dcb20"}
{"time":"2025-02-07T14:59:31.863230911Z","level":"WARN","source":{"function":"github.com/openshift-kni/oran-o2ims/internal/service/cluster/collector.(*Collector).execute","file":"/workspace/internal/service/cluster/collector/collector.go","line":175},"msg":"failed to collect data from data source","source":"Alarms","error":"failed to sync alarm dictionaries: no alarm dictionaries were processed"}

### DIFF
--- a/internal/service/cluster/collector/collector_alarms.go
+++ b/internal/service/cluster/collector/collector_alarms.go
@@ -101,11 +101,12 @@ func nodeClusterTypesWithAlarmDictionaryID(nodeClusterTypes []models.NodeCluster
 
 		alarmDictionaryIDString := (*nodeClusterType.Extensions)[utils.ClusterAlarmDictionaryIDExtension]
 		if alarmDictionaryIDString != nil {
-			_, err := uuid.Parse(alarmDictionaryIDString.(string))
+			id, err := uuid.Parse(alarmDictionaryIDString.(string))
 			if err != nil {
 				slog.Error("error parsing alarm dictionary ID", "NodeClusterType ID", nodeClusterType.NodeClusterTypeID, "error", err)
 				continue
 			}
+			(*nodeClusterType.Extensions)[utils.ClusterAlarmDictionaryIDExtension] = id
 
 			filteredNodeClusterTypes = append(filteredNodeClusterTypes, nodeClusterType)
 		}


internal/service/cluster/collector/collector_alarms.go
```diff
@@ -99,17 +99,17 @@ func nodeClusterTypesWithAlarmDictionaryID(nodeClusterTypes []models.NodeCluster
...
+			_, err := uuid.Parse(alarmDictionaryIDString)
 			if err != nil {
 				slog.Error("error parsing alarm dictionary ID", "NodeClusterType ID", nodeClusterType.NodeClusterTypeID, "error", err)
 				continue
 			}
-
-			filteredNodeClusterTypes = append(filteredNodeClusterTypes, nodeClusterType)
 		}
+
+		filteredNodeClusterTypes = append(filteredNodeClusterTypes, nodeClusterType)
```
alegacy, line 110: Shouldn't this be inside the previous `if` given that it supposed to be filtering the list down to only those with a valid extension?

internal/service/cluster/collector/collector_alarms.go
```diff
@@ -110,6 +109,7 @@ func nodeClusterTypesWithAlarmDictionaryID(nodeClusterTypes []models.NodeCluster
 			filteredNodeClusterTypes = append(filteredNodeClusterTypes, nodeClusterType)
```
mlguerrero12, line 109: We only needed to change the format of the extension from String to UUID for the ones with valid parsed UUID. The problem here is that the DB returns String not UUID as expected. Note that in the future, it will be UUID once the alarmDictionaryID is properly defined.

We don´t need to modify anything else. In fact, your changes below will break for dynamic additions of nodeClusterTypes since the make of the nodeClusterType sets an UUID in the extension

alegacy: the extensions should always be strings.  we shouldn't store UUID typed objects directly in them.

internal/service/cluster/collector/collector_alarms.go
pixelsoccupied: reverted

internal/service/cluster/collector/collector_alarms.go
alegacy: the only reason they are defined as `interface{}` is to allow nested sets of extensions that seem to be required in some cases to pass the compliance tests.

mlguerrero12: They are stored as string in the DB and presented in the API as strings. For convenience, the UUID is directly added in the extension at the moment of creating the nodeClusterType object to avoid parsing later. That's for dynamic addition. For the resync mechanism, it gets the DB returns Strings, again, to have a single UUID parse, the idea is to transform the extension here. 

It is to simplify the rest of the code. Note that all this will be deleted once the dictionary ID is out of the extension

PR #552, opened 2025-02-10
send AlarmAcknowledged notification immediately

pixelsoccupied:
When an alarm is Acknowledged we should send the notification right away


internal/service/alarms/api/server.go
```diff
@@ -374,6 +379,12 @@ func (a *AlarmsServer) PatchAlarm(ctx context.Context, request api.PatchAlarmReq
 		return nil, fmt.Errorf("failed to patch Alarm Event Record: %w", err)
 	}
 
+	// Send a notification since we have now an ack event
```
alegacy, line 382: This is another example where an transaction based outbox pattern would help.  If the process restarts/crashes in-between lines 377 and 385 that notification would be lost.

pixelsoccupied: Agreed. (looking into adopting outbox for alarms with another PR)

PR #558, opened 2025-02-11
Return string instead of int for PerceivedSeverity

pixelsoccupied:
PerceivedSeverity used to to be int but now we are returning string. 

/cc @alegacy


pixelsoccupied:
@alegacy no worries at all! Closing this for now! 

/close

PR #559, opened 2025-02-12
Fix alarm conformance tests

alegacy:
* Fix capitalization of alarm event attributes
    
    The interface spec uses "ID" instead of "Id" as the suffix on several
    alarm event record attributes.  Aligning them now to pass the
    conformance tests.
    
* Fix filter comparison of boolean fields
    
    The conformance tests are performance an alarm query with a filter set
    to: (eq,alarmAcknowledged,false) therefore, we need to support boolean
    fields in the filter search evaluator.


internal/search/selector_evaluator.go
```diff
@@ -239,6 +239,9 @@ func (e *SelectorEvaluator) evaluateEq(value any, args []any) (result bool,
 	case float64:
 		arg := arg.(float64)
 		result = value == arg
+	case bool:
+		arg := arg.(bool)
+		result = value == arg
```
pixelsoccupied, line 244: wonder if there's a cleaner way to handle maybe with another library? feels like we have to keep doing this we find new types (last time you had to tackle float) 🫤

PR #560, opened 2025-02-12
Fix provisioning status for hw configuration

Missxiaoguo:
Description:
- For cases where the `Configured` condition is not yet present in the NodePool, but hardware configuration updates have occurred, requeue to wait for the NodePool to be processed.
- Set hardware provisioning startTime after the NodePool is created, and hardware configuration startTime after the NodePool is updated. This helps distinguish between cases where no hardware updates were requested and cases where updates were made but the NodePool hasn’t been processed yet (i.e., the Configured condition hasn’t been returned).
- Set the `HardwareProvisioned` and `HardwareConfigured` conditions, along with the `provisioningPhase`, to `“Waiting for NodePool (name) to be processed”` when the Provisioned or Configured condition is not yet present in the NodePool.
- Add retries for some essential resource queries.
- Add unittests for hw configuration.

### DIFF
--- a/internal/controllers/provisioningrequest_clusterconfig.go
+++ b/internal/controllers/provisioningrequest_clusterconfig.go
@@ -115,6 +115,13 @@ func (t *provisioningRequestReconcilerTask) updateConfigurationAppliedStatus(
 
 	// Update the ConfigurationApplied condition.
 	if allPoliciesCompliant {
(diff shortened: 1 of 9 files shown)


Missxiaoguo:
Tested provisioning and configuration with hw lookback adaptor

PR #562, opened 2025-02-13
Update doc to include ArgoCD integration

borball:
Draft version and ask for a review

### DIFF
--- a/docs/argocd/app-project.yaml
+++ b/docs/argocd/app-project.yaml
@@ -1,0 +1,51 @@
+apiVersion: argoproj.io/v1alpha1
+kind: AppProject
+metadata:
+  name: ocloud-cluster-templates
+  namespace: openshift-gitops
+spec:
+  clusterResourceWhitelist:
+  - group: 'hive.openshift.io'
+    kind: ClusterImageSet
+  - group: 'cluster.open-cluster-management.io'
+    kind: ManagedCluster
+  - group: ''
+    kind: Namespace
+  destinations:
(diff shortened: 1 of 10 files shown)


borball:
Thanks @irinamihai for your comments, the reason I added the argoCD integration here is that when I did the integration during my test I realized there was plenty minor difference than the instruction in the [tranditional ZTP](https://github.com/openshift-kni/cnf-features-deploy), for example some ORAN related CRDs shall be added in the cluster AppProject:

```
  - group: o2ims.provisioning.oran.org
    kind: ClusterTemplate
  - group: hwmgr-plugin.oran.openshift.io
    kind: HardwareManager
  - group: o2ims-hardwaremanagement.oran.openshift.io
    kind: HardwareTemplate
```

And more namespaces shall be added in policy AppProject like 'oran*':
```
  destinations:
  - namespace: 'ztp*'
    server: '*'
  - namespace: 'policies-sub'
    server: '*'
  - namespace: 'sno*'
    server: '*'
  - namespace: 'oran*'
    server: '*'
  - namespace: 'open-cluster-management*'
    server: '*'
```
The argocd-openshift-gitops-patch.json may not need the 'ztp-site-generate' plugin (this was not addressed in my current commit, I can remove it in a new commit if we can confirm). 

To me all those are ORAN specific settings and it doesn't make sense to include those in [cnf-features-deploy](https://github.com/openshift-kni/cnf-features-deploy) repo, how do you think?

PR #573, opened 2025-02-20
CNF-16727: improve alarms notification reliability with an outbox and event-driven with pg listen/notify

pixelsoccupied:
- alarms events are now handled with a transaction outbox table. this allows us to make sure every change is detected and handled as a unique notification, improving overall reliability
- each outbox entry will now trigger a function automatically with PG listen/notify feature to forward the notifications, adhering to more event-driven design 
- new set of manager functions exposed to easily register listen/notify functions


internal/service/common/repo/repository.go
```diff
@@ -84,3 +89,111 @@ func (r *CommonRepository) DeleteDataChangeEvent(ctx context.Context, id uuid.UU
 func (r *CommonRepository) GetDataChangeEvents(ctx context.Context) ([]models2.DataChangeEvent, error) {
 	return utils.FindAll[models2.DataChangeEvent](ctx, r.Db)
 }
+
+// ClaimDataChangeEvent claims a batch of DataChangeEvent and updates the high-water mark.
+func ClaimDataChangeEvent(pool *pgxpool.Pool, ctx context.Context) ([]models2.DataChangeEvent, error) {
```
alegacy, line 94: Would you mind renaming the `models2` import to `models` while you are in here.  Must have gotten messed up when the code was moved around.

internal/service/common/repo/repository.go
```diff
@@ -84,3 +89,111 @@ func (r *CommonRepository) DeleteDataChangeEvent(ctx context.Context, id uuid.UU
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to execute queryGetHighWater: %w", err)
+	}
+
+	return highWatermark.LastEventID, nil
+}
+
+// updateHighWatermark update notification cursor with the highest datachange event
+func updateHighWatermark(ctx context.Context, tx pgx.Tx, dataChangeEvents []models2.DataChangeEvent, highWatermark int) error {
+	if len(dataChangeEvents) > 0 {
```
alegacy, line 172: Could rewrite this to the following and save the nesting level.

```
if len(dataChangeEvents) == 0 {
    return nil
}
```

internal/service/alarms/internal/db/repo/alarms_repository.go
```diff
@@ -111,14 +113,14 @@ func (ar *AlarmsRepository) GetAlarmSubscription(ctx context.Context, id uuid.UU
 }
 
 // UpsertAlarmEventRecord insert and updating an AlarmEventRecord.
-func (ar *AlarmsRepository) UpsertAlarmEventRecord(ctx context.Context, records []models.AlarmEventRecord) error {
+func (ar *AlarmsRepository) UpsertAlarmEventRecord(ctx context.Context, records []models.AlarmEventRecord, globalCloudId uuid.UUID) error {
```
alegacy, line 116: Wouldn't it be cleaner to add the globalCloudID when the `AlarmEventRecord` is built rather than deferring this step to the repository/db code to manage?

alegacy:
Mostly minor comments; otherwise, looks great!

internal/service/common/repo/repository.go
pixelsoccupied: renamed to `commonmodels` (just `models` was causing an error when generating the mock file)

internal/service/common/repo/repository.go
pixelsoccupied: updated

internal/service/alarms/internal/db/repo/alarms_repository.go
pixelsoccupied: yeah it was originally like that. Will revert

alegacy: There other option is to defer it all the way out to the converter to add it just before it is sent out as either a notification or an API response.

pixelsoccupied: > defer it all the way out to the converter to add it just before it is sent out 

Yeah going with that

pixelsoccupied:
/label tide/merge-method-rebase
/cc @alegacy 

should be good for another pass!

PR #574, opened 2025-02-25
Replace kube-rbac-proxy with middleware (part 1)

alegacy:
This adds middleware directly into the servers to handle authentication and authorization.  This activity was mostly triggered by the deprecation of the gcr.io kube-rbac-proxy image and the recommendation from the kubebuilder teams to remove the use of the proxy in favor of middleware.  Unfortunately, the middleware recommended was geared directly towards metrics and wasn't generic enough to use as is, but this code follows a similar approach with some customizations for our own circumstances as well as additional code to support the OAuth/OIDC use case.

This also includes a 2nd commit to do some minor cleanup in the mTLS space so that we can support a separate client certificate for each server if that becomes necessary.

This is part 1... we'll need to remove the scaffolded usage as well.


Missxiaoguo (commented):
/lgtm

PR #576, opened 2025-02-26
Fix the provisioning status for upgrade case

Missxiaoguo:
With the updates, the provisioning status now displays correctly throughout the upgrade.

When an upgrade is triggered or in progress, we should continue handling the upgrade instead of requeuing earlier due to NonCompliant policies that were bound as part of the upgrade process.

When an upgrade is requested, the status should move to fulfilled only after it is completed.


Missxiaoguo:
Tested provisioning, day2 config update and upgrade.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -224,42 +224,50 @@ func (t *provisioningRequestReconcilerTask) run(ctx context.Context) (ctrl.Resul
...
+		// An upgrade is requested or upgrade has started but not completed
+		if shouldUpgrade ||
+			(utils.IsClusterUpgradeInitiated(t.object) &&
+				!utils.IsClusterUpgradeCompleted(t.object)) {
+			upgradeCtrlResult, err := t.handleUpgrade(ctx, renderedClusterInstance.GetName())
 			if err != nil {
 				return requeueWithError(err)
 			}
-			return requeue, nil
+			if !upgradeCtrlResult.IsZero() {
```
irinamihai, line 254: Would it be the same here to use `if upgradeCtrlResult != doNotRequeue()`? I think it would be more clear, but it's up to you.
A comment above explaining the use case a bit more would definitely be helpful. :pray:

Missxiaoguo: yup, sure. I will change to `upgradeCtrlResult.RequeueAfter > 0` which seems to be more clear

internal/controllers/provisioningrequest_clusterconfig.go
```diff
@@ -155,6 +151,8 @@ func (t *provisioningRequestReconcilerTask) updateConfigurationAppliedStatus(
 			"The Cluster is not yet ready",
 		)
 		if utils.IsClusterProvisionCompleted(t.object) &&
+			(!utils.IsClusterUpgradeInitiated(t.object) ||
```
irinamihai, line 154: Looks like
`!utils.IsClusterUpgradeInitiated(t.object) ||
				utils.IsClusterUpgradeCompleted(t.object)`
could be moved in its own function since it's been used 3 times, something like `IsClusterUpgradeInProgress`.
It can be done in another PR, for sure. I think it would greatly improve the code's readability.

PR #577, opened 2025-02-26
CNF-16737: Validate the ClusterInstance defaults ConfigMap against the ClusterInstance CRD schema

irinamihai:
Steps:
- obtain the ClusterInstance CRD and get the schema for the CRD version that is stored and served
- from the schema obtained above, remove all the `required` properties since the ClusterInstance default ConfigMap will probably not contain all the required configuration.
- from the ClusterInstance default ConfigMap, remove the `label` property for each node interface as that is not part of the ClusterInstance schema, but a mechanism used by the O-Cloud Manager for provisioning
- validate the ConfigMap content against the CRD schema using the existing `ValidateJsonAgainstJsonSchema`  function

Updated existing unit tests and added more for the extra functionality.

### DIFF
--- a/internal/controllers/utils/provision.go
+++ b/internal/controllers/utils/provision.go
@@ -14,11 +14,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
 	provisioningv1alpha1 "github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1"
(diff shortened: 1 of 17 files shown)


internal/controllers/utils/provision.go
```diff
@@ -262,6 +265,127 @@ func ValidateDefaultInterfaces[T any](data T) error {
...
+			nodeMap, ok := node.(map[string]interface{})
+			if !ok {
+				return fmt.Errorf("unexpected: invalid node data structure")
+			}
+			interfaces := getInterfaces(nodeMap)
+			if interfaces == nil {
+				return fmt.Errorf("failed to extract the interfaces from the node map")
+			}
+			for _, intf := range interfaces {
+				if value, exists := intf["label"]; exists && value != "" {
```
Missxiaoguo, line 284: should the label be removed if its value is empty? It seems like this `if` check can be removed since if it is nil or there is no such element, `delete` is a no-op.

Missxiaoguo (commented):
Looks good to me! Just one small comment

internal/controllers/utils/provision.go
irinamihai: Right, we can remove the check and do the delete anyway, it won't be an overhead.

PR #579, opened 2025-02-26
CNF-16878: Add client certificate binding support (RFC8705)

alegacy:
This adds support for client certificate bound access tokens (e.g., RFC8705).  When this feature is enabled on the authorization server each token generated contains an attribute which represents the SHA256 fingerprint of the TLS certificate which was used by the client who requested the token.

When this attribute is added to the token, Resource Servers are supposed to validate that the token is only used by a client presenting that same TLS certificate.  This is verified by comparing the fingerprint contained within the token with the fingerprint of the certificate being used to make the request.

We only support being behind an ingress proxy; therefore, we never expect TLS to be terminated directly on each server for OAuth use cases. For this reason, we only expect to get the client certificate information from the proxy headers rather than from the local TLS stack.  Since we only expect to be behind an HAProxy, we currently only support the client certificate headers added by HAProxy and the proposed standard of RFC9440.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,12 @@ regarding the OAuth settings and JWT contents.
    cases, this may require special configuration steps on the Authorization Server to ensure the proper format in the
    JWT tokens.
 
+7. Client certificate binding (i.e., RFC8705) is enabled by default. If the authorization server has inserted a claim
+   into the token which contains the SHA256 fingerprint of the client certificate which requested a token then it will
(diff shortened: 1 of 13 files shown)


pixelsoccupied:
this is so cool...kind of like 2fa for your token

/lgtm
/approve

alegacy:
> this is so cool...kind of like 2fa for your token
exactly!

Issue #365, opened 2024-12-01
:warning: Action Required: Replace Deprecated gcr.io/kubebuilder/kube-rbac-proxy


camilamacedo86:
## Description

:warning: **The image `gcr.io/kubebuilder/kube-rbac-proxy` is deprecated and will become unavailable.**
**You must move as soon as possible, sometime from early 2025, the GCR will go away.**

> _Unfortunately, we're unable to provide any guarantees regarding timelines or potential extensions at this time. Images provided under GRC will be unavailable from March 18, 2025, [as per announcement](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr). However, `gcr.io/kubebuilder/`may be unavailable before this date due [to efforts to deprecate infrastructure](https://github.com/kubernetes/k8s.io/issues/2647)._

- **If your project uses `gcr.io/kubebuilder/kube-rbac-proxy`, it will be affected.**
  Your project may fail to work if the image cannot be pulled. **You must take action as soon as possible.**

- However, if your project is **no longer using this image**, **no action is required**, and you can close this issue.

## Using the image `gcr.io/kubebuilder/kube-rbac-proxy`?

[kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) was historically used to protect the metrics endpoint. However, its usage has been discontinued in [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder). The default scaffold now leverages the [`WithAuthenticationAndAuthorization`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization) feature provided by [Controller-Runtime](https://github.com/kubernetes-sigs/controller-runtime).

This feature provides integrated support for securing metrics endpoints by embedding authentication (`authn`) and authorization (`authz`) mechanisms directly into the controller manager's metrics server, replacing the need for (https://github.com/brancz/kube-rbac-proxy) to secure metrics endpoints.

### What To Do?

**You must replace the deprecated image `gcr.io/kubebuilder/kube-rbac-proxy` with an alternative approach. For example:**
- Update your project to use [`WithAuthenticationAndAuthorization`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization):
[...]

PR #580, opened 2025-02-27
CNF-16877: Replace kube-rbac-proxy for controller metrics

alegacy:
This replaces the usage of kube-rbac-proxy to protect the /metrics endpoint with the recommended WithAuthenticationAndAuthorization middleware.  The endpoint remains protected by a TLS certificate signed by the OpenShift Service Serving Signer CA.

Fixes: #365


alegacy:
@rauhersu FYI, this removes the last remnants of kube-rbac-proxy in this repo.

config/default/manager_webhook_patch.yaml
```diff
@@ -5,13 +5,13 @@
     secret:
       defaultMode: 420
       secretName: oran-o2ims-webhook-server-cert
-- path: /spec/template/spec/containers/1/ports/0
+- path: /spec/template/spec/containers/0/ports/0
```
donpenney, line 8: @rauhersu does this impact the formal build?

alegacy: I wouldn't think so.  Why would it?

rauhersu:
@alegacy , have these changes been implemented inspired by https://issues.redhat.com/browse/CNF-16321 finally  ? 
If so, would you drop a note there + link this MR ? 

@donpenney , we need to fix our overlay too then. Can we apply the same logic to get rid of this on the plugin side ? 

These changes are excellent news, and if we make both ims+plugin independent of kube-rbac-proxy, you will greatly benefit of this in the future: you won't have extra delays when you require a production release because of ocp synchronisation as all our operators have atm. The effort will pay off !

rauhersu:
keeping this merge on hold at the moment, as it will likely break your builds (or make it even worse: you will get malformed brew builds.

/hold

config/default/manager_webhook_patch.yaml
rauhersu: yes, we need to re-design our midstream overlay.

rauhersu:
/unhold

We can merge this asap and will check my midstream code next.

PR #585, opened 2025-03-05
Add envTest for controllers

irinamihai:
This PR adds a new envTest suite under the `test` directory. Right now there is only one test to ensure the functionality of the cluster template & provisioning controllers. This change prepares the ground for removing the ClusterInstance [template](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/files/controllers/clusterinstance-template.yaml).

With this change we can test the provisioning controller against a minimalistic kube API server and the actual ClusterInstance CRD, thus enabling us to properly validate the rendered ClusterInstance CR with the `DryRun` option.

In the future, envTest can be configured to run on a test (kind) or any other available cluster.


test/utils/functions.go
```diff
@@ -0,0 +1,160 @@
...
+		Expect(commit).To(HaveKey("sha"))
+		// Extract the commit sha.
+		fullCommitSha := commit["sha"].(string)
+
+		// Get the CRD file.
+		crdFilePath := fmt.Sprintf(
+			GithubUserContentLink,
+			externalCrd["owner"], externalCrd["repoName"],
+			fullCommitSha, externalCrd["crdPath"], externalCrd["crdFileName"])
+		err = DownloadFile(crdFilePath, externalCrd["crdFileName"], destDir)
```
rauhersu, line 117: This is great to download only the specific filenames leveraging the github api !

irinamihai: I've tried downloading the full repos, like the hw manager plugin tests are doing, but it was taking too long for IMS :(
Although I don't like hard coding the names of the CRD files, I figured this is the next best thing. If those repos change their CRD names, then the tests will fail and at least it will be obvious what needs to be updated.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+const testHwMgrPluginNameSpace = "hwmgr"
+
+var (
+	K8SClient                     client.Client
+	K8SManager                    ctrl.Manager
+	ProvReqTestReconciler         *provisioningcontrollers.ProvisioningRequestReconciler
+	ClusterTemplateTestReconciler *provisioningcontrollers.ClusterTemplateReconciler
+	testEnv                       *envtest.Environment
+	ctx                           context.Context
```
pixelsoccupied, line 58: it seems like this global ctx is being overwritten in BeforeSuite and BeforeEach (with and without cancel)...might be a little unclear of the intention and potential resource leak? Might worth splitting it up? e.g in BeforeEach you have its own `testCtx := context.Background() // or context.WithTimeout if needed`

test/utils/functions.go
rauhersu: Don't know if this would be viable and the github api allows this, but maybe if all the CRDs we need are under a specific directory we might try to download the whole dir then ... Just thinking out loud as a future improvement.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+		Eventually(func() bool {
+			newct := &provisioningv1alpha1.ClusterTemplate{}
+			Expect(K8SClient.Get(context.Background(), client.ObjectKeyFromObject(ct), newct)).To(Succeed())
+			return newct.Status.Conditions != nil
+		}, timeout, interval).Should(BeTrue())
+	})
+
+	Context("Provisioning Request is created", func() {
+		It("Simple test for validating functionality", func() {
```
pixelsoccupied, line 357: how about `It("validates basic functionality", func() {`

pixelsoccupied (commented):
Just a few tiny comments but lgtm! 

/cc @alegacy

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,397 @@
...
+		tName         = "clustertemplate-a"
+		tVersion      = "v1.0.0"
+		ctNamespace   = "clustertemplate-a-v4-16"
+		ciDefaultsCm  = "clusterinstance-defaults-v1"
+		ptDefaultsCm  = "policytemplate-defaults-v1"
+		hwTemplate    = "hwtemplate-v1"
+		crName        = "aaaaaaaa-ac56-4143-9b10-d1a71517d04f"
+	)
+
+	BeforeEach(func() {
```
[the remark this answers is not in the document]
irinamihai: These will definitely be needed once more tests are added. I skipped it for this one example test, but I'll them once the suite evolves.

PR #592, opened 2025-03-10
Remove transition from failed to progressing

Missxiaoguo:
Updates:
- Removed the transition from failed to progressing. The system cannot autonomously recover from fatal errors. It requires either a PR update or delete/recreate to get rid of the error.
- Moved the preparation steps into a separate function to reduce the complexity of `run` function.

### DIFF
--- a/api/provisioning/v1alpha1/provisioningrequest_validation.go
+++ b/api/provisioning/v1alpha1/provisioningrequest_validation.go
@@ -350,3 +350,17 @@ func matchesAnyPattern(path []string, patterns [][]string) bool {
 	}
 	return false
 }
+
+// HasFatalProvisioningFailure checks if the ProvisioningRequest
+// has a fatal provisioning failure that cannot be recovered
+// on its own.
+func HasFatalProvisioningFailure(conditions []metav1.Condition) bool {
+	for _, condType := range FatalPRconditionTypes {
+		cond := meta.FindStatusCondition(conditions, string(condType))
+		if cond != nil && cond.Status == metav1.ConditionFalse &&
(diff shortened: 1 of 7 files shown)


api/provisioning/v1alpha1/provisioningrequest_validation.go
```diff
@@ -350,3 +350,27 @@ func matchesAnyPattern(path []string, patterns [][]string) bool {
 	}
 	return false
 }
+
+// HasFatalProvisioningFailure checks if the ProvisioningRequest
+// has a fatal provisioning failure that cannot be recovered
+// on its own.
+func HasFatalProvisioningFailure(conditions []metav1.Condition) bool {
+	var conditionTypes = []ConditionType{
```
donpenney, line 358: nit: It would be nice to avoid having a hardcoded list of these types, which could end up being stale if a new condition is added but the various places we have such a hardcoded list aren't updated. I think it would be better to either have this list defined in the same place as the PRconditionTypes var-struct, or determine it programmatically. I'm sure there's a nice way to do it, but one option that relies on the struct field name being the same as the string representation would be:
```
for _, f := range reflect.VisibleFields(reflect.TypeOf(PRconditionReasons)) {
```
This `reflect.VisibleFields` returns a list of the field names in the struct, and would avoid hardcoding.

Missxiaoguo: Thanks for your suggestion, Don! In our case, this cannot be fully determined programmatically since only a subset of the available condition types are considered fatal. Reflection could enumerate all condition types, but we would still need a way to mark only some as fatal (e.g., struct tag or a metadata map), which is essentially equivalent to defining a list of fatal conditions. 
For simplicity, I will just move this list to the same location where the PRconditionTypes is defined. Hopefully, the condition changes will be infrequent.

PR #597, opened 2025-03-12
CNF-16968: drop dependency on full payload to sync with webhook and instead leverage Alertmanager API

pixelsoccupied:
This total e2e improvement started with this PR https://github.com/openshift-kni/oran-o2ims/pull/583 

During the investigation it turns out that the child `group_by` (a.k.a oran route) can not be set to empty array which was needed to force a full list of old and current alerts during a webhook call. Our code was dependent on this since we were also using that callback from AM to sync the database. The workaround was to simply override the parent `group_by` to be empty array. But this solution was not really scalable (other than the obvious issue of override something existing/global!)

To solve the issue and make the server resilient + predictable and objectively faster few new things are added/improved
- Start alertmanager API communication (Pull data in) 
With the `/alerts` endpoint from AM we can view the full list of active alerts at anytime. This allowed us to run at 1 hour interval and sync with the DB. This means we are longer dependent on any `group_by` value to stay up-to-date. The 1 hour interval also means we sync more predictably without relying on webhook to call us. 
 - Start using `GenerationID` instead of `NOT IN` to sync  
We were originally using PG's problematic NOT IN to determine if an alert is stale. With alert payload of sufficiently large size tests showed delay of few SECONDS (instead of ms). With GenerationID we are able do the same thing a lot faster.
-  Cleanup and unit tests as needed

This API interaction additionally unlocked a new feature: Now we dont have to wait until the first webhook (which could be forever depending on the state of the system) as we can now do a sync at startup.


internal/service/alarms/internal/alertmanager/api.go
```diff
@@ -0,0 +1,272 @@
+package alertmanager
```
donpenney, line 1: Missing header block comment, in all new files

internal/service/alarms/internal/alertmanager/api.go
```diff
@@ -0,0 +1,271 @@
...
+	"github.com/openshift-kni/oran-o2ims/internal/service/alarms/internal/infrastructure"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ACMObsAMRouteName     = "alertmanager"                              // ACM's AM API host
+	ACMObsAMAuthSecrtName = "observability-alertmanager-accessor-token" // ACM's AM API BEARER and ca.Crt
```
alegacy, line 28: Missing `e` in `Secrt`

alegacy:
just a minor comment...otherwise, lgtm

internal/service/alarms/internal/alertmanager/api.go
pixelsoccupied: updated! (had to add no lint gosec)

pixelsoccupied:
/test scorecard
/cc @alegacy 

scorecard being flaky

PR #603, opened 2025-03-17
Align with the ClusterInstance validation check introduced in ACM 2.13

Missxiaoguo:
Siteconfig has introduced validation check in ACM 2.13+ that rejects any spec updates during cluster installation. This update aligns with that validation. 

Updates:
- ProvisioningRequest webhook and controller validation updates have been updated to disallow any updates during installation.
- The defaults set in the ClusterInstance CRD will be applied to the rendered ClusterInstance after dry-run, so run dry-run before checking ClusterInstance changes in the controller. If we start requiring ACM 2.13+, we could remove our post provisioning updates check and instead leverage the validation in the siteconfig.
- To pass the dry-run for the spec updates that caused by IMS:
     - Extract nodeInfo(e.g. bmc and macAddress that were from hw manager) from the existing ClusterInstance and merge it with the rendered ClusterInstance.
     - Remove auto-disable-import annotation from ClusterInstance CR when installation is completed.

### DIFF
--- a/internal/controllers/provisioningrequest_clusterinstall.go
+++ b/internal/controllers/provisioningrequest_clusterinstall.go
@@ -35,6 +35,15 @@ const (
 	disableAutoImportAnnotation = "import.open-cluster-management.io/disable-auto-import"
 )
 
+// nodeInfo represents the node details extracted
+// from the existing ClusterInstance
+type nodeInfo struct {
+	bmcAddress          *string
+	bootMACAddress      *string
+	bmcCredentialsName  *string
+	interfaceMACAddress map[string]string // keyed by interface name
+}
+
 func (t *provisioningRequestReconcilerTask) renderClusterInstanceTemplate(
 	ctx context.Context) (*siteconfig.ClusterInstance, error) {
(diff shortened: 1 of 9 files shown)


internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -447,3 +419,237 @@ func (t *provisioningRequestReconcilerTask) updateClusterProvisionStatus(ci *sit
...
+			return nil, fmt.Errorf("valid bmcCredentialsName not found")
+		}
+		v, ok = bmcCreds["name"].(string)
+		if !ok {
+			return nil, fmt.Errorf("valid name not found in bmcCredentialsName")
+		}
+		extractedNodeInfo.bmcCredentialsName = v
+
+		// Extract interface macAddress by interface name
+		nodeNetwork, ok := nodeMap["nodeNetwork"].(map[string]any)
```
irinamihai, line 580: Maybe not worth returning errors on fields that are not required in the ClusterInsance CRD.

PR #611, opened 2025-03-20
CNF-16133: run reachability check before creating subscription for applicable servers

pixelsoccupied:
when a subscription is requested to be added we now perform a GET call with the appropriate auth config and expect 204 as response.

if the call goes through we can be sure that a real notification be sent out later

/cc @alegacy @mlguerrero12

### DIFF
--- a/internal/service/alarms/DEVELOPING.md
+++ b/internal/service/alarms/DEVELOPING.md
@@ -157,51 +157,57 @@ SPDX-License-Identifier: Apache-2.0
 package main
 
 import (
-  "encoding/json"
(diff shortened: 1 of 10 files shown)


pixelsoccupied:
hold it for a few until folks get a chance update SMOs for testing as needed but ready for review till then

alegacy:
/lgtm

tested on my setup with mTLS/OAuth enabled with our mock SMO.

pixelsoccupied:
/unhold 

QE has till 4.20 to update stuff if needed. we can now merge.

PR #618, opened 2025-03-24
CNF-16737: Remove ClusterInstance template

irinamihai:
- removes the golang template used to render the ClusterInstance
- when building a ClusterInstance, we assign the data obtained from merging the ProvisioningRequest & the Cluster Instance default configmap directly to the `spec` of the unstructured ClusterInstance.
- hardware details are still set as defaults initially for the case of hardware provisioning since they will be overwritten.
- envTest added, fakeclient tests updates
- tested end to end install with and without hardware provisioning; day 2 changes


Missxiaoguo:
Since the ClusterTemplate has now been removed, we could also consider using unstructured ClusterInstance for create/update in the future, so we don't have to bump the siteconfig version when new spec fields are added.

PR #633, opened 2025-04-11
fix manager konflux build

rauhersu:


.tekton/o-cloud-manager-4-19-build.yaml
```diff
@@ -0,0 +1,634 @@
+lapiVersion: tekton.dev/v1
```
fontivan, line 1: you have a typo here, with an accidental `L` in front of `apiVersion`

.konflux/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build stage
-FROM ${BUILDER_IMAGE} AS builder
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.23@sha256:2d5976ded2a3abda6966949c4d545d0cdd88a4d6a15989af38ca5e30e430a619 AS builder
 
 WORKDIR app
```
fontivan, line 4: not really Konflux related but should use `/app` instead of `app`

PR #650, opened 2025-04-25
build the bundle...

rauhersu:
This will be refined in posterior PRs

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,4 +1,5 @@
 *
+!/.konflux
 !/LICENSE.txt
 !/Makefile
 !/README.md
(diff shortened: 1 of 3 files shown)


.containerignore
```diff
@@ -1,4 +1,5 @@
 *
+!/.konflux
```
fontivan, line 2: I forgot this was a thing, will add it to the low priority to do list for the rest of the products

rauhersu: Yeah, got an error message initially "1 filtered out using /var/workdir/source/.containerignore" when copying.

PR #659, opened 2025-05-05
fix catalog ...   

rauhersu:
We needed a new 'annotations.yaml' with the proper configuration matching csv files. This was causing an invalid catalog when                                                                                                                                                                                                   
rendering the catalog template.                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                               
Full description of the problem: https://redhat-internal.slack.com/archives/C074JM28DTP/p1746193302778409                                                                                                                                                      
                                                                                                                                                                                                                                                               
Additionally, I have introduced our own scorecard config and minor renaming.

### DIFF
--- a/.konflux/Dockerfile.bundle
+++ b/.konflux/Dockerfile.bundle
@@ -14,8 +14,7 @@ RUN mkdir -p $OVERLAY_PATH
 COPY .konflux/overlay/ $OVERLAY_PATH
 
 # Run the overlay
-RUN mv $MANIFESTS_PATH/oran-o2ims.clusterserviceversion.yaml $MANIFESTS_PATH/o-cloud-manager.clusterserviceversion.yaml \
-    && $OVERLAY_PATH/overlay.bash --set-pinning-file $OVERLAY_PATH/pin_images.in.yaml --set-csv-file $MANIFESTS_PATH/o-cloud-manager.clusterserviceversion.yaml
+RUN $OVERLAY_PATH/overlay.bash --set-pinning-file $OVERLAY_PATH/pin_images.in.yaml --set-csv-file $MANIFESTS_PATH/oran-o2ims.clusterserviceversion.yaml
 
 # Bundle stage: store the files a bundle needs
 #
--- a/.konflux/Dockerfile.bundle
+++ b/.konflux/Dockerfile.bundle
@@ -14,7 +14,8 @@ RUN mkdir -p $OVERLAY_PATH
 COPY .konflux/overlay/ $OVERLAY_PATH
 
 # Run the overlay
-RUN $OVERLAY_PATH/overlay.bash --set-pinning-file $OVERLAY_PATH/pin_images.in.yaml --set-csv-file $MANIFESTS_PATH/oran-o2ims.clusterserviceversion.yaml
+RUN mv $MANIFESTS_PATH/oran-o2ims.clusterserviceversion.yaml $MANIFESTS_PATH/o-cloud-manager.clusterserviceversion.yaml \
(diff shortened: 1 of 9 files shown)


.konflux/Dockerfile.bundle
```diff
@@ -38,18 +38,21 @@ LABEL com.redhat.openshift.versions="=v4.19"
...
 LABEL io.openshift.expose-services=""
 LABEL io.openshift.maintainer.component="o-cloud-manager"
 
-
 # Copy files to locations specified by labels
 COPY --from=overlay /tmp/manifests /manifests/
 COPY bundle/metadata /metadata/
 COPY bundle/tests/scorecard /tests/scorecard/
+
+# Replace additional bundle files
```
yanirq, line 56: do you still need to copy over:
```
COPY bundle/metadata /metadata/
COPY bundle/tests/scorecard /tests/scorecard/
```
?

.konflux/bundle/metadata/annotations.yaml
```diff
@@ -0,0 +1,14 @@
+annotations:
+  # Core bundle annotations.
+  operators.operatorframework.io.bundle.mediatype.v1: registry+v1
+  operators.operatorframework.io.bundle.manifests.v1: manifests/
+  operators.operatorframework.io.bundle.metadata.v1: metadata/
+  operators.operatorframework.io.bundle.package.v1: o-cloud-manager
+  operators.operatorframework.io.bundle.channels.v1: stable,4.19
```
yanirq, line 7: should we set this already to 4.20 ?

.konflux/Dockerfile.bundle
rauhersu: Just in case, by any reason, they put further files here. I am only interested in overriding 'annotations' and 'scorecard' specific files so far.

.konflux/bundle/metadata/annotations.yaml
rauhersu: Not yet, all this development on ims is targeting 4.19 exclusively, 'main' branch here does not indicate 4.20 so far.

.konflux/Dockerfile.bundle
yanirq: ack . asking since the dockerfile is already konflux specific.

.konflux/bundle/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: scorecard.operatorframework.io/v1alpha3
```
fontivan, line 1: where does the scorecard configuration come from?

.konflux/overlay/overlay.bash
```diff
@@ -52,9 +52,9 @@ add_related_images() {
     # create a new section from scratch
     declare -i index=0
     for image_name in "${!IMAGE_TO_SOURCE[@]}"; do
-        echo "Adding related image: image_name: $image_name source: ${IMAGE_TO_SOURCE[$image_name]}, target: ${IMAGE_TO_TARGET[$image_name]}"
+        echo "Adding related image: name: $image_name source: ${IMAGE_TO_SOURCE[$image_name]}, image: ${IMAGE_TO_TARGET[$image_name]}"
```
fontivan, line 55: these fixes will be required in all operators right?

.konflux/bundle/tests/scorecard/config.yaml
rauhersu: Our current cpaas file.

.konflux/overlay/overlay.bash
rauhersu: True (line 57 specifically) ! Cc: @yanirq

PR #690, opened 2025-05-26
Add new HardwarePlugin CRD for registering HW plugins

sakhoury:
# Summary

This PR introduces a new CustomResourceDefinition called `HardwarePlugin` for registering hardware plugins. It also includes standard scaffolding content generated from the operator-sdk. 

/cc @browsell @donpenney @alegacy @tliu2021


api/hardwaremanagement/v1alpha1/hardwareplugin_types.go
```diff
@@ -0,0 +1,55 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// HardwarePluginSpec defines the desired state of HardwarePlugin
+type HardwarePluginSpec struct {
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// ApiEndpoint is the root URL for the Hardware Plugin.
+	// +kubebuilder:validation:MinLength=1
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Hardware Plugin API Endpoint",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	ApiEndpoint string `json:"apiEndpoint"`
```
alegacy, line 23: Since this is expected to be something like `https://my-plugin.example.com:8443` then could we rename this to `apiRoot` to avoid confusing users into thinking that they need to provide some functional API endpoint.

PR #703, opened 2025-05-30
Add shared common authentication data types

sakhoury:
# Summary

This PR refactors the CRDs to use shared common authentication types. The `HardwarePlugin` CRD is also updated to incorporate the following fields:
- `CaBundleName`
- `AuthConfig`

/cc @alegacy @browsell


api/common/common_types.go
```diff
@@ -0,0 +1,120 @@
...
+	// fingerprint.  This is expected to be a CEL mapper expression.  It should only be changed in advanced scenarios.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OAuth Client Binding Claim",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	//+kubebuilder:default="has(claims.cnf) ? claims.cnf['x5t#S256'] : []"
+	ClientBindingClaim string `json:"clientBindingClaim"`
+}
+
+// +kubebuilder:object:generate=true
+// TLSConfig defines the TLS specific attributes
+type TLSConfig struct {
+	// SecretName represents the name of a secret (in the current namespace) which contains an X.509 certificate and
```
alegacy, line 52: This doc is likely going to need to be reworded to be a little more generic and a little less ORAN specific.

api/common/common_types.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// +kubebuilder:object:generate=true
+// AuthType defines the authorization type used for authentication.
+type AuthType string
+
+const (
+	// ServiceAccount represents authentication using a Kubernetes ServiceAccount token. This method leverages
+	// a Kubernetes ServiceAccount to authenticate requests, typically used for internal cluster communication
+	// where the client presents a JWT token issued by the Kubernetes API server.
```
alegacy, line 70: I would be explicit here and state that we assume that the plugin is going to be a service running in this same cluster... rather than in another cluster because then this wouldn't work at all and OAuth2 would need to be used instead... or we'd need to support TLS client certificate authentication to that other cluster's ingress. ... which may not be a bad idea to support, but not today!

api/common/common_types.go
```diff
@@ -0,0 +1,120 @@
...
+	//+kubebuilder:validation:Enum=ServiceAccount;Basic;OAuth
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Authentication Type",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:ServiceAccount","urn:alm:descriptor:com.tectonic.ui:select:Basic","urn:alm:descriptor:com.tectonic.ui:select:OAuth"}
+	Type AuthType `json:"type"`
+
+	// ServiceAccountName specifies the name of the Kubernetes ServiceAccount to use for authentication.
+	// This field is required when Type is set to "ServiceAccount". The ServiceAccount must exist in the
+	// current namespace, and its associated token will be used for authentication.
+	//+optional
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Service Account Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	ServiceAccountName *string `json:"serviceAccountName,omitempty"`
```
alegacy, line 100: I would leave this out.   The hardware-manager plugin should be instantiated with a service account when the operator creates it (e.g., like we do for the alarm-server, resource-server, etc... they each have there own service account).

PR #706, opened 2025-05-30
Add validation logic when registering a new HardwarePlugin

sakhoury:
# Summary

This PR introduces the registration validation process for onboarding new `HardwarePlugins`. It includes logic to verify that the `apiRoot` of the HardwarePlugin server is accessible.

/cc @browsell @alegacy

### DIFF
--- a/hwmgr-plugins/controller/hardwareplugin_controller.go
+++ b/hwmgr-plugins/controller/hardwareplugin_controller.go
@@ -10,9 +10,12 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net/http"
+	"net/url"
+	"strings"
 
-	"github.com/openshift-kni/oran-o2ims/hwmgr-plugins/controller/utils"
 	"github.com/openshift-kni/oran-o2ims/internal/logging"
+	"github.com/openshift-kni/oran-o2ims/internal/service/common/notifier"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -21,6 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	pluginv1alpha1 "github.com/openshift-kni/oran-o2ims/api/hardwaremanagement/v1alpha1"
+	"github.com/openshift-kni/oran-o2ims/hwmgr-plugins/controller/utils"
+	sharedutils "github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 )
 
 // HardwarePluginReconciler reconciles a HardwarePlugin object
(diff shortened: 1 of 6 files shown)


hwmgr-plugins/controller/hardwareplugin_controller.go
```diff
@@ -99,9 +104,111 @@ func (r *HardwarePluginReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
+
+	if hwplugin.Spec.AuthClientConfig.TLSConfig != nil && hwplugin.Spec.AuthClientConfig.TLSConfig.SecretName != nil {
+		secretName := *hwplugin.Spec.AuthClientConfig.TLSConfig.SecretName
+		cert, key, err := sharedutils.GetKeyPairFromSecret(ctx, r.Client, secretName, hwplugin.Namespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get certificate and key from secret: %w", err)
+		}
+
+		config.TLSConfig.ClientCert = sharedutils.NewStaticKeyPairLoader(cert, key)
```
alegacy, line 133: Using a static loader here is ok if this client is only ever going to be used once for validation, but if this is a long-lived client that we'll use to talk to the plugin then we'll need to handle this differently with something that can refresh the token periodically whenever the secret is replaced with an updated certificate.  That's best handled if the secret is mounted to a volume and we can use a dynamic cert loader.   We can discuss what's needed here offline.

hwmgr-plugins/controller/hardwareplugin_controller.go
```diff
@@ -99,9 +104,111 @@ func (r *HardwarePluginReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	// Build request to validation endpoint url
+	url := fmt.Sprintf("%s%s", apiRoot, sharedutils.HardwarePluginValidationEndpoint)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return false, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// Send request to HardwarePlugin server
+	resp, err := httpClient.Do(req)
```
alegacy, line 200: This is ok for now, but what we should really be doing here is code-generating the client code for our API and using that to access the plugin API endpoints.  We'll have to do that anyway for when we actually access the plugin for 'real' stuff.

internal/controllers/hardwareplugin_manager_setup.go
```diff
@@ -197,6 +197,16 @@ func (t *reconcilerTask) createHardwarePluginManagerClusterRole(ctx context.Cont
 					"update",
 				},
 			},
+			// HardwarePlugin registration verification
+			{
+				NonResourceURLs: []string{
+					"/hardware-manager/provisioning/*",
+					utils.HardwarePluginValidationEndpoint,
```
alegacy, line 204: Is this line necessary if the wildcard above it already exists?

hwmgr-plugins/controller/hardwareplugin_controller.go
sakhoury: 100% agreed - I have this for now in order get the validation mechanism working in the absence of a HardwarePlugin server to test with. I have a PR ready that refactors the generated code and introduces the client code for the plugin API. I will add a "TODO comment here" to replace this segment with the generated version.

internal/controllers/hardwareplugin_manager_setup.go
sakhoury: Agreed!

hwmgr-plugins/controller/hardwareplugin_controller.go
sakhoury: The client will only be used once for validation - hence static loader.

PR #709, opened 2025-06-03
Add Loopback HardwarePlugin server infrastructure

sakhoury:
# Summary

This PR adds the Loopback Hardware Plugin server infrastructure to `hwmgr-plugins`. Subsequent PRs will incorporate the API endpoint logic as well as `NodeAllocationRequest` handling.

/cc @browsell @alegacy @donpenney @tliu2021


internal/controllers/inventory_controller.go
```diff
@@ -865,6 +865,12 @@ func (t *reconcilerTask) run(ctx context.Context) (nextReconcile ctrl.Result, er
 		return
 	}
 
+	// Start the Loopback HardwarePlugin server
```
browsell, line 868: Since this is only a development tool we do not want to start this by default

PR #710, opened 2025-06-04
nonproduction: Update manager config to use postgresql-16-c9s image

clwheel:
Replaced the postgresql image with one that is publicly available in order to avoid requiring a pull secret in deployment automation being contributed to the ORAN-SC community.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -76,7 +76,7 @@ spec:
         - name: IMAGE
           value: controller:latest
         - name: POSTGRES_IMAGE
-          value: registry.redhat.io/rhel9/postgresql-16:9.5-1731610873
+          value: quay.io/sclorg/postgresql-16-c9s:c9s
         - name: HWMGR_PLUGIN_NAMESPACE
           # A placeholder for the replacement kustomization that will inject the value from the config map
           value: plugin-namespace-placeholder


pixelsoccupied:
Just a FYI we have [poc.0](https://github.com/openshift-kni/oran-o2ims/tree/poc.0) branch with this image and working well. But note that we actually explicitly set to use the `rhel9` flavour in the main branch back in the days (but dont remember the reasoning behind this). But IMO using `c9s` everywhere is good idea as its the the upstream of `rhel9` minus the creds requirements.

just fyi for @rauhersu ...not sure if you would need to match this change in midstream or rhel9 is expected there. 

/lgtm

rauhersu:
Folks, at the end of the day, we need an image present on registry.redhat.io to ship this operator to any customer, quay.io images are not allowed. You are free to use quay for development / testing but I need to create an overlay for that image in the productisation process.

Can anyone indicate me which postgres image should I use from the official catalog (just need a pointer to an stream) ? 
https://catalog.redhat.com/search?gs&q=postgresql+16&searchType=all

Cc: @alegacy @pixelsoccupied

config/manager/manager.yaml
```diff
@@ -76,7 +76,7 @@ spec:
         - name: IMAGE
           value: controller:latest
         - name: POSTGRES_IMAGE
-          value: registry.redhat.io/rhel9/postgresql-16:9.5-1731610873
+          value: quay.io/sclorg/postgresql-16-c9s:c9s
```
rauhersu, line 79: I believe we need 'make bundle' to map these changes on the CSV, and include the new CSV in the MR.
Cc: @alegacy @pixelsoccupied

alegacy: ya, looks like this branch isn't running our ci-job.  Not sure it needs to or not... cc: @donpenney

clwheel: @rauhersu Just to clarify, the `osc-l-release` branch is intended to be consumed by the [ORAN-SC OKD O-Cloud deployment automation](https://gerrit.o-ran-sc.org/r/gitweb?p=pti/rtp.git;a=tree;f=okd;hb=HEAD) that we have contributed upstream and not intended to be merged into the main branch. The reason for the quay.io image is that we don't want to force users of an open source project to obtain a pull secret.

donpenney: As @clwheel notes, this PR is in a non-production branch and will not be merged to main. @clwheel I'd suggest updating the title of the PR with a prefix to note this, for this and future PRs, to avoid confusion. Maybe something like "nonproduction" or "POC", like:
nonproduction: Update manager config to use ...

donpenney (approved):

PR #711, opened 2025-06-05
Handle NodeAllocationRequest creation / updation / deletion for the Loopback HardwarePlugin

sakhoury:
# Summary

This PR adds the functionality to handle the creation, updation and deletion of `NodeAllocationRequest` objects for the Loopback HardwarePlugin.

/cc @browsell @alegacy @donpenney @tliu2021

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ ENVTEST_VERSION = release-0.19
 OCLOUD_MANAGER_NAMESPACE ?= oran-o2ims
 
 # HWMGR_PLUGIN_NAMESPACE refers to the namespace of the hardware manager plugin.
-HWMGR_PLUGIN_NAMESPACE ?= oran-hwmgr-plugin
+HWMGR_PLUGIN_NAMESPACE ?= oran-o2ims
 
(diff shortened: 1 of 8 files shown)


hwmgr-plugins/controller/utils/utils.go
```diff
@@ -9,15 +9,86 @@ package utils
...
 )
 
 // Resource operations
 
+const (
+	UPDATE = "Update"
+	PATCH  = "Patch"
+)
+
+// CreateOrUpdateK8sCR creates/updates/patches an object.
```
browsell, line 30: Duplicate of https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/utils/utils.go#L81 ?

sakhoury: Nice find! I had searched the IMS repo and found [`CreateOrUpdate`](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/service/common/clients/k8s/k8s.go#L131) - but this does not support the `patch` operation. [This](https://github.com/openshift-kni/oran-o2ims/blob/main/internal/controllers/utils/utils.go#L81) however does! The name `CreateK8sCR` is a bit misleading. 

I've updated the code to use that!

PR #743, opened 2025-07-03
Merge plugin-poc content to main

donpenney:
Cherry-picked commits from plugin-poc to latest on main for merge


donpenney:
/override ci/prow/scorecard

The scorecard test is broken in this update due to restructuring and renaming of CRDs, but passes when manually run on the PR. The failure occurs because the upstream build of the operator expects a CRD that no longer exists

PR #789, opened 2025-07-31
fix dependabot Prow error...

rauhersu:
"The property '#/updates/0/ignore' of type null did not match the following type: array"

This stops our automatic konflux merges.


donpenney (approved):

donpenney:
/override ci/prow/ci-bundle-operator-bundle ci/prow/ci-job ci/prow/images

PR #820, opened 2025-08-07
split internal and external alarms apis with overlay

pixelsoccupied:
/cc @alegacy


internal/service/alarms/api/overlay.yaml
```diff
@@ -0,0 +1,53 @@
...
+info:
+  title: Internal Endpoints Overlay
+  version: 1.0.0
+actions:
+  - target: "$.paths"
+    update:
+      /internal/v1/caas-alerts/alertmanager:
+        post:
+          operationId: AmNotification
+          summary: Receive Alertmanager notifications
```
donpenney, line 11: curiousity nit: why isn't this camelcase, ie. AlertManager, throughout, such as the schema defs (line 20 below)? Not asking for it to be changed, if it's intentional, or that it should be changed as part of this PR since it's pre-existing and unrelated to this PR

donpenney: Yeah, it's absolutely unrelated to this PR. It just struck me as odd.

PR #909, opened 2025-08-20
Fix golangci lint version

alegacy:
+ Fix golangci version to be compatible with go1.24
+ Fix kubectl version


Makefile
```diff
@@ -399,8 +399,8 @@ golangci-lint: ## Run golangci-lint against code.
 		DOWNLOAD_GOLANGCI_LINT_VERSION=$(GOLANGCI_LINT_VERSION)
 	@echo "Golangci-lint downloaded successfully."
 	@echo "Running golangci-lint on repository go files..."
-	golangci-lint --version
-	golangci-lint run -v
+	$(PROJECT_DIR)/bin/golangci-lint --version
+	$(PROJECT_DIR)/bin/golangci-lint run -v
```
donpenney, line 403: I think we need to do the same for all the downloaded tools, like ocm, etc. Maybe set variables for the path+executable globally?

alegacy: I've changed it to follow the existing model for the other tools.   Didn't noticed we had those.

donpenney: Unfortunately, it looks like we're not actually using the variables everywhere, like the `opm version` call that isn't using the variable... line 486

alegacy: k... fixed that one and a couple of others.

donpenney (approved):

PR #946, opened 2025-08-26
Update project makefile to improve shared library script usage

fontivan:
- Define versions for all tools downloaded
- Make use of new download-go-tool script for several features
- Update telco5g-konflux subproject to latest commit
- Also fix some yamllint warnings

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,7 @@ SHELL = /usr/bin/env bash -o pipefail
 PROJECT_DIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 
 ## Location to install dependencies to
+# If you are setting this externally then you must use an aboslute path
 LOCALBIN ?= $(PROJECT_DIR)/bin
 $(LOCALBIN):
 	mkdir -p $(LOCALBIN)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -551,7 +599,7 @@ ci-job: deps-update go-generate generate fmt vet lint shellcheck bashate fmt tes
 
 .PHONY: clean
 clean:
-	-rm $(LOCALBIN)/$(BINARY_NAME)
+	rm -rf $(LOCALBIN)
```
donpenney, line 602: This is one spot where it feels dangerous to me. I wonder if this should have an if/else so that it only removes the dir if `$(LOCALBIN)` is equal to `$(PROJECTDIR)/bin`, and just gives a warning otherwise.

fontivan: what about if I individually removed the binaries inside the folder instead?
e.g.
```
rm $(YQ)
rm $(YAMLLINT)
rm $(OPM)
... etc
```

Makefile
```diff
@@ -24,17 +24,35 @@ VERSION ?= 4.20.0
...
-# YQ_VERSION defines the yq version to download from GitHub releases.
-YQ_VERSION ?= v4.45.4
+# CONTROLLER_GEN_VERSION defines the controller-gen version to download from go modules.
+CONTROLLER_GEN_VERSION ?= v0.18.0
+
+# GOLANGCI_LINT_VERSION defines the golangci-lint version to download from GitHub releases.
+GOLANGCI_LINT_VERSION ?= v2.4.0
+
+# KUSTOMIZE_VERSION defines the kustomize version to download from go modules.
+KUSTOMIZE_VERSION ?= v5@v5.7.1
```
rauhersu, line 37: Thinking out loud, is this a valid version specification ? 

The original one was defined here: https://github.com/openshift-kni/oran-o2ims/pull/946/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L265

donpenney (approved):

PR #962, opened 2025-08-28
Makefile robustness updates

fontivan:
Some additional updates to improve the robustness of the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ $(LOCALBIN):
 
 # This allows all tools in the LOCALBIN folder, ie:opm,yq ... To be used by targets containing scripts
 # Prefer binaries in the local bin directory over system binaries.
-export PATH  := $(LOCALBIN):$(PATH)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -696,13 +699,13 @@ konflux-validate-catalog-template-bundle: yq operator-sdk
 		ENGINE=$(ENGINE)
 
 .PHONY: konflux-validate-catalog
-konflux-validate-catalog: opm ## validate the current catalog file
+konflux-validate-catalog: opm sync-git-submodules ## validate the current catalog file
```
donpenney, line 702: `sync-git-submodules` should be the first in any target list, I think, right?

fontivan: I did not consider that the order mattered but I think you make a good point here

donpenney (approved):

donpenney:
/override ci/prow/ci-bundle-operator-bundle ci/prow/images ci/prow/ci-job

PR #1054, opened 2025-09-17
Workaround to deal with stuck ObservabilityAddon finalizer

browsell:
Generated-By: Cursor/claude-4-sonnet

### DIFF
--- a/.konflux/overlay/pin_images.in.yaml
+++ b/.konflux/overlay/pin_images.in.yaml
@@ -2,7 +2,7 @@
 ---
 - key: manager
   source: quay.io/openshift-kni/oran-o2ims-operator:4.21.0
-  target: quay.io/redhat-user-workloads/telco-5g-tenant/o-cloud-manager-4-21@sha256:1ec0b12194d147544ac2ce7359bac5d7231c55675e840a63cf754636ce412718
+  target: quay.io/redhat-user-workloads/telco-5g-tenant/o-cloud-manager-4-21@sha256:b1f81dd7bf60146903edbdc9c048ebb0a580cc9c58f9251762d0375a289ca97e
 - key: postgres_image
   source: registry.redhat.io/rhel9/postgresql-16:9.5-1731610873
   target: registry.redhat.io/rhel9/postgresql-16@sha256:4f46bed6bce211be83c110a3452bd3f151a1e8ab150c58f2a02c56e9cc83db98


internal/controllers/provisioningrequest_controller.go
```diff
@@ -1189,6 +1200,52 @@ func (r *ProvisioningRequestReconciler) handleProvisioningRequestDeletion(
...
+
+	for _, addon := range observabilityAddonList.Items {
+		if len(addon.Finalizers) > 0 {
+			r.Logger.InfoContext(ctx, "Removing finalizers from ObservabilityAddon",
+				slog.String("name", addon.Name),
+				slog.String("namespace", addon.Namespace),
+				slog.Any("finalizers", addon.Finalizers))
+
+			// Remove all finalizers
+			addon.Finalizers = []string{}
```
pixelsoccupied, line 1230: Shouldn't this workaround be more targeted to the problematic finalizer instead of dropping everything?

internal/controllers/provisioningrequest_controller.go
```diff
@@ -1172,6 +1174,15 @@ func (r *ProvisioningRequestReconciler) handleProvisioningRequestDeletion(
...
 	}
 	for _, ns := range namespaceList.Items {
+		// Workaround to deal with stuck ObservabilityAddon finalizer
+		// Check for ObservabilityAddon CR and remove finalizers before deleting namespace
+		if err := r.handleObservabilityAddonCleanup(ctx, ns.Name); err != nil {
+			r.Logger.WarnContext(ctx, "Failed to handle ObservabilityAddon cleanup",
+				slog.String("namespace", ns.Name),
+				slog.String("error", err.Error()))
+			// Continue with deletion attempt even if cleanup fails
+		}
```
pixelsoccupied, line 1184: i think here a little more safeguard would make sense instead of brute forcing this at every requeue...something like....
```
 // Only remove finalizers if deletion timestamp is set and older than threshold
  if ns.DeletionTimestamp.IsZero() &&
     time.Since(ns.DeletionTimestamp.Time) > time.Minute*5 {
      // Safe to remove stuck finalizers
  }
```

internal/controllers/provisioningrequest_controller.go
```diff
@@ -1189,6 +1200,52 @@ func (r *ProvisioningRequestReconciler) handleProvisioningRequestDeletion(
 	return deleteCompleted, nil
 }
 
+// handleObservabilityAddonCleanup checks for ObservabilityAddon CRs in the given namespace
+// and removes their finalizers to allow proper namespace deletion
```
pixelsoccupied, line 1204: consider adding a note or todo to understand the reasoning behind this workaround and when to remove this.

taking over finalizer from some other component might have unintended consequences right? resource leaks etc...so should make sure this has a deadline.

PR #1144, opened 2025-09-29
OCPBUGS-61948: Propagate failed condition status to provisioning state and detail

sakhoury:
# Summary

This PR propagates the failed condition status with the corresponding error message to the provisioningState and provisioningDetails.

Problem: ProvisioningRequests would get stuck in a progressing state during hardware provisioning failures because the failure detection logic wasn't properly checking hardware-specific conditions.

Solution: Enhanced the condition failure detection to include hardware provisioning and configuration conditions, and improved the logic to only treat conditions as failed when they have explicit Failed or TimedOut reasons. Additionally, added reason mapping between hardware management conditions and provisioning request conditions to ensure consistent state transitions across all failure scenarios.

ℹ️ Assisted-by: Cursor and claude-4-sonnet

Resolves: [OCPBUGS-61948](https://issues.redhat.com/browse/OCPBUGS-61948)

/cc @donpenney @tliu2021 @irinamihai @Missxiaoguo

### DIFF
--- a/internal/controllers/provisioningrequest_controller.go
+++ b/internal/controllers/provisioningrequest_controller.go
@@ -709,7 +709,7 @@ func (t *provisioningRequestReconcilerTask) checkClusterDeployConfigState(ctx co
 
 	// Check the ClusterInstance status if exists
 	if t.object.Status.Extensions.ClusterDetails == nil {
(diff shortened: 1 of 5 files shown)


Missxiaoguo (commented):
/lgtm
/approve

PR #1171, opened 2025-10-01
Clear managedFields when reusing ClusterInstance for SSA

sakhoury:
# Summary

When `canSkipClusterInstanceRendering` returns an existing ClusterInstance, certain `metadata` fields need to be cleared to avoid Server-Side Apply errors. Without this fix, PRs could remain stuck in the "progressing" state even after cluster provisioning has completed.

/cc @donpenney @Missxiaoguo


sakhoury:
/hold
for testing

Missxiaoguo:
Tested on my side too. It worked well!

PR #1184, opened 2025-10-02
Update catalog-deploy and catalog-undeploy Makefile targets and helpers

donpenney:


Makefile
```diff
@@ -24,6 +24,8 @@ VERSION ?= 4.21.0
 
 PACKAGE_NAME ?= oran-o2ims
 
+CHANNEL ?= alpha
```
rauhersu, line 27: Just a caveat: this operator does not have an alpha channel. Only 'stable' and '4.20' : https://github.com/openshift-kni/oran-o2ims/blob/release-4.20/.konflux/bundle/metadata/annotations.yaml#L7

PR #1185, opened 2025-10-02
OCPBUGS-61999: Preserve NAR error and context messages in PR status

sakhoury:
# Summary

Previously, `processExistingHardwareCondition` replaced detailed messages from the NodeAllocationRequest (NAR) with generic ones like "Hardware provisioning failed" or "Hardware provisioning is in progress", causing valuable error context and progress details to be lost.

This change ensures both error details and progress context from the NAR are reflected in the ProvisioningRequest, improving debugging and monitoring.

ℹ️ Assisted-by: Cursor and claude-4.5-sonnet

Resolves: [OCPBUGS-61999](https://issues.redhat.com/browse/OCPBUGS-61999)

/cc @donpenney @Missxiaoguo @tliu2021 @irinamihai


internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -501,17 +501,21 @@ func (t *provisioningRequestReconcilerTask) processExistingHardwareCondition(
...
-		message = fmt.Sprintf("Hardware %s is in progress", ctlrutils.GetStatusMessage(condition))
-		ctlrutils.SetProvisioningStateInProgress(t.object, message)
-
 		if reason == string(hwmgmtv1alpha1.Failed) || reason == string(hwmgmtv1alpha1.TimedOut) {
 			timedOutOrFailed = true
-			if reason == string(hwmgmtv1alpha1.TimedOut) {
-				message = fmt.Sprintf("Hardware %s timed out", ctlrutils.GetStatusMessage(condition))
-			} else {
-				message = fmt.Sprintf("Hardware %s failed", ctlrutils.GetStatusMessage(condition))
+			if reason == string(hwmgmtv1alpha1.Failed) {
```
Missxiaoguo, line 506: It seems like we removed the timeout check because we can just use whatever HandleHardwareTimeout returns. At first glance, it doesn’t feel very intuitive 🤔 since the message is being handled both in HandleHardwareTimeout and here. I think it would be clearer if everything were set in one place. Just my two cents—it’s really not a big deal!

Missxiaoguo (commented):
/lgtm

internal/controllers/provisioningrequest_hwprovision.go
sakhoury: Good point! The split is intentional for separation of concerns: HandleHardwareTimeout produces reusable, generic timeout messages, while the logic here adds the NAR context. For timeouts, we keep the original wording. Pulling NAR logic into HandleHardwareTimeout would couple things unnecessarily, and previously we were just re-assigning the same timeout/failed message. This version keeps responsibilities clear and avoids redundancy. I will add a comment to explain this.

PR #1338, opened 2025-10-31
CNF-20276: use BMH uid for hardware resources

pixelsoccupied:
Changed resource collection to use BareMetalHost UID directly instead of generating new UUIDs, enabling direct correlation between alarms and physical infrastructure.

tested with new cluster installation 

```
# agent.yaml
clcm.openshift.io/hardwarePluginRef: metal3-hwplugin
clcm.openshift.io/hwMgrNodeId: 9be672ba-3f59-4f3b-ba92-e04e32a11b59
```

### DIFF
--- a/internal/service/resources/collector/collector_hwplugin.go
+++ b/internal/service/resources/collector/collector_hwplugin.go
@@ -155,7 +155,13 @@ func (d *HwPluginDataSource) GetResources(ctx context.Context, _ []models.Resour
 
 	resources := make([]models.Resource, 0)
 	for _, resource := range *result.JSON200 {
-		resources = append(resources, *d.convertResource(&resource))
+		converted, err := d.convertResource(&resource)
+		if err != nil {
+			// Log error but continue processing other resources
+			slog.Error("Skipping resource due to conversion error", "error", err, "resourceId", resource.ResourceId)
+			continue
+		}
+		resources = append(resources, *converted)
 	}
 
 	return resources, nil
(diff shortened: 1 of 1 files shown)


internal/service/resources/collector/collector_hwplugin.go
```diff
@@ -199,15 +199,14 @@ func (d *HwPluginDataSource) convertResourcePool(pool *inventoryclient.ResourceP
...
-	return ctlrutils.MakeUUIDFromNames(ResourceUUIDNamespace, cloudID, hwPluginRef, hwMgrNodeID)
-}
-
 func (d *HwPluginDataSource) convertResource(resource *inventoryclient.ResourceInfo) *models.Resource {
-	// The resourceID computed here must
-	resourceID := MakeResourceID(d.cloudID, d.hwplugin.Name, resource.ResourceId)
+	// Parse the resource ID directly (now it's the BMH UID from the hardware plugin)
+	resourceID, err := uuid.Parse(resource.ResourceId)
+	if err != nil {
+		slog.Warn("Failed to parse resource ID as UUID, using nil UUID", "resourceId", resource.ResourceId, "error", err)
```
alegacy: this should be an error.  not being able to identify the resource by its id value is critical.

alegacy: I agree it should never happen, but we should error out here if it does.  Sending bad data up to the client is not going to be helpful.

PR #2217, opened 2026-04-02
feat: speed up PR deletion with force-detach and propagate skip-cleanup from PR to NAR

donpenney:
## Summary

- Set `hubAcceptsClient=false` on the ManagedCluster during ProvisioningRequest
  deletion to trigger ACM's force-detach path, reducing deletion time from ~8
  minutes to ~3 minutes by avoiding the graceful detach timeout when the spoke
  is powered off.

- Add `SkipCleanup` field to the NAR and AllocatedNode specs. When the user sets
  the `clcm.openshift.io/skip-cleanup` annotation on the ProvisioningRequest,
  it is propagated through: PR → NAR → AllocatedNodes → deallocation logic.
  This replaces the previous approach of annotating the BMH directly.

- When skip-cleanup is set, `hubAcceptsClient` is left as `true` so ACM can
  gracefully detach while the spoke remains running (e.g., during seed image
  generation for IBI/IBU).

- Update IBI user guide to reflect that the skip-cleanup annotation is now set
  on the ProvisioningRequest instead of the BMH.


donpenney:
[Comment authored by Claude]

Addressed CodeRabbit review from review 4059826932:

1. **StateAvailable wait with skipCleanup** — Fixed. The `isNodeProvisioningInProgress` wait is now skipped when `skipCleanup=true`, preventing a potential infinite loop since the BMH may never reach Available when cleanup is skipped.

2. **Test for externally provisioned BMH with skipCleanup** — Added. New test verifies that `finalizeBMHDeallocation` with `skipCleanup=true` on an externally provisioned BMH skips cleaning, power-off, and IBI warning annotation.

PR #2378, opened 2026-04-21
Remove deprecated BootInterfaceLabel field from NAR CRD

donpenney:


donpenney:
Overriding long post-rebase tests, since the only change would be rebasing against merged konflux PRs
/override ci/prow/scorecard ci/prow/ci-job

PR #2384, opened 2026-04-21
docs: add FirmwareCatalog CRD design proposal

browsell:
Introduce a design document for a new FirmwareCatalog CRD that centralizes firmware image management. HardwareProfile firmware fields would reference catalog entries by name instead of embedding URL and version inline.

### DIFF
--- a/docs/proposals/firmware-catalog.md
+++ b/docs/proposals/firmware-catalog.md
@@ -1,0 +1,608 @@
+# FirmwareCatalog CRD
+
+```yaml
+title: firmware-catalog
+authors:
+  - @browsell
+reviewers:
+  - TBD
+approvers:
+  - TBD
+creation-date: 2026-04-21
+last-updated: 2026-04-21
+```
+
+## Table of Contents
+
+- [FirmwareCatalog CRD](#firmwarecatalog-crd)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Current State](#current-state)
+  - [Proposed Design](#proposed-design)
+    - [FirmwareCatalog Resource](#firmwarecatalog-resource)
+      - [API Definition](#api-definition)
+      - [Example CR](#example-cr)
+      - [Lifecycle](#lifecycle)
+    - [HardwareProfile Changes](#hardwareprofile-changes)
+      - [Updated API Definition](#updated-api-definition)
+      - [Example CR (Before)](#example-cr-before)
+      - [Example CR (After)](#example-cr-after)
+    - [Controller Changes](#controller-changes)
+      - [Catalog Resolution](#catalog-resolution)
+      - [Integration Points](#integration-points)
+    - [Validation](#validation)
+  - [Impact](#impact)
+    - [Breaking Changes](#breaking-changes)
+    - [Files to Create or Modify](#files-to-create-or-modify)
+  - [CR Relationships](#cr-relationships)
+
+## Summary
+
+Introduce a new `FirmwareCatalog` CRD that serves as a centralized registry of firmware
+images. Today, each HardwareProfile embeds firmware download URLs and version strings
+inline. This couples firmware image management to hardware profile definitions, causing
+duplication when multiple profiles use the same firmware and requiring profile recreation
+to change a URL.
+
+The FirmwareCatalog decouples these concerns: firmware images are defined once in a
+shared catalog, and HardwareProfiles reference catalog entries by name. This enables
+reuse across profiles and centralized management of firmware image metadata.
+
+## Goals
+
+- Define a single, shared catalog of firmware images that can be populated and updated
+  by users at runtime.
+- Allow HardwareProfiles to reference firmware images by name rather than embedding
+  URL and version inline.
+- Minimize changes to the existing controller reconciliation logic by resolving catalog
+  references into the same data structures the downstream code already consumes.
+
+## Non-Goals
+
+- Firmware image hosting or download management.
+- Automatic discovery or population of firmware images.
+- Multi-namespace or cluster-scoped catalog support.
+- Versioning or approval workflows for catalog changes.
+
+## Current State
+
+The `HardwareProfile` CR defines firmware targets as inline objects with `url` and
+`version` fields:
+
+```go
+type Firmware struct {
+    Version string `json:"version,omitempty"`
+    URL     string `json:"url,omitempty"`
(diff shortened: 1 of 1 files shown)


docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,480 @@
...
+A new namespaced CRD in the `clcm.openshift.io/v1alpha1` API group. The operator
+maintains a singleton instance with a fixed well-known name. Users populate and update
+the `spec.images` list at runtime.
+
+#### API Definition
+
+File: `api/hardwaremanagement/v1alpha1/firmwarecatalog_types.go`
+
+```go
+type FirmwareImage struct {
```
donpenney, line 116: Let's add an optional Description field, as well, so users can provide more info about the image like what NIC models it's for, etc

docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,480 @@
...
+
+type FirmwareCatalogSpec struct {
+    // Images is the set of firmware images available in this catalog.
+    // +optional
+    // +listType=map
+    // +listMapKey=name
+    Images []FirmwareImage `json:"images,omitempty"`
+}
+
+type FirmwareCatalogStatus struct {
```
donpenney, line 163: I wonder if we should have a list for validated or invalidated images, so if one particular image fails validation, it's easily checked programmatically

docs/proposals/firmware-catalog.md
browsell: added

docs/proposals/firmware-catalog.md
browsell: added

docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,509 @@
...
+it cannot resolve the entry name. This is by design — removing a firmware image that
+is in use is an error. No deletion protection mechanism is provided; users are
+responsible for ensuring entries are not in use before removing them.
+
+**Entry updates (URL or version change):** Modifying a catalog entry's URL or version
+does not automatically trigger firmware updates on nodes using that entry. Firmware is
+only applied when a HardwareProfile change is detected on a ProvisioningRequest (i.e.,
+the `hwProfile` field in the NodeGroupData is updated). To roll out updated firmware
+from a modified catalog entry, users should create a new catalog entry with a new name
+and update the HardwareProfile to reference it.
```
Missxiaoguo: I think this is a bit confusing - "from a modified catalog entry" implies editing an existing entry in place and "update the HardwareProfile" suggests modifying the existing hardwareprofile instead of creating a new one. It's inconsistent with the workflow  described above which requires updating the PR to point to the new hardware profile with a new name referencing the new image.

docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,509 @@
...
+
+### Catalog Update Behavior
+
+The FirmwareCatalog is resolved at HardwareProfile reconciliation time. Changes to the
+catalog do not automatically trigger reconciliation of HardwareProfiles that reference
+it. This has the following implications:
+
+**Entry deletion:** Deleting a catalog entry that is actively referenced by a
+HardwareProfile will cause the next reconciliation of that profile to fail. The
+controller will set the `Validation` condition to `Failed` on the HardwareProfile when
```
Missxiaoguo: This is not clear to me.  Could we clarify which controller sets the Validation condition on the HarewareProfile? Currently, there is not hardware profile controller. Would the NodeAllocationRequest controller set this condition everytime when it fetches the firmware data? or a different controller

docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,509 @@
...
+
+**Entry updates (URL or version change):** Modifying a catalog entry's URL or version
+does not automatically trigger firmware updates on nodes using that entry. Firmware is
+only applied when a HardwareProfile change is detected on a ProvisioningRequest (i.e.,
+the `hwProfile` field in the NodeGroupData is updated). To roll out updated firmware
+from a modified catalog entry, users should create a new catalog entry with a new name
+and update the HardwareProfile to reference it.
+
+**No FirmwareCatalog watch:** The HardwareProfile controller does not watch the
+FirmwareCatalog resource. It reads the catalog on demand during reconciliation. This
```
Missxiaoguo: Same comment as above - no hardware profile controller today. The expected behaviour is not very clear.

docs/proposals/firmware-catalog.md
```diff
@@ -0,0 +1,509 @@
...
+
+**Entry deletion:** Deleting a catalog entry that is actively referenced by a
+HardwareProfile will cause the next reconciliation of that profile to fail. The
+controller will set the `Validation` condition to `Failed` on the HardwareProfile when
+it cannot resolve the entry name. This is by design — removing a firmware image that
+is in use is an error. No deletion protection mechanism is provided; users are
+responsible for ensuring entries are not in use before removing them.
+
+**Entry updates (URL or version change):** Modifying a catalog entry's URL or version
+does not automatically trigger firmware updates on nodes using that entry. Firmware is
```
Missxiaoguo: I have a concern about modifying the existing catalog entry.  If a node was provisioned with a HardwareProfile that references  a catalog entry, and someone later updates that entry, the HardwareProfile still points to the same entry name but now silently resolves to a different version than what was originally applied. 
This also affects the PR: it would appear fulfilled with a HardwareProfile whose resolved firmware doesn't match what was applied.

Missxiaoguo: The above mentions `resolveFirmwareFromCatalog` is called at the start of each code path, another problem is if a catalog entry is modified while a firmware update is in progress, the post-validation step (`validateFirmwareVersions`) would resolve the new version from the catalog and compare it against the running firmware, which was successfully updated to the original version. This would cause a false validation failure even though the update succeeded.

I think we should really think about if updating the existing entry is allowed or not

docs/proposals/firmware-catalog.md
browsell: updated

docs/proposals/firmware-catalog.md
browsell: updated

docs/proposals/firmware-catalog.md
browsell: updated

docs/proposals/firmware-catalog.md
browsell: updated

PR #2401, opened 2026-04-23
docs: add NAR API removal proposal

donpenney:
## Summary
- Add proposal document for removing the internal REST API layer between the ProvisioningRequest controller and NodeAllocationRequest CRs
- 5-phase plan: direct K8s client access, callback removal, generated code cleanup, server removal, and renaming
- Includes metadata header, post-Phase 1 simplification opportunities (HardwarePluginLabel removal, mock server removal), and risk mitigations

### DIFF
--- a/docs/proposals/nar-api-removal-proposal.md
+++ b/docs/proposals/nar-api-removal-proposal.md
@@ -154,7 +154,7 @@ operations on NAR CRs:
   map from the ClusterTemplate's `hwMgmtDefaults`), so no HardwareTemplate
   lookup is needed.
 - Rewrite `createOrUpdateNodeAllocationRequest` to use `client.Create` and
-  `client.Update`
+  `client.Patch`
 - Rewrite `getNodeAllocationRequestResponse` to use `client.Get` and return
   the CRD type directly
 - Rewrite `setNARClusterProvisioned` and `syncNARSkipCleanup` to use
@@ -291,4 +291,4 @@ the complexity reduction justifies the churn.
 | Large change scope | Phased approach allows incremental delivery and testing |
 | Test coverage gaps | Each phase includes test refactoring. Phase 1 replaces mock REST with fake K8s clients. |
 | Phase 5 is a breaking change | Align with a release boundary that already includes breaking changes |
-| RBAC changes for PR controller | Straightforward — add NAR CR permissions to the operator's ClusterRole |
+| RBAC changes for PR controller | Already covered — existing kubebuilder markers provide the necessary permissions |
--- a/docs/proposals/nar-api-removal-proposal.md
+++ b/docs/proposals/nar-api-removal-proposal.md
@@ -6,6 +6,18 @@ SPDX-License-Identifier: Apache-2.0
 
 # Proposal: Remove NAR REST API and Simplify Hardware Manager Architecture
 
+```yaml
+title: nar-api-removal
+authors:
+  - @dpenney
+reviewers:
+  - TBD
+approvers:
+  - TBD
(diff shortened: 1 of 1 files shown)


docs/proposals/nar-api-removal-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+- **Callback complexity**: The NAR controller must maintain a separate HTTP
+  client to notify the PR controller of status changes, rather than simply
+  updating the CR and letting the watch mechanism trigger reconciliation.
+- **Test complexity**: Tests must mock the REST interface rather than using
+  standard fake Kubernetes clients.
+
+### Benefits of Direct CR Access
+
+- **Uniform field handling**: `client.Patch` handles all fields atomically — no
+  separate create/update paths to maintain.
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch. The implementation does use `client.Patch` (not `client.Update`) for NAR updates, consistent with the atomicity goal. Fixed the doc to say `client.Create` and `client.Patch` in d7f857b.

docs/proposals/nar-api-removal-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+  interfaces
+- The generated REST client/server code stays in the tree (unused) until
+  Phase 3
+- The callback mechanism stays until Phase 2
+
+**RBAC changes**: The PR controller already has the necessary RBAC
+permissions for NAR CRs via existing kubebuilder markers (`get`, `list`,
+`watch`, `create`, `update`, `patch`, `delete` on
+`nodeallocationrequests.plugins.clcm.openshift.io`). No RBAC changes
+needed.
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch on the contradiction. The PR controller already has the necessary RBAC via existing kubebuilder markers. Updated the risk table to say "Already covered" instead of "add permissions", consistent with the Phase 1 section. Fixed in d7f857b.

docs/proposals/nar-api-removal-proposal.md
```diff
@@ -0,0 +1,294 @@
...
+controller and the hardware manager:
+
+```text
+PR Controller  ──REST──►  hardwareplugin-manager-server  ──CR──►  NAR CR
+                                                                    │
+PR Controller  ◄──REST──  hardwareplugin-manager-server  ◄──REST──  NAR Controller
+                 (callback)
+```
+
+This was originally designed to support third-party hardware plugins via a
```
Missxiaoguo, line 42: I think the main motivation here is that we no longer plan to support third-party hardware plugins, and no third-party plugin exists today? Otherwise, the REST API layer would still be necessary

donpenney: Correct. We're shifting to only having the single metal3-hwmgr. The PR controller would then be able to access the CRs managed by the metal3-hwmgr directly, and vice versa - ie. the PR controller could have a watch on the NAR rather than polling or callbacks, the metal3-hwmgr could see the status of the PR (if it had any reason to)... the PR controller could look at the BMH to see it is externallyProvisioned, if it needed to know the cluster was provisioned with IBI.

The PR controller would no longer go through the REST API for creating, updating, deleting the NAR, and the metal3-hwmgr would not use the callback mechanism for responses

Missxiaoguo: yeah, removing the REST API layer is a definitly significant simplication with the single metal3-hwmgr.

PR #2472, opened 2026-05-04
Add design decisions to AGENTS.md for AI code review tools

donpenney:
## Summary
- Inline accepted design decisions (DD-001, DD-002) directly in `AGENTS.md`
  so both CodeRabbit (auto-detects AGENTS.md) and Claude Code (imports via
  CLAUDE.md) apply them during code reviews
- Single source of truth — no separate design-decisions doc to maintain

## Background
CodeRabbit was repeatedly flagging two accepted patterns:
1. In-memory AllocatedNode filtering (DD-001)
2. Namespace-less NamespacedName for cluster-scoped CRs (DD-002)

After discussion with CodeRabbit on this PR, we confirmed that:
- CodeRabbit auto-detects `AGENTS.md` and uses it as review guidelines
- Inlining the rules in `AGENTS.md` serves all consumers (CodeRabbit,
  Claude Code, human readers) from a single source of truth
- `path_instructions` does not support `@filename` imports, so the
  original approach was non-functional

## Test plan
- [ ] Verify CodeRabbit no longer flags DD-001/DD-002 patterns in future PRs
- [ ] Verify markdownlint passes: `make markdownlint`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,6 +83,25 @@ This project has not reached GA, so there are no production databases to migrate
 - `ocloud.openshift.io` - Inventory
 - `plugins.clcm.openshift.io` - NodeAllocationRequest, AllocatedNode
 
+### Design Decisions
+
+The following are accepted design decisions. Do not flag these patterns as
+issues during code review.
+
+- **DD-001: In-memory filtering for AllocatedNode lookups.**
+  `listAllocatedNodesForNAR` in the PR controller uses in-memory filtering
+  instead of server-side `MatchingFields` because the fake Kubernetes client
+  in unit tests does not support field selectors. Per-cluster node counts are
+  small (1-11 nodes), so the cost is negligible. The Metal3 NAR controller
+  uses a proper field index (`spec.nodeAllocationRequest`) since it runs with
+  a real manager cache.
+
+- **DD-002: Cluster-scoped ProvisioningRequest watch mappers.**
+  Watch mappers that enqueue ProvisioningRequests (e.g.,
+  `enqueueProvisioningRequestForNAR`) intentionally omit the namespace from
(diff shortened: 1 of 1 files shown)


donpenney:
@coderabbitai help

donpenney:
@coderabbitai The purpose of the `.coderabbit.yaml` configuration change is to provide decision decisions that resolve repeated code review comments from you so we can avoid further repeats, since we have "knowledge base" disabled currently. If we had an architectural decisions folder (ie. `docs/adr`) instead of putting this in the config file, would that also avoid future review comments? Do you check for such architecture decision records when performing reviews?

donpenney:
@coderabbitai Can you verify for me that the new approach in this PR is valid, that you are now reading the design decisions from design-decisions.md and using the rules we have in place there to affect the review comments you make?

donpenney:
@coderabbitai I'm trying to avoid duplicating these rules in multiple files. Is there no way to encode in `.coderabbit.yaml` an instruction to import rules from a specified markdown file? You also said that you read from CLAUDE.md. What if we added these rules to AGENTS.md, which is imported by CLAUDE.md?

alegacy:
@coderabbitai what if we did a `@docs/dev/design-decisions.md` from either the AGENTS.md or CLAUDE.md... would that effectively import/include the context into what you'll consider during reviews?

donpenney:
@coderabbitai Please verify that you will now read these design decision rules from `AGENTS.md` when you start processing a PR and will not add repeat comments based on the "In-memory filtering for AllocatedNode lookups", etc. For example, the comment you added here, I would not expect to see again based on these rules: https://github.com/openshift-kni/oran-o2ims/pull/2369#discussion_r3182493353

PR #2501, opened 2026-05-08
docs: add resource collector improvements proposal

donpenney:
Propose replacing the 60-second polling loop in the hardware data source with event-driven K8s watches, aligning it with the pattern already used by the other four data sources (Location, OCloudSite, ResourcePool, DeploymentManager). Also covers data change event table cleanup and other operational improvements.

### DIFF
--- a/docs/proposals/resource-collector-improvements.md
+++ b/docs/proposals/resource-collector-improvements.md
@@ -146,28 +146,39 @@ across three CRD types (BMH + HardwareData + AllocatedNode) to build a
 complete `ResourceInfo`. With watches, a change to any one of these three
 CRDs needs to trigger a rebuild of the affected resource.
 
-The three CRDs are linked by naming conventions:
+The CRDs are linked as follows:
 
 - **BMH ↔ HardwareData**: HardwareData has the same name and namespace as
   the corresponding BMH (1:1 relationship, created by the Metal3 Bare Metal
-  Operator).
+  Operator). HardwareData is currently not updated after creation (only
+  deleted and recreated on re-inspection), but we should still watch it
+  in case future BMO changes introduce updates.
 - **BMH ↔ AllocatedNode**: AllocatedNode references its BMH via
   `spec.hwMgrNodeId` (BMH name) and `spec.hwMgrNodeNs` (BMH namespace).
+  Additionally, the hardware manager attaches labels to the BMH when it is
+  allocated: `clcm.openshift.io/allocated: "true"` and
+  `clcm.openshift.io/allocated-node: <allocatedNode name>`. These labels
+  can be used for reverse lookups.
 
 The join logic for watch events needs to handle:
 
-- BMH created/updated → look up HardwareData by same name/namespace,
-  look up AllocatedNode by `spec.hwMgrNodeId`/`spec.hwMgrNodeNs`
-- HardwareData updated → find corresponding BMH by same name/namespace
-- AllocatedNode created/deleted → find corresponding BMH via
-  `spec.hwMgrNodeId`/`spec.hwMgrNodeNs`, update allocation state
+- **BMH watch**: On BMH create/update, look up HardwareData by same
+  name/namespace (for hardware details) and check allocation labels for
+  AllocatedNode association. BMH updates are infrequent in practice.
+- **HardwareData watch**: On HardwareData create (or delete/recreate during
+  re-inspection), find the corresponding BMH by same name/namespace and
+  rebuild the resource. Currently rare but future-proofs against BMO changes.
+- **AllocatedNode watch**: On AllocatedNode create/delete, find the
+  corresponding BMH via `spec.hwMgrNodeId`/`spec.hwMgrNodeNs` and update
+  the resource's allocation state. AllocatedNode is only updated during
+  day-2 operations or cluster deprovisioning.
 
 **Approach options:**
 
 1. **Three separate watches with cross-reference**: Watch each CRD
    independently. On any event, look up the related objects and rebuild the
-   full ResourceInfo. Simple to implement but may cause duplicate database
-   writes when multiple related CRDs change simultaneously.
+   full ResourceInfo. Simple to implement and consistent with the existing
+   Reflector pattern.
 
 2. **Single composite watch with debouncing**: Watch all three CRDs, queue
    events keyed by BMH name/namespace, and debounce with a short delay
@@ -176,18 +187,19 @@ The join logic for watch events needs to handle:
 
 3. **BMH-primary watch with lazy enrichment**: Watch only BMH changes.
    On each BMH event, fetch HardwareData and AllocatedNode data on demand.
-   Simplest approach but misses changes to HardwareData or AllocatedNode
-   that don't coincide with BMH changes.
+   Simplest approach but misses changes that don't coincide with BMH changes.
 
 **Recommendation:** Option 1 (three separate watches) is the simplest
-starting point and consistent with the existing Reflector pattern. The
-cross-reference lookups are cheap (single Get by name), and duplicate writes
-are handled by the existing change detection in `PersistObjectWithChangeEvent`.
+starting point. Updates are infrequent in practice (HardwareData is not
+currently updated after creation, AllocatedNode only changes during day-2
+or deprovisioning), and
+duplicate writes are handled by the existing change detection in
+`PersistObjectWithChangeEvent`.
 
 **Impact on polling loop:** Once HardwareDataSource implements
 `WatchableDataSource`, the `execute()` polling loop will have no
-`ResourceDataSource` instances to process. It can be simplified or
-removed, with the `pollingDelay` timer replaced by watch-driven events.
+`ResourceDataSource` instances to process and should be removed as dead
+code, along with the `pollingDelay` timer.
 
 ### Phase 2: Data Change Event Cleanup
 
--- a/docs/proposals/resource-collector-improvements.md
+++ b/docs/proposals/resource-collector-improvements.md
@@ -221,9 +221,6 @@ Add TTL-based cleanup for the `data_change_event` outbox table:
 
 These are lower-priority items that can be addressed independently:
 
-- **Configurable poll interval**: If polling is retained as a fallback,
-  expose the interval as an environment variable.
-
 - **API pagination**: The inventory REST API endpoints (e.g.,
   `GET /resourcePools/{poolId}/resources`) currently load all matching
   records from the database and serialize them into a single JSON response.
@@ -260,7 +257,7 @@ These are lower-priority items that can be addressed independently:
 | Watch disconnection loses events | K8s Reflector handles reconnection and re-list automatically |
 | Increased memory from watch caches | The collector's custom `ReflectorStore` uses an operation queue rather than a full object cache (unlike the standard client-go Informer cache). Memory usage is proportional to the queue depth, not the total object count. |
 | Cross-reference lookups add latency | Single Get-by-name is <1ms on the local API server |
-| Stale data if watch lags | Generation ID sweep can be retained as a periodic consistency check |
+| Stale data if watch lags | K8s Reflector automatically re-lists on reconnection, which catches any missed events. If additional protection is needed, an infrequent (e.g., hourly) consistency check could be added as a separate mechanism. |
 
 ## Alternatives Considered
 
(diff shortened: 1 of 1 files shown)


docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,255 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Proposal: Resource Collector Improvements
+
+```yaml
+title: resource-collector-improvements
+authors:
+  - @dpenney
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — updated to `@donpenney`.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,255 @@
...
+**Key design consideration:** The current `GetResources` function joins data
+across three CRD types (BMH + HardwareData + AllocatedNode) to build a
+complete `ResourceInfo`. With watches, a change to any one of these three
+CRDs needs to trigger a rebuild of the affected resource. The join logic
+needs to handle:
+
+- BMH created → look up HardwareData by name, check AllocatedNode mapping
+- HardwareData updated → find corresponding BMH, rebuild resource
+- AllocatedNode created/deleted → find corresponding BMH, update allocation
+  state
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good point. Added the mapping details: HardwareData uses the same name/namespace as the BMH (1:1, created by Metal3 BMO), and AllocatedNode references its BMH via `spec.hwMgrNodeId` (name) and `spec.hwMgrNodeNs` (namespace). Updated the watch event handling descriptions to reference these specific fields.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,255 @@
...
+`ResourceDataSource` instances to process. It can be simplified or
+removed, with the `pollingDelay` timer replaced by watch-driven events.
+
+### Phase 2: Data Change Event Cleanup
+
+Add TTL-based cleanup for the `data_change_event` outbox table:
+
+- Delete events older than a configurable retention period (default 7 days)
+- Implement as a periodic cleanup in the collector's main loop
+- Ensure the notifier's `event_cursor` tracking is not affected
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Added the specific mechanism: only delete events whose `sequence_id` is older than the minimum `event_cursor` across all active subscriptions. If no subscriptions exist, all events older than the retention period can be safely deleted.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+
+Add TTL-based cleanup for the `data_change_event` outbox table:
+
+- Delete events older than a configurable retention period (default 7 days)
+- Implement as a periodic cleanup in the collector's main loop
+- To protect active subscribers, only delete events whose `sequence_id` is
+  older than the minimum `event_cursor` across all active subscriptions.
+  This ensures no subscriber misses events it has not yet processed. If no
+  subscriptions exist, all events older than the retention period can be
+  safely deleted.
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch. Added a stale subscription exclusion policy — subscriptions that have not advanced their cursor beyond a configurable staleness timeout are excluded from the minimum cursor calculation to prevent indefinite blocking.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,269 @@
...
+  resource types) to a single in-memory lookup for the common case where
+  nothing has changed.
+
+## Risks and Mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Watch disconnection loses events | K8s Reflector handles reconnection and re-list automatically |
+| Increased memory from watch caches | Reflector does not cache full objects — it uses a queue |
+| Cross-reference lookups add latency | Single Get-by-name is <1ms on the local API server |
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Clarified. The collector uses a custom `ReflectorStore` (not the standard client-go Informer cache) that stores an operation queue rather than full object copies. Memory usage is proportional to queue depth, not total object count. Updated the wording to reflect this distinction.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,273 @@
...
+  the corresponding BMH (1:1 relationship, created by the Metal3 Bare Metal
+  Operator).
+- **BMH ↔ AllocatedNode**: AllocatedNode references its BMH via
+  `spec.hwMgrNodeId` (BMH name) and `spec.hwMgrNodeNs` (BMH namespace).
+
+The join logic for watch events needs to handle:
+
+- BMH created/updated → look up HardwareData by same name/namespace,
+  look up AllocatedNode by `spec.hwMgrNodeId`/`spec.hwMgrNodeNs`
+- HardwareData updated → find corresponding BMH by same name/namespace
```
donpenney: Currently, the HardwareData CR is never updated. It gets created when inspection is completed and is never updated. I asked Claude to look at the current BMO codebase to verify my understanding, and it replied with:

> Your understanding is correct. HardwareData is effectively immutable after creation.
>
>  Here's the summary:
>
>  - Created once — in actionInspecting() at baremetalhost_controller.go:1099, after inspection completes successfully. The Spec is populated with the HardwareDetails from the
>  inspection result.
>  - Never updated — the Spec is never modified after creation. The only "updates" are finalizer removals (metadata-only changes) during host deletion or re-inspection cleanup.
>  - Re-inspection uses delete + recreate — if a host is re-inspected while a HardwareData already exists, the controller deletes the old one (line 1088) and creates a fresh one
>  (line 1099) rather than updating it in place.
>  - Read-only after creation — reconcileHostData() (line 2341+) reads HardwareData to restore BareMetalHost status and infer architecture, but never writes to the Spec.
>
>  So HardwareData is designed as an immutable snapshot of the hardware discovered at inspection time.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,273 @@
...
+- **BMH ↔ HardwareData**: HardwareData has the same name and namespace as
+  the corresponding BMH (1:1 relationship, created by the Metal3 Bare Metal
+  Operator).
+- **BMH ↔ AllocatedNode**: AllocatedNode references its BMH via
+  `spec.hwMgrNodeId` (BMH name) and `spec.hwMgrNodeNs` (BMH namespace).
+
+The join logic for watch events needs to handle:
+
+- BMH created/updated → look up HardwareData by same name/namespace,
+  look up AllocatedNode by `spec.hwMgrNodeId`/`spec.hwMgrNodeNs`
```
donpenney: The Hardware Manager attaches labels to the BMH when it is allocated, so we could use these for lookups:
    clcm.openshift.io/allocated: "true"
    clcm.openshift.io/allocated-node: <allocatedNode name>

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,273 @@
...
+
+3. **BMH-primary watch with lazy enrichment**: Watch only BMH changes.
+   On each BMH event, fetch HardwareData and AllocatedNode data on demand.
+   Simplest approach but misses changes to HardwareData or AllocatedNode
+   that don't coincide with BMH changes.
+
+**Recommendation:** Option 1 (three separate watches) is the simplest
+starting point and consistent with the existing Reflector pattern. The
+cross-reference lookups are cheap (single Get by name), and duplicate writes
+are handled by the existing change detection in `PersistObjectWithChangeEvent`.
```
donpenney: I don't think updates will happen that often in practice. The HardwareData CR doesn't get updated. The AllocatedNode CR is only going to get updated if there's a day2 operation or the cluster is deprovisioned. And I don't think the BMH is likely to be updated that often.

Option 1 seems like a reasonable choice

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Impact on polling loop:** Once HardwareDataSource implements
+`WatchableDataSource`, the `execute()` polling loop will have no
+`ResourceDataSource` instances to process. It can be simplified or
+removed, with the `pollingDelay` timer replaced by watch-driven events.
```
donpenney: This polling loop should just be removed altogether as dead code once the change is made

docs/proposals/resource-collector-improvements.md
donpenney: [Comment authored by Claude]

Updated the proposal — noted that HardwareData is not currently updated after creation, but we'll still watch it to future-proof against BMO changes. Back to three watches (BMH + HardwareData + AllocatedNode).

docs/proposals/resource-collector-improvements.md
donpenney: [Comment authored by Claude]

Added to the proposal — noted the `clcm.openshift.io/allocated` and `clcm.openshift.io/allocated-node` labels on the BMH as available for reverse lookups.

docs/proposals/resource-collector-improvements.md
donpenney: [Comment authored by Claude]

Incorporated — updated the proposal to note that updates are infrequent in practice, confirming Option 1 as the right approach.

docs/proposals/resource-collector-improvements.md
donpenney: [Comment authored by Claude]

Updated — the proposal now says the polling loop should be removed as dead code, not just simplified.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,285 @@
...
+starting point. Updates are infrequent in practice (HardwareData is not
+currently updated after creation, AllocatedNode only changes during day-2
+or deprovisioning), and
+duplicate writes are handled by the existing change detection in
+`PersistObjectWithChangeEvent`.
+
+**Impact on polling loop:** Once HardwareDataSource implements
+`WatchableDataSource`, the `execute()` polling loop will have no
+`ResourceDataSource` instances to process and should be removed as dead
+code, along with the `pollingDelay` timer.
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed. Removed the "Configurable poll interval" item from Phase 3 (which referenced "if polling is retained as a fallback") and updated the risks table entry for "Stale data if watch lags" to describe the Reflector's built-in re-list on reconnection as the primary mitigation, with an optional infrequent consistency check as a separate mechanism if needed. The proposal is now consistent: the polling loop is removed entirely in Phase 1.

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Phase 2: Data Change Event Cleanup
+
+Add TTL-based cleanup for the `data_change_event` outbox table:
+
+- Delete events older than a configurable retention period (default 7 days)
+- Implement as a periodic cleanup in the collector's main loop
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Updated — replaced "collector's main loop" with "periodic goroutine (e.g., `time.Ticker`) in the collector, independent of the watch event processing." This makes it clear the TTL cleanup is a standalone mechanism, not coupled to the removed polling loop.

donpenney:
Rebase pushed to branch

docs/proposals/resource-collector-improvements.md
```diff
@@ -0,0 +1,283 @@
...
+DeploymentManager (ManagedCluster) use K8s Reflectors for event-driven
+updates. Changes are detected in <1 second and persisted via async events.
+
+**Polling-based (1 data source)**: HardwareDataSource does a full re-list of
+all BareMetalHost, HardwareData, and AllocatedNode CRs every 60 seconds.
+No incremental updates or deltas — every cycle re-fetches all resources and
+uses a generation ID mark-and-sweep to detect deletions.
+
+```text
+                    Watch-based (real-time)          Polling (60s interval)
```
rauhersu, line 49: There is at least this server-onboarding.md [section](https://github.com/openshift-kni/oran-o2ims/blob/main/docs/user-guide/server-onboarding.md#verify-via-api) referencing the 60s polling interval. Should we change this, we will need to update it as well.

donpenney: [Comment authored by Claude]

Good catch. Updated the proposal to note that user-facing documentation must be updated alongside the code change. The implementation PR (#2517) includes updates to both `server-onboarding.md` and `inventory-api.md` to remove the polling references.

PR #2505, opened 2026-05-08
CNF-23666: Go 1.25 Phase 1 — build config, linting, and tooling

donpenney:
## Summary

Phase 1 of the Go 1.25 upgrade ([CNF-23329](https://redhat.atlassian.net/browse/CNF-23329)). Build configuration, tooling
updates, and lint fixes only — no dependency or vendor changes.

## Changes

**Build images:**
- `.ci-operator.yaml`: rhel-9-release-golang-1.25-openshift-4.22
- `.konflux/Dockerfile`: openshift-golang-builder rhel_9_golang_1.25
- `Dockerfile`: golang:1.25

**Tooling:**
- `go.mod`: go 1.25.0
- `Makefile`: golangci-lint v2.4.0 → v2.12.2 (old version was built with
  Go 1.24 and cannot lint Go 1.25 code)
- `Makefile`: envtest K8s 1.32 → 1.35, setup-envtest release-0.22
- `hack/update_deps.sh`: read Go version from go.mod instead of hardcoding
  `PINNED_GO`
- `.github/dependabot.yml`: remove ignore list that blocked Go 1.25-dependent
  package updates (k8s.io, ocm, golang.org, pgx)

**Lint fixes for new golangci-lint rules:**
- `govet`: `reflect.Ptr` → `reflect.Pointer` (deprecated constant)
- `misspell`: `requiered` → `required`
- `gosec`: nolint for false-positive path traversal on constant file paths
- `goconst`: configure exclusions for test files, API server files, and
  common K8s/RBAC strings; use existing constants in crd-watcher; create
  `apiGroupCLCM`/`apiGroupMetal3` constants in controllers

## Test plan
- [x] `make ci-job` passes
- [x] Lab tested: operator deployment with Inventory CR update

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -2,4 +2,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.24-openshift-4.22
+  tag: rhel-9-release-golang-1.25-openshift-4.22
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -133,7 +133,7 @@ linters:
         - .WrapPrefixf(
     goconst:
       min-occurrences: 5
-      ignore-string-values: 'get|list|watch|create|update|patch|delete|app|name|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true'
+      ignore-string-values: '^(get|list|watch|create|update|patch|delete|app|name|namespace|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true|agent-install\.openshift\.io/clusterdeployment-namespace)$'
   exclusions:
     generated: lax
     presets:
@@ -145,7 +145,7 @@ linters:
       - linters: [goconst]
         path: _test\.go$
       - linters: [goconst]
-        path: test/
+        path: (^|/)test/
       - linters: [goconst]
         path: internal/service/.*/api/server\.go$
       - linters: [gosec]
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -131,6 +131,9 @@ linters:
         - .WithMessagef(
         - .WithStack(
         - .WrapPrefixf(
+    goconst:
+      min-occurrences: 5
+      ignore-string-values: 'get|list|watch|create|update|patch|delete|app|name|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true'
   exclusions:
     generated: lax
     presets:
@@ -138,6 +141,16 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
+      - linters: [goconst]
+        path: _test\.go$
+      - linters: [goconst]
+        path: test/
+      - linters: [goconst]
+        path: internal/service/.*/api/server\.go$
+      - linters: [gosec]
+        path: _test\.go$
+        text: G101
     paths:
       - third_party$
       - builtin$
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ BASHATE_VERSION ?= 2.1.1
 CONTROLLER_GEN_VERSION ?= v0.18.0
 
 # GOLANGCI_LINT_VERSION defines the golangci-lint version to download from GitHub releases.
-GOLANGCI_LINT_VERSION ?= v2.4.0
+GOLANGCI_LINT_VERSION ?= v2.12.2
 
 # KUSTOMIZE_VERSION defines the kustomize version to download from go modules.
 KUSTOMIZE_VERSION ?= v5@v5.7.1
@@ -128,8 +128,8 @@ endif
 IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.32.0
-ENVTEST_VERSION = release-0.21
+ENVTEST_K8S_VERSION = 1.35.0
+ENVTEST_VERSION = release-0.22
 
 # OCLOUD_MANAGER_NAMESPACE refers to the namespace of the O-Cloud Manager
 OCLOUD_MANAGER_NAMESPACE ?= oran-o2ims
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift-kni/oran-o2ims
 
-go 1.24.0
+go 1.25.0
 
 // Needed for importing the siteconfig operator, taken from the siteconfig operator repo.
 replace (
--- a/hack/update_deps.sh
+++ b/hack/update_deps.sh
@@ -7,12 +7,17 @@
 
 set -e
 
-PINNED_GO="1.24.0"
+# Read the Go version from go.mod rather than hardcoding it
+GO_VERSION=$(grep '^go ' go.mod | awk '{print $2}')
+if [ -z "${GO_VERSION}" ]; then
+    echo "Error: could not determine Go version from go.mod"
+    exit 1
+fi
 
 # Use of GOTOOLCHAIN=local will cause failure if the local go version doesn't
-# support the pinned go version, rather than automatically downloading a newer
+# support the go.mod go version, rather than automatically downloading a newer
 # go version
-GOFLAGS='' GOTOOLCHAIN=local go mod tidy -go="${PINNED_GO}"
+GOFLAGS='' GOTOOLCHAIN=local go mod tidy -go="${GO_VERSION}"
 GOFLAGS='' GOTOOLCHAIN=local go mod vendor
 
 if grep -q "^toolchain " go.mod; then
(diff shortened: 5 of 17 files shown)


donpenney:
[Comment authored by Claude]

Re: [review comment](https://github.com/openshift-kni/oran-o2ims/pull/2505#pullrequestreview-4254391092) on gosec path traversal in alarms API

Fixed. Added `filepath.Clean` on both env-overridden paths. The upgraded golangci-lint (v2.12.2) recognizes `filepath.Clean` as path sanitization, so no suppression directives (`nolint`/`nosec`) are needed — gosec passes clean.

donpenney:
@fontivan @by2waysprojects any concerns with this update from a konflux point of view? I copied the golang build image ref for the `.konflux/Dockerfile` from the current version of https://github.com/openshift-kni/cluster-group-upgrades-operator/blob/main/.konflux/container_build_args.conf#L7 to be safe. I assume the golang version is now self-contained in the repo, that there's nothing that needs to be updated elsewhere for this

.ci-operator.yaml
```diff
@@ -2,4 +2,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.24-openshift-4.22
+  tag: rhel-9-release-golang-1.25-openshift-4.22
```
fontivan, line 5: Should probably be `...-openshift-5.0` at this point

fontivan: Also important to note that Prow will not actually test changes to `.ci-operator.yaml` within a PR, so once this merges you'll want to check the pipelines to make sure it passes

donpenney: It does use the content of the `.ci-operator.yaml` from the PR. That's actually why I recently migrated oran-o2ims to using this mechanism instead of the previous `Dockerfile.tools` setup, which worked in the way you're thinking. You can see in the logs for the ci-job run, for example, that the image grabbed for the build image is the `rhel-9-release-golang-1.25-openshift-4.22`.

With the previous setup, I would have had to separately update `Dockerfile.tools` with the new build image, merge that, then have a second PR that updates the `go.mod`.

donpenney:
Rebase pushed

.ci-operator.yaml
fontivan: I did not know there was a better way to do this, we should consider this for other operators as well

donpenney: I followed what we have in TALM repo and compared the two setups, because this was something that has been an irritation in this project, and migrated our build config in a few steps in prep for upgrading golang:
https://redhat.atlassian.net/browse/CNF-23325

donpenney: I'd prefer to keep this at `openshift-4.22` for now, since this would currently be deployed on 4.20-4.22 (and 4.22 isn't even GA yet). We can move to `openshift-5.0` at a later point, separate from this upgrade

donpenney:
Rebase pushed

.golangci.yml
```diff
@@ -131,13 +131,26 @@ linters:
         - .WithMessagef(
         - .WithStack(
         - .WrapPrefixf(
+    goconst:
+      min-occurrences: 5
+      ignore-string-values: 'get|list|watch|create|update|patch|delete|app|name|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true'
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — anchored the pattern with `^...$`. This correctly surfaced two previously over-suppressed strings (`namespace` and `agent-install.openshift.io/clusterdeployment-namespace`) which were being matched by the `name` substring. Added both to the ignore list.

.golangci.yml
```diff
@@ -131,13 +131,26 @@ linters:
...
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
+      - linters: [goconst]
+        path: _test\.go$
+      - linters: [goconst]
+        path: test/
+      - linters: [goconst]
+        path: internal/service/.*/api/server\.go$
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — changed to `(^|/)test/` to anchor the pattern as a path component.

Makefile
```diff
@@ -128,8 +128,8 @@ endif
 IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.32.0
```
karampok, line 131: this is for unit test: in other repo we sometime to
```

ENVTEST_K8S_VERSION ?= $(shell go list -m -f "{{ .Version }}" k8s.io/api | awk -F'[v.]' '{printf "1.%d", $$3}')
ENVTEST_VERSION ?= $(shell go list -m -f "{{ .Version }}" sigs.k8s.io/controller-runtime | awk -F'[v.]' '{printf "release-%d.%d", $$2, $$3}')
```
which reads the values from go.mod

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift-kni/oran-o2ims
 
-go 1.24.0
```
karampok, line 3: Adding go toolchain is a good practise
```

go 1.25.0

toolchain go1.25.9
```

hack/update_deps.sh
```diff
@@ -7,12 +7,17 @@
...
+if [ -z "${GO_VERSION}" ]; then
+    echo "Error: could not determine Go version from go.mod"
+    exit 1
+fi
 
 # Use of GOTOOLCHAIN=local will cause failure if the local go version doesn't
-# support the pinned go version, rather than automatically downloading a newer
+# support the go.mod go version, rather than automatically downloading a newer
 # go version
-GOFLAGS='' GOTOOLCHAIN=local go mod tidy -go="${PINNED_GO}"
```
karampok, line 15: This script looks like just being a wrapper when doing
"go mod tidy && go mod vendor", do we really need script for that?

Remote build is using
```
 cat .ci-operator.yaml
---
build_root_image:
  name: release
  namespace: openshift
  tag: rhel-9-release-golang-1.25-openshift-4.22
```
where GOTOOLCHAIN=local

.golangci.yml
```diff
@@ -133,7 +133,7 @@ linters:
         - .WrapPrefixf(
     goconst:
       min-occurrences: 5
-      ignore-string-values: 'get|list|watch|create|update|patch|delete|app|name|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true'
+      ignore-string-values: '^(get|list|watch|create|update|patch|delete|app|name|namespace|master|worker|role|macAddress|nodes|interfaces|label|extraAnnotations|extraLabels|nodeNetwork|apiextensions\.k8s\.io|CustomResourceDefinition|default|true|agent-install\.openshift\.io/clusterdeployment-namespace)$'
```
karampok, line 136: the commit `fix: anchor golangci-lint regex patterns to prevent substring matches`  could be squashed into commit 1 since they work on the file of golang-int

donpenney: The commits get squashed when merged

go.mod
donpenney: Why? What is the benefit?

.golangci.yml
donpenney: Also, the additional commit was handling the response to CodeRabbit comments. I generally try to make such changes in a separate commit for the benefit of reviewers, and also to try to avoid CodeRabbit re-reviewing the entire update.

go.mod
donpenney: The final stage of our `ci-job` is a check for a dirty git tree, intended to catch stale generated files, etc. This `toolchain` directive can be problematic with that check, if it gets updated with whatever tool version is being used. Not just from users being on slightly different versions of the tool, but the potential for the build image to be updated regularly, which in turn could end up causing CI failures.

karampok: the go 1.25 is the minimum specification version (e.g. if code has generics, it must be e.g. 1.21)

toolchain is the actually version to build which should be latest (so all libs have the latest code for security) 

https://docs.google.com/presentation/d/1sPeFM6Qpt8ZeQLVGk2mYz6jvfECnx2uD4LJ0s6fzDmw/edit?slide=id.g2dafff8071a_1_24#slide=id.g2dafff8071a_1_24

.golangci.yml
karampok: so is like fixups? By whom do they get squashed?

donpenney: Our default merge method in the prow config is to squash when merging:
https://github.com/openshift/release/blob/main/core-services/prow/02_config/openshift-kni/oran-o2ims/_prowconfig.yaml#L3

go.mod
donpenney: The build image used by the CI would be the latest tagged image defined in the `.ci-operator.yaml`, and the build image used by the konflux build gets updated on a regular basis as well. So the formal builds should get updated appropriately.

Makefile
donpenney: [Comment authored by Claude]

Good suggestion — deriving these from `go.mod` would keep them in sync automatically. Right now it would give us `1.34` (from `k8s.io/api v0.34.7`) instead of the `1.35.0` we've pinned, because this PR only upgrades Go and tooling (Phase 1) — the k8s dependency upgrade to v0.35 is Phase 2 (CNF-23667). Once that lands, the derivation would produce the correct value.

We'll switch to the `go list` approach when Phase 2 brings the k8s deps in line.

donpenney:
Force-pushed for rebase, and image ref in `.konflux/Dockerfile` updated to latest SHA

PR #2538, opened 2026-05-12
CNF-23691: cache alarm dictionaries on the resource server

donpenney:
## Summary
- Add in-memory cache of pre-built alarm dictionary API models on the ResourceServer
- Cache is lazily populated on first API request, invalidated via the existing `resource_type_changed` PostgreSQL NOTIFY channel and the 15-minute catch-up sync
- Eliminates N+1 query pattern in `GET /alarmDictionaries` (was 1 + N queries)
- Reduces `GET /alarmDictionaries/{id}` and `GET /resourceTypes/{id}/alarmDictionary` from 2 queries each to zero (served from cache)
- Thread-safe via RWMutex for concurrent API requests

## Test plan
- [x] `make test` passes (28 suites, 78 specs in resources API suite)
- [x] `make golangci-lint` passes
- [x] CodeRabbit CLI review: 0 findings
- [x] Coverage thresholds met (listener 39.4%, was 28.7%)
- [x] Unit tests verify cache hit (no repeated DB queries), invalidation + repopulation, and lookup by both dictionary ID and resource type ID
- [x] Deploy and verify alarm dictionary API endpoints return correct data

### DIFF
--- a/internal/service/cluster/api/server_test.go
+++ b/internal/service/cluster/api/server_test.go
@@ -101,6 +101,26 @@ var _ = Describe("Cluster Server", func() {
 		})
 	})
 
+	Describe("Alarm dictionary cache mid-loader failure", func() {
+		It("returns 500 when alarm definitions query fails during cache load", func() {
+			dictID := uuid.New()
+			mockRepo.EXPECT().
+				GetAlarmDictionaries(ctx).
+				Return([]models.AlarmDictionary{{AlarmDictionaryID: dictID}}, nil)
+			mockRepo.EXPECT().
+				GetThanosAlarmDefinitions(ctx).
+				Return([]models.AlarmDefinition{}, nil)
+			mockRepo.EXPECT().
+				GetAlarmDefinitionsByAlarmDictionaryID(ctx, dictID).
+				Return(nil, fmt.Errorf("definitions query failed"))
+
+			resp, err := server.GetAlarmDictionaries(ctx, apigenerated.GetAlarmDictionariesRequestObject{})
+
+			Expect(err).To(BeNil())
+			Expect(resp).To(BeAssignableToTypeOf(apigenerated.GetAlarmDictionaries500ApplicationProblemPlusJSONResponse{}))
+		})
+	})
+
 	Describe("GetAlarmDictionaries", func() {
 		When("cache loading fails", func() {
 			It("returns internal server error", func() {
--- a/internal/service/cluster/api/server_test.go
+++ b/internal/service/cluster/api/server_test.go
@@ -19,7 +19,6 @@ import (
 	apigenerated "github.com/openshift-kni/oran-o2ims/internal/service/cluster/api/generated"
 	"github.com/openshift-kni/oran-o2ims/internal/service/cluster/db/models"
 	"github.com/openshift-kni/oran-o2ims/internal/service/cluster/db/repo/generated"
-	svcutils "github.com/openshift-kni/oran-o2ims/internal/service/common/utils"
 )
 
 var _ = Describe("Cluster Server", func() {
@@ -37,15 +36,16 @@ var _ = Describe("Cluster Server", func() {
 		server = &ClusterServer{
 			Repo: mockRepo,
 		}
+		server.InitAlarmDictCache()
 		ctx = context.Background()
 		testUUID = uuid.New()
 	})
 
 	Describe("GetNodeClusterTypeAlarmDictionary", func() {
-		When("repository returns error", func() {
+		When("cache loading fails", func() {
 			It("returns internal server error", func() {
 				mockRepo.EXPECT().
-					GetNodeClusterTypeAlarmDictionary(ctx, testUUID).
+					GetAlarmDictionaries(ctx).
 					Return(nil, fmt.Errorf("db error"))
 
 				resp, err := server.GetNodeClusterTypeAlarmDictionary(ctx, apigenerated.GetNodeClusterTypeAlarmDictionaryRequestObject{
@@ -57,11 +57,15 @@ var _ = Describe("Cluster Server", func() {
 				Expect(resp.(apigenerated.GetNodeClusterTypeAlarmDictionary500ApplicationProblemPlusJSONResponse).Status).To(Equal(http.StatusInternalServerError))
 			})
 		})
-		When("repository does not have the alarm dictionary", func() {
+
+		When("node cluster type not found in cache", func() {
 			It("returns 404 not found response", func() {
 				mockRepo.EXPECT().
-					GetNodeClusterTypeAlarmDictionary(ctx, testUUID).
+					GetAlarmDictionaries(ctx).
 					Return([]models.AlarmDictionary{}, nil)
+				mockRepo.EXPECT().
+					GetThanosAlarmDefinitions(ctx).
+					Return([]models.AlarmDefinition{}, nil)
 
 				resp, err := server.GetNodeClusterTypeAlarmDictionary(ctx, apigenerated.GetNodeClusterTypeAlarmDictionaryRequestObject{
 					NodeClusterTypeId: testUUID,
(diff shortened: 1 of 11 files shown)


donpenney:
@coderabbitai review

alegacy:
@coderabbitai does the cluster server need this same treatment for its alarm dictionary related api endpoints?

alegacy:
@donpenney & @donpenney's Claude... given that the cluster server also has an alarm dictionary API endpoint then it probably needs the same treatment.  There is probably an opportunity to make some of the code common between the two -- including the cache mechanism, and the pg listener implementation.

alegacy:
[Comment authored by Claude]

Nice work on eliminating the N+1 query pattern and leveraging the existing PG NOTIFY infrastructure for invalidation. A few thoughts on how this could be extended, particularly since the cluster server has the same alarm dictionary endpoints with the same N+1 pattern and would benefit from the same treatment.

## Reusability concern

The cluster server (`internal/service/cluster/api/server.go`) has three equivalent alarm dictionary endpoints — `GetAlarmDictionaries`, `GetAlarmDictionary`, and `GetNodeClusterTypeAlarmDictionary` — all exhibiting the same N+1 query pattern. To apply this PR's approach there, you'd need to:

1. Create a `listener/` package for the cluster server with a new PG NOTIFY trigger on a `node_cluster_type_changed` channel
2. Duplicate the `alarmDictCache` struct, `ensureAlarmDictCache`, and `InvalidateAlarmDictCache` — modified to also fold in Thanos alarm definitions (the cluster server appends `GetThanosAlarmDefinitions()` results to every dictionary)
3. Wire up the new listener goroutine in `cluster/serve.go`

That's ~200+ lines of largely duplicated infrastructure. And if we later want to cache other entity types (ResourcePools, NodeClusters, etc.), each one adds another copy of the mutex+struct+ensure+invalidate pattern.

## Suggestion: extract a generic `cache.Entry[T]` helper

A small generic helper in `internal/service/common/cache/` (~50 lines, stdlib only) could support both TTL-based expiration and explicit invalidation, making the pattern reusable across both servers and any future entity types:

```go
package cache

type Entry[T any] struct {
    mu        sync.RWMutex
    data      T
    valid     bool
    expiresAt time.Time
    maxAge    time.Duration
    loader    func(ctx context.Context) (T, error)
}

func NewEntry[T any](maxAge time.Duration, loader func(ctx context.Context) (T, error)) *Entry[T] {
    return &Entry[T]{maxAge: maxAge, loader: loader}
}

func (c *Entry[T]) Get(ctx context.Context) (T, error) {
    c.mu.RLock()
    if c.valid && time.Now().Before(c.expiresAt) {
        defer c.mu.RUnlock()
        return c.data, nil
    }
    c.mu.RUnlock()

    c.mu.Lock()
    defer c.mu.Unlock()
    if c.valid && time.Now().Before(c.expiresAt) {
        return c.data, nil
    }

    data, err := c.loader(ctx)
    if err != nil {
        var zero T
        return zero, err
    }
    c.data = data
    c.valid = true
    c.expiresAt = time.Now().Add(c.maxAge)
    return c.data, nil
}

func (c *Entry[T]) Invalidate() {
    c.mu.Lock()
    defer c.mu.Unlock()
    c.valid = false
}
```

Both servers would define a small result struct and a loader function:

```go
// Resource server — with PG NOTIFY invalidation + TTL safety net
alarmDicts: cache.NewEntry(15*time.Minute, func(ctx context.Context) (alarmDictData, error) {
    records, err := repo.GetAlarmDictionaries(ctx)
    // ... build indexed result with definitions ...
    return data, nil
})

// Cluster server — TTL-only (no PG listener exists today)
alarmDicts: cache.NewEntry(5*time.Minute, func(ctx context.Context) (alarmDictData, error) {
    records, err := repo.GetAlarmDictionaries(ctx)
    thanosDefinitions, err := repo.GetThanosAlarmDefinitions(ctx)
    // ... build indexed result, appending thanos defs ...
    return data, nil
})
```

The resource server keeps its PG NOTIFY calling `Invalidate()` exactly as this PR does — the TTL just acts as a safety net. The cluster server relies on TTL alone.

## TTL-only on the cluster server: mechanism and tradeoffs

**How TTL works:** Each cache entry has a maximum age (e.g., 5 minutes). When `Get()` is called, it checks both the `valid` flag and `expiresAt`. If the entry is expired, the calling goroutine acquires a write lock, re-queries the DB, and refreshes the cache. All other concurrent readers block briefly during the refresh (typically milliseconds for these queries). Between refreshes, all requests are served from memory with zero DB queries.

**Why TTL is acceptable for the cluster server without a PG listener:**

- The collector already polls on a ~1 minute interval. Data in the DB is already up to ~1 minute stale relative to the real cluster state. Adding a 5-minute TTL on top means worst-case ~6 minutes end-to-end staleness — still well within what's acceptable for alarm dictionary metadata that rarely changes.
- Alarm dictionaries specifically describe the *types* of alarms a cluster *can* produce (severity definitions, clearing types, probable causes). They change only when a new cluster type is deployed or the alarm rule configuration changes — far less frequently than even the collector polling interval.
- The API contract (O-RAN O2 IMS) doesn't mandate real-time consistency for dictionary metadata. Clients that need real-time alarm *events* use the subscription/notification mechanism, not polling the dictionary endpoint.

**What you'd lose compared to implementing a PG listener for the cluster server:**

- **Latency after changes:** With PG NOTIFY, the cache invalidates within seconds of a DB write. With a 5-minute TTL, a client could see stale data for up to 5 minutes after a dictionary change. In practice this matters rarely — dictionary changes are operational events (new cluster type deployed), not user-facing workflows where someone is waiting for the API to reflect their change.
- **Unnecessary refreshes:** TTL causes periodic cache rebuilds even when nothing changed. With PG NOTIFY, the cache only rebuilds when data actually changes. For alarm dictionaries this is negligible cost (a handful of rows), but it's worth noting for larger entity types.
- **No "instant consistency" guarantee:** If you ever need a workflow where a client writes data and immediately reads it back via the API, TTL can't guarantee they'll see their own write. PG NOTIFY can (assuming the notification propagates before the read). This isn't a current requirement for alarm dictionaries, but it's a design difference.

**When it would be worth adding a PG listener to the cluster server:**

- If you start caching high-cardinality entity types (ClusterResources, NodeClusters) where periodic full-table reloads on TTL expiry would be expensive
- If an operational workflow emerges where users expect immediate API consistency after a config change
- If you add caching to endpoints that serve data which changes more frequently than every few minutes

For alarm dictionaries specifically, TTL-only is the pragmatic choice — it delivers the same N+1 elimination and query reduction with zero new infrastructure.

## Minor additional note

The current `ensureAlarmDictCache` holds the write lock during DB queries. If the DB is slow or briefly unreachable, all concurrent API readers block. The generic helper above has the same characteristic. A future improvement could use a build-then-swap pattern (query with no lock, then lock only to swap the pointer), but for alarm dictionaries the query is fast enough that this is unlikely to matter in practice.

donpenney:
[Comment authored by Claude]

@alegacy Good call on both points. The new commit addresses both:

1. **Generic cache extracted** — `cache.CacheEntry[T]` in `internal/service/common/cache/` with lazy loading, TTL expiration, explicit invalidation, and RWMutex for concurrency. 7 unit tests (95.8% coverage).

2. **Cluster server updated** — all three alarm dictionary endpoints (`GetAlarmDictionaries`, `GetAlarmDictionary`, `GetNodeClusterTypeAlarmDictionary`) now use the same cache pattern. The cluster server uses a 5-minute TTL since it has no PG listener, while the resource server keeps PG NOTIFY invalidation with no TTL. Thanos alarm definitions are folded into the cache during loading.

Both servers initialize via `InitAlarmDictCache()` which configures the loader function with the appropriate query logic (resource server: dictionaries + definitions; cluster server: dictionaries + definitions + Thanos definitions).

internal/service/common/cache/cache.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+)
+
+// Entry is a thread-safe, lazily populated cache entry with optional TTL
+// expiration and explicit invalidation support.
+type CacheEntry[T any] struct {
```
alegacy: [Comment authored by Claude]

Two naming suggestions:

1. **Doc comment mismatch**: Line 17 says `// Entry is a thread-safe...` but the type is `CacheEntry`. Line 29 says `// NewEntry creates...` but the function is `NewCacheEntry`. The comments should match the actual identifiers.

2. **Redundant qualifier**: Idiomatic Go would be `cache.Entry` and `cache.NewEntry` rather than `cache.CacheEntry` and `cache.NewCacheEntry` — the package name already provides the `cache` context. Callers would write `cache.Entry[AlarmDictData]` instead of `cache.CacheEntry[AlarmDictData]`. Since this is a new public API, worth getting right before it's used more widely.

internal/service/common/cache/cache.go
```diff
@@ -0,0 +1,78 @@
...
+func NewCacheEntry[T any](name string, maxAge time.Duration, loader func(ctx context.Context) (T, error)) *CacheEntry[T] {
+	return &CacheEntry[T]{
+		name:   name,
+		maxAge: maxAge,
+		loader: loader,
+	}
+}
+
+// Get returns the cached data, loading it if necessary.
+func (c *CacheEntry[T]) Get(ctx context.Context) (T, error) {
```
alegacy: [Comment authored by Claude]

`Get()` returns a reference to the internal `c.data`. When `T` is `AlarmDictData` (containing slices and maps of pointers), all concurrent callers share the same memory. If any caller mutated the returned value, it would corrupt the cache for everyone.

In practice the handlers only serialize to JSON (read-only), so this is safe today. But worth adding a note to the doc comment so future callers know the contract:

```go
// Get returns the cached data, loading it if necessary.
// The returned value is shared across all callers and must not be modified.
```

internal/service/resources/api/server.go
```diff
@@ -40,53 +41,75 @@ type ResourceServerConfig struct {
 	ExternalAddress string
 }
 
+// AlarmDictData holds pre-built alarm dictionary API models with lookup
+// indexes for serving from memory.
+type AlarmDictData struct {
+	All          []generated.AlarmDictionary
+	ByID         map[uuid.UUID]*generated.AlarmDictionary
```
alegacy, line 48: [Comment authored by Claude]

`maxAge: 0` means no TTL — the cache relies entirely on PG NOTIFY + the 15-minute catch-up calling `Invalidate()`. Looking at the catch-up handler in `pg_listener.go`, it calls `onResourceTypeChanged()` unconditionally on every successful sync, so the cache is effectively invalidated every 15 minutes regardless.

This makes TTL=0 acceptable in practice, but the guarantee is implicit — it depends on the catch-up's behavior. If the catch-up ever changes to be conditional (only invalidate when data actually changed), this server would silently lose its staleness bound.

Consider either:
- Using `maxAge: 15*time.Minute` to make the bound explicit in the cache itself, or
- Adding a comment here explaining why TTL=0 is safe (because the catch-up acts as an implicit TTL)

internal/service/cluster/api/server_test.go
```diff
@@ -73,107 +77,49 @@ var _ = Describe("Cluster Server", func() {
 			})
```
alegacy, line 77: [Comment authored by Claude]

Several error-path tests were removed — e.g., "alarm definitions fetch fails for a specific dictionary", "GetThanosAlarmDefinitions fails within GetAlarmDictionary". These failure modes still exist in the loader functions but are now covered only by the generic `propagates loader errors` test in the cache suite.

The generic test proves that *some* loader error propagates, but doesn't exercise the specific failure modes within a multi-step loader (e.g., `GetAlarmDictionaries` succeeds but `GetAlarmDefinitionsByAlarmDictionaryID` fails partway through).

Not a blocker, but consider keeping one test per server that exercises a mid-loader failure to verify those specific error paths remain wired correctly.

alegacy (commented):
[Comment authored by Claude]

Overall this is a solid implementation — the generic `CacheEntry[T]` cleanly addresses the reusability concern and the cluster server integration with Thanos definitions is well done. A few items below, none are blockers.

internal/service/common/cache/cache.go
donpenney: [Comment authored by Claude]

Fixed — renamed to `cache.Entry` and `cache.NewEntry`. Used explicit ginkgo import in the cache tests to avoid the name conflict with Ginkgo's `Entry` type.

internal/service/common/cache/cache.go
donpenney: [Comment authored by Claude]

Added: `// The returned value is shared across all callers and must not be modified.`

internal/service/resources/api/server.go
donpenney: [Comment authored by Claude]

Added a comment explaining why TTL=0 is safe: the PG listener's `resource_type_changed` handler and its 15-minute catch-up sync both call `InvalidateAlarmDictCache`, providing the staleness bound externally.

internal/service/cluster/api/server_test.go
donpenney: [Comment authored by Claude]

Added a mid-loader failure test for each server — `GetAlarmDictionaries` succeeds but `GetAlarmDefinitionsByAlarmDictionaryID` fails partway through. This verifies the specific error path remains wired correctly through the cache loader.

PR #2560, opened 2026-05-18
CNF-23710: Remove payload reflection from error responses

rauhersu:
## Summary

Fix two locations where error responses reflect request data back to the client, preventing potential information disclosure. Removes callback URL and filter values from subscription creation error AdditionalAttributes, and replaces query parse error messages that included the raw query string with a generic message.

**Jira**: [CNF-23710](https://redhat.atlassian.net/browse/CNF-23710) — REQ-034: Remove payload reflection from error responses
**Jira URL**: https://redhat.atlassian.net/browse/CNF-23710

## Changes

**[CNF-23711](https://redhat.atlassian.net/browse/CNF-23711): Remove request data from error AdditionalAttributes and messages**
- Removed `callback` and `filter` keys from `AdditionalAttributes` in subscription creation validation error responses (`server.go`)
- Removed `callback` key from `AdditionalAttributes` in unique callback constraint error responses (`server.go`)
- Replaced query parse error message containing `r.URL.RawQuery` with generic "failed to parse query parameters" (`filtering.go`)
- `consumerSubscriptionId` retained in AdditionalAttributes as it is a server-generated identifier

## Acceptance Criteria

- [x] Subscription creation error responses do not include `request.Body.Callback` or `filter` values in `AdditionalAttributes`
- [x] Query parsing error messages do not include `r.URL.RawQuery` or any raw request data
- [x] Unit tests verify that error responses for these paths do not contain request-submitted data
- [x] All error responses use descriptive messages about the error condition rather than echoing submitted data

## Test Plan

- Added test verifying subscription validation error (http:// callback) does not include callback URL or filter in AdditionalAttributes
- Added test verifying unique_callback error does not include callback URL in AdditionalAttributes (uses httptest.NewTLSServer for callback validation)
- Added test verifying query parse error response does not contain the malicious/raw query string
- All 77 API tests and 27 middleware tests pass
- golangci-lint: 0 issues

### DIFF
--- a/internal/service/common/api/middleware/filtering.go
+++ b/internal/service/common/api/middleware/filtering.go
@@ -28,7 +28,15 @@ const (
 	fields        = "fields"
 	excludeFields = "exclude_fields"
 	filter        = "filter"
+
+	parserErrorPrefix = "invalid filter syntax:"
 )
+
+// isParserError distinguishes selector parser errors (which embed raw user tokens and
+// must not be reflected) from schema validation errors (which contain only safe identifiers).
+func isParserError(err error) bool {
+	return strings.HasPrefix(err.Error(), parserErrorPrefix)
+}
 
 // FilterAdapter is an abstraction that wraps the search projector/selector functionality so that
 // these objects can be created once at server initialization time and re-used in the ResponseFilter
@@ -245,7 +253,11 @@ func ResponseFilter(adapter *FilterAdapter) Middleware {
 					result, err := adapter.ParseFilter(value)
 					if err != nil {
 						slog.Error("failed to parse filter", "value", value, "err", err)
-						_ = adapter.Error(w, err.Error(), http.StatusBadRequest)
+						msg := err.Error()
+						if isParserError(err) {
+							msg = "invalid filter syntax"
+						}
+						_ = adapter.Error(w, msg, http.StatusBadRequest)
 						return
 					}
 					selector.Terms = append(selector.Terms, result.Terms...)
--- a/internal/service/common/api/middleware/filtering.go
+++ b/internal/service/common/api/middleware/filtering.go
@@ -87,7 +87,7 @@ func NewFilterAdapterWithSchemas(logger *slog.Logger, schemas map[string]*openap
 func (a *FilterAdapter) ParseFilter(query string) (*search.Selector, error) {
 	selector, err := a.selectorParser.Parse(query)
 	if err != nil {
-		return nil, fmt.Errorf("invalid filter syntax in '%s': %w", query, err)
+		return nil, fmt.Errorf("invalid filter syntax: %w", err)
 	}
 
 	// Validate field names against schema if validator is available
--- a/internal/service/common/api/middleware/filtering.go
+++ b/internal/service/common/api/middleware/filtering.go
@@ -233,7 +233,7 @@ func ResponseFilter(adapter *FilterAdapter) Middleware {
 				slog.Error("failed to parse query", "RawQuery", r.URL.RawQuery, "err", err)
 				_ = adapter.Error(
 					w,
-					fmt.Sprintf("failed to parse query: %s; error: %s", r.URL.RawQuery, err.Error()),
+					"failed to parse query parameters",
 					http.StatusBadRequest,
 				)
 				return
--- a/internal/service/resources/api/server.go
+++ b/internal/service/resources/api/server.go
@@ -290,7 +290,8 @@ func (r *ResourceServer) GetSubscriptions(ctx context.Context, request api.GetSu
 // validateSubscription validates a subscription before accepting the request
 func (r *ResourceServer) validateSubscription(ctx context.Context, request api.CreateSubscriptionRequestObject) error {
 	if err := commonapi.ValidateCallbackURL(ctx, r.SubscriptionEventHandler.GetClientFactory(), request.Body.Callback); err != nil {
-		return fmt.Errorf("invalid callback url: %w", err)
+		slog.Error("callback URL validation failed", "error", err)
+		return fmt.Errorf("callback URL validation failed")
 	}
 
 	// TODO: add validation of filter and move to common if filter syntax is the same for all servers
--- a/internal/service/resources/api/server.go
+++ b/internal/service/resources/api/server.go
@@ -306,15 +306,9 @@ func (r *ResourceServer) CreateSubscription(ctx context.Context, request api.Cre
 
 	// Validate the subscription
 	if err := r.validateSubscription(ctx, request); err != nil {
-		filter := "<null>"
-		if request.Body.Filter != nil {
-			filter = *request.Body.Filter
-		}
 		return api.CreateSubscription400ApplicationProblemPlusJSONResponse{
 			AdditionalAttributes: &map[string]string{
 				"consumerSubscriptionId": consumerSubscriptionId,
-				"callback":               request.Body.Callback,
-				"filter":                 filter,
 			},
 			Detail: err.Error(),
 			Status: http.StatusBadRequest,
@@ -334,7 +328,6 @@ func (r *ResourceServer) CreateSubscription(ctx context.Context, request api.Cre
 			return api.CreateSubscription400ApplicationProblemPlusJSONResponse{
 				AdditionalAttributes: &map[string]string{
 					"consumerSubscriptionId": consumerSubscriptionId,
-					"callback":               request.Body.Callback,
 				},
 				Detail: "callback value must be unique",
 				Status: http.StatusBadRequest,
--- a/internal/service/resources/api/server_test.go
+++ b/internal/service/resources/api/server_test.go
@@ -10,12 +10,14 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"go.uber.org/mock/gomock"
 
+	commonapi "github.com/openshift-kni/oran-o2ims/api/common"
 	commonmodels "github.com/openshift-kni/oran-o2ims/internal/service/common/db/models"
 	"github.com/openshift-kni/oran-o2ims/internal/service/common/notifier"
 	"github.com/openshift-kni/oran-o2ims/internal/service/resources/api"
@@ -39,6 +41,25 @@ func (m *mockSubscriptionEventHandler) SubscriptionEvent(_ context.Context, _ *n
 
 func (m *mockSubscriptionEventHandler) GetClientFactory() notifier.ClientProvider {
 	return nil
+}
+
+type mockClientProvider struct {
+	client *http.Client
+}
+
+func (m *mockClientProvider) NewClient(_ context.Context, _ commonapi.AuthType) (*http.Client, error) {
+	return m.client, nil
+}
+
+type mockSubscriptionEventHandlerWithProvider struct {
+	provider notifier.ClientProvider
+}
+
+func (m *mockSubscriptionEventHandlerWithProvider) SubscriptionEvent(_ context.Context, _ *notifier.SubscriptionEvent) {
+}
+
+func (m *mockSubscriptionEventHandlerWithProvider) GetClientFactory() notifier.ClientProvider {
+	return m.provider
 }
 
 var _ = Describe("ResourceServer", func() {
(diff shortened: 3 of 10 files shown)


alegacy (commented):
[Comment authored by Claude]

## Review Summary

This PR makes the right fixes for the resources service and filtering middleware, but the scope needs to be expanded to fully satisfy REQ-034. Three areas need attention:

### 1. Incomplete fix — cluster and alarms services have the same vulnerability

The same free-form string reflection pattern fixed in the resources service exists in two other services. Apply the same fix (remove `callback` and `filter` keys from `AdditionalAttributes`, retain validated UUID fields like `consumerSubscriptionId`):

1. `internal/service/cluster/api/server.go`, `CreateSubscription` method:
   - Lines 395-399: remove `"callback"` and `"filter"` keys from the validation error `AdditionalAttributes`
   - Lines 415-419: remove `"callback"` key from the unique_callback error `AdditionalAttributes`

2. `internal/service/alarms/api/server.go`, `CreateSubscription` method:
   - Lines 134-137: remove `"callback"` key from the validation error `AdditionalAttributes`
   - Lines 148-151: remove `"callback"` key from the unique_callback error `AdditionalAttributes`

Add corresponding tests in each service's `server_test.go` matching the pattern used in `internal/service/resources/api/server_test.go`.

### 2. Generated `ErrorHandlerFunc` reflects raw user input in path parameter validation errors

The default `ErrorHandlerFunc` in the oapi-codegen generated code sends `err.Error()` directly to the client via `http.Error()`. When a malformed path parameter is submitted where a UUID is expected, the error includes the raw input verbatim:

```
Invalid format for parameter subscriptionId: error unmarshaling
  '<script>alert(1)</script>' text as *uuid.UUID: invalid UUID length: 25
```

This applies to all services since they all use the generated server wrapper with the default error handler.

Fix: Provide a custom `ErrorHandlerFunc` when calling `HandlerWithOptions()` that returns only the parameter name and expected type without including the inner error detail. For example:

```go
ErrorHandlerFunc: func(w http.ResponseWriter, r *http.Request, err error) {
    var invalidParam *generated.InvalidParamFormatError
    if errors.As(err, &invalidParam) {
        http.Error(w, fmt.Sprintf("Invalid format for parameter %s", invalidParam.ParamName), http.StatusBadRequest)
        return
    }
    http.Error(w, "Bad request", http.StatusBadRequest)
}
```

This should be implemented in the common server setup so it applies to all services consistently. Check `internal/service/common/` for shared server configuration where `HandlerWithOptions` is called.

### 3. Commit message clarification

The commit message states `consumerSubscriptionId` is retained because it's "a server-generated identifier, not user input." It is actually consumer-provided (the API spec says "The value provided by the consumer in the subscription"). The decision to retain it is correct, but the reasoning should be updated to establish the guiding principle. Suggested wording:

*"The consumerSubscriptionId is retained in AdditionalAttributes because it is a validated UUID type that cannot carry arbitrary user data — only syntax-validated constrained types are safe to reflect in error responses; free-form strings (callback URLs, filter expressions, raw query strings) are not."*

internal/service/resources/api/server.go
```diff
@@ -297,15 +297,9 @@ func (r *ResourceServer) CreateSubscription(ctx context.Context, request api.Cre
...
-		if request.Body.Filter != nil {
-			filter = *request.Body.Filter
-		}
 		return api.CreateSubscription400ApplicationProblemPlusJSONResponse{
 			AdditionalAttributes: &map[string]string{
 				"consumerSubscriptionId": consumerSubscriptionId,
-				"callback":               request.Body.Callback,
-				"filter":                 filter,
 			},
 			Detail: err.Error(),
```
alegacy, line 314: [Comment authored by Claude]

The `Detail` field can still reflect the callback URL through Go's HTTP client error wrapping. When `ValidateCallbackURL` → `CheckCallbackReachabilityGET` → `http.Client.Do` fails, Go wraps the full URL in the error, e.g.:
```
invalid callback url: reachability check failed: failed to perform GET request:
  Get "https://attacker.com/callback?secret=token": dial tcp: ...
```

Fix: In `validateSubscription` (this file, ~line 284), replace `return fmt.Errorf("invalid callback url: %w", err)` with a generic message that does not propagate the inner error text, e.g.:
```go
slog.Error("callback URL validation failed", "error", err)
return fmt.Errorf("callback URL validation failed")
```
Log the detailed error for debugging before returning, similar to how `filtering.go` logs the raw query before returning the generic message. Apply the same pattern to `validateSubscription` in `internal/service/cluster/api/server.go`.

internal/service/common/api/middleware/filtering.go
```diff
@@ -233,7 +233,7 @@ func ResponseFilter(adapter *FilterAdapter) Middleware {
 				slog.Error("failed to parse query", "RawQuery", r.URL.RawQuery, "err", err)
 				_ = adapter.Error(
 					w,
-					fmt.Sprintf("failed to parse query: %s; error: %s", r.URL.RawQuery, err.Error()),
+					"failed to parse query parameters",
```
alegacy, line 244: [Comment authored by Claude]

Good fix for the query parse error. However, the filter parse error 12 lines below (line 248) has the same issue. `ParseFilter` at line 90 of this file wraps errors with the raw filter value:
```go
return nil, fmt.Errorf("invalid filter syntax in '%s': %w", query, err)
```
This flows to the client via `adapter.Error(w, err.Error(), ...)` at line 248.

Fix: Change line 90 to remove the `query` interpolation:
```go
return nil, fmt.Errorf("invalid filter syntax: %w", err)
```
The raw filter value is already logged at line 247 via `slog.Error`, so no debugging information is lost.

internal/service/resources/api/server_test.go
```diff
@@ -248,6 +269,66 @@ var _ = Describe("ResourceServer", func() {
 	//
 	// Subscription endpoints
 	//
+	Describe("CreateSubscription", func() {
+		When("validation fails due to invalid callback URL", func() {
+			It("returns 400 without reflecting the callback or filter in AdditionalAttributes", func() {
```
alegacy, line 275: [Comment authored by Claude]

This test uses an `http://` callback which fails at the scheme check — a code path that doesn't reflect the full URL in `Detail`. Add a second test case using an unreachable `https://` URL (e.g., `https://192.0.2.1/callback?secret=token` — RFC 5737 TEST-NET) to verify that `Detail` does not contain the callback URL when the reachability check fails. This will exercise the `http.Client.Do` error wrapping path. Use the same `mockSubscriptionEventHandlerWithProvider` / `mockClientProvider` pattern from the unique_callback test, but with a short-timeout `http.Client` instead of `ts.Client()`.

alegacy:
For the comment related to adding a `ErrorHandlerFunc` we should ensure that the exact error be logged so that we can debug the issue from a server perspective since we won't be sending the error back to the user.   This is a bit unfortunate from a user debuggability point of view, but I see the reasoning from a security perspective.

internal/service/resources/api/server.go
rauhersu: [Comment authored by Claude]

Fixed in 47f580a6. Both `validateSubscription` methods (resources and cluster) now log the detailed error and return a generic message. The alarms service had inline validation — applied the same pattern there: `slog.Error` + hardcoded `"callback URL validation failed"` Detail.

internal/service/common/api/middleware/filtering.go
rauhersu: [Comment authored by Claude]

Fixed in 347572c. `ParseFilter` now uses `fmt.Errorf("invalid filter syntax: %w", err)` without the query interpolation. Added a test verifying `<script>alert(1)</script>` in a filter value doesn't appear in the response.

internal/service/resources/api/server_test.go
rauhersu: [Comment authored by Claude]

Added in 347572c. Each service now has a reachability-failure test using `https://192.0.2.1/callback?secret=token123` with a short-timeout client, verifying `Detail` doesn't contain the URL, IP, or query params.

rauhersu:
[Comment authored by Claude]

All review feedback addressed:

**1. Cluster and alarms services** — Fixed in 47f580a6. Removed `callback` and `filter` from `AdditionalAttributes` in both services. Added corresponding tests (3 per service: scheme validation, reachability failure, unique_callback).

**2. ErrorHandlerFunc / InvalidParamFormatError** — Fixed in 506537e3. `GetOranReqErrFunc` now detects `InvalidParamFormatError` messages and strips the inner error containing raw user input. The client receives only `"Invalid format for parameter <name>"` without the raw input. The full detailed error continues to be logged server-side via `slog.Warn` so it remains available for debugging.

**3. Commit message** — Updated. New commit uses: *"The consumerSubscriptionId is retained in AdditionalAttributes because it is a validated UUID type that cannot carry arbitrary user data — only syntax-validated constrained types are safe to reflect in error responses."*

internal/service/common/api/middleware/filtering.go
```diff
@@ -87,7 +87,7 @@ func NewFilterAdapterWithSchemas(logger *slog.Logger, schemas map[string]*openap
 func (a *FilterAdapter) ParseFilter(query string) (*search.Selector, error) {
 	selector, err := a.selectorParser.Parse(query)
 	if err != nil {
-		return nil, fmt.Errorf("invalid filter syntax in '%s': %w", query, err)
+		return nil, fmt.Errorf("invalid filter syntax: %w", err)
```
alegacy, line 98: [Comment authored by Claude]

This removes the explicit query reflection from the error format string — good. However, the wrapped parser error (`%w`) still contains user input and is reflected to the client at line 248:
```go
_ = adapter.Error(w, err.Error(), http.StatusBadRequest)
```

The selector parser (`internal/search/selector_parser.go`) includes raw token text in its error messages — e.g., `"unknown operator '%s'"` (line 162), `"unexpected token '%s'"` (line 199, 217, 277). After wrapping, the full error sent to the client becomes something like `"invalid filter syntax: unexpected token '<script>alert(1)</script>' while expecting..."`.

The existing test passes by coincidence: the input `(invalid_op,name,<script>alert(1)</script>)` triggers a parser error on `invalid_op` *before* reaching the XSS payload. But a crafted input like `(eq,name,value);<script>alert(1)</script>` would parse the first term successfully, then reflect the payload in the error for the unexpected next token.

Line 248 should use a generic message, consistent with the query parse error fix at line 236:
```go
_ = adapter.Error(w, "invalid filter syntax", http.StatusBadRequest)
```
The detailed error is already logged server-side at line 247.

alegacy (changes requested):
[Comment authored by Claude]

Requesting changes — see inline and general comments.

alegacy:
[Comment authored by Claude]

### Minor: `consumerSubscriptionId` is not server-generated

The PR description says:
> `consumerSubscriptionId` retained in AdditionalAttributes as it is a server-generated identifier

This field comes from `request.Body.ConsumerSubscriptionId` — it is client-supplied, not server-generated. It IS safe to reflect because it's typed as `*openapi_types.UUID` (parsed and validated by the API layer, can only contain hex digits and hyphens). But the characterization in the description should be corrected.

alegacy:
[Comment authored by Claude]

### Test gap: filter error reflection test passes by coincidence

The test `"should not reflect the raw filter value in filter parse error responses"` uses `(invalid_op,name,<script>alert(1)</script>)`. This triggers a parser error on `invalid_op` — the operator position — before the parser ever reaches the XSS payload in the value position. The test passes because the specific error is `"unknown operator 'invalid_op'"`, which doesn't contain `<script>`.

A more thorough test would use an input where the malicious content appears in the token that actually causes the error, e.g. `(eq,name,value);<script>alert(1)</script>`, to confirm the response doesn't reflect it.

rauhersu:
[Comment authored by Claude]

Re: https://github.com/openshift-kni/oran-o2ims/pull/2560#issuecomment-4537190887

Agreed — fixed. The commit message in 347572c uses the corrected reasoning: *"The consumerSubscriptionId is retained in AdditionalAttributes because it is a validated UUID type that cannot carry arbitrary user data."*

rauhersu:
[Comment authored by Claude]

Re: https://github.com/openshift-kni/oran-o2ims/pull/2560#issuecomment-4537191026

Good catch. Fixed in 5a37ad8 — the test now uses `(eq,name,value)<script>alert(1)</script>`, which parses the first term successfully and then fails on the XSS payload as an unexpected token. This exercises the actual reflection path instead of short-circuiting on `invalid_op`.

internal/service/common/api/middleware/filtering.go
rauhersu: [Comment authored by Claude]

Fixed in 5a37ad8. Line 248 now uses a generic "invalid filter syntax" message for parser errors while preserving schema validation messages (which only contain safe field names). An `isParserError` helper distinguishes the two by checking the "invalid filter syntax:" prefix. The test input was also updated to exercise the actual token-reflection path.

alegacy:
I think this looks good now, but I do think this is a step back in term of useability since now a client isn't given as much information about a failure for it to understand what it did wrong.   ...but security trumps useability I guess.

/cc:  @donpenney

donpenney:
> I think this looks good now, but I do think this is a step back in term of useability since now a client isn't given as much information about a failure for it to understand what it did wrong. ...but security trumps useability I guess.
> 
> /cc: @donpenney

@alegacy I'd like to see more info in the logs, but this is consistent with some other error logs in the code currently. We've got a contextual logging system setup that we need to be using everywhere (https://github.com/openshift-kni/oran-o2ims/tree/main/docs/dev/logging), but I'll add a work item / story to review the logging and move the remaining bits to use this system.

PR #2623, opened 2026-05-25
test: cleanup naming and misleading comments

donpenney:
Phase 1 of test code improvements:

- Rename test fixture prefix "cnfdg"/"cnfdf" to "test-worker"/"test-master" in test/utils/vars.go to avoid confusion with internal lab naming conventions (these names caused confusion during OCPBUGS-85549 investigation)
- Update "cnfdg16" in docs/user-guide/cluster-api.md to generic "my-sno-cluster" example name
- Fix stale test name "Should PR reach HardwareConfigured=True after callback" — remove "after callback" which referenced the removed webhook callback architecture
- Fix self-referential comment in testNonCachingListAllocatedNodesForNAR to correctly reference the production function listAllocatedNodesForNAR in internal/controllers/helpers.go
- Remove misleading TODO comment from test/e2e/bmh_test.go that claimed tests "do not test any actual functionality" — the tests do validate bootMACAddress workflow and BMH state for allocation

### DIFF
--- a/test/e2e/bmh_test.go
+++ b/test/e2e/bmh_test.go
@@ -22,7 +22,6 @@ import (
 	testutils "github.com/openshift-kni/oran-o2ims/test/utils"
 )
 
-// TODO: These tests do not test any actual functionality, leave them here for now for revisiting.
 // Test empty bootMACAddress workflow where bootMAC is populated from interface labels
 var _ = Describe("BareMetalHost with empty bootMACAddress", Ordered, func() {
 	const timeout = time.Second * 90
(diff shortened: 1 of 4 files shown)


test/e2e/bmh_test.go
```diff
@@ -22,7 +22,6 @@ import (
 	testutils "github.com/openshift-kni/oran-o2ims/test/utils"
 )
 
-// TODO: These tests do not test any actual functionality, leave them here for now for revisiting.
```
Missxiaoguo, line 25: I think the comment is accurate. The tests don't actually test controller logic. They just create a BMH, update its status to simulate inspection, then read it back and assert the values they just wrote. No ProvisioningRequest or NodeAllocationRequest is created to trigger the controller logic.

donpenney: [Comment authored by Claude]

Agreed — confirmed on closer inspection that the tests just write and read back BMH state without triggering any controller logic. The bootMAC lookup from interface labels is well covered by unit tests in `baremetalhost_manager_test.go` (10 test cases for `setBootMACAddressFromLabel`). Removing bmh_test.go in PR #2630.

PR #2674, opened 2026-05-29
test: consolidate getFakeClientFromObjects into shared test/fakeclient package

donpenney:
Move the duplicated getFakeClientFromObjects helper and SSACompatibleClient wrapper from internal/controllers/suite_test.go into a new test/fakeclient package. This eliminates the duplication between the comprehensive version (suite_test.go, 9 status subresources + field indexer + SSA wrapper) and the minimal version (utils_test.go, 2 status subresources).

A new package is needed because test/utils/ already imports internal/controllers/utils, which would create a circular import. The new test/fakeclient/ package has no internal/controllers dependency. It conditionally registers status subresources and the AllocatedNode field indexer based on which types are present in the caller's scheme.

Fix provision_test.go to use Status().Update() for ManagedCluster status changes — previously relied on the minimal fake client not having ManagedCluster as a status subresource, which allowed c.Update() to write status fields directly.

### DIFF
--- a/internal/controllers/suite_test.go
+++ b/internal/controllers/suite_test.go
@@ -84,8 +84,6 @@ cluster template validation through complete cluster deployment and configuratio
 package controllers
 
 import (
-	"context"
-	"fmt"
 	"log/slog"
 	"os"
 	"testing"
@@ -100,16 +98,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	bmhv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	ibguv1alpha1 "github.com/openshift-kni/cluster-group-upgrades-operator/pkg/api/imagebasedgroupupgrades/v1alpha1"
@@ -117,7 +110,6 @@ import (
 	inventoryv1alpha1 "github.com/openshift-kni/oran-o2ims/api/inventory/v1alpha1"
 	provisioningv1alpha1 "github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1"
 	"github.com/openshift-kni/oran-o2ims/internal/constants"
-	ctlrutils "github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 	assistedservicev1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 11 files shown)


test/fakeclient/fakeclient.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return nil
+}
+
+func (c *SSACompatibleClient) handleServerSideApply(ctx context.Context, obj client.Object, opts ...client.PatchOption) error {
+	isDryRun := false
+	for _, opt := range opts {
+		optStr := fmt.Sprintf("%T", opt)
+		if optStr == "client.dryRunAll" {
+			isDryRun = true
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch — replaced the fragile `fmt.Sprintf("%T")` type-name matching with `PatchOptions.ApplyOptions(opts)` and checking the `DryRun` field. Fixed in 316a8af.

internal/controllers/suite_test.go
Missxiaoguo: A compheresive list of types in the scheme is already registered here, and now that every test uses thie shared fakeclient with this same scheme. I feel it makes sense to consolidate the type registration from individual tests into this single place in the suite setup, and then getFakeClientFromObjects no longer needs to accept a scheme, so any new types should be added here in the further as well. What do you think?
Right now there are not many registrations from the indiviual tests, and the only type is missing from the list is IngressController{} - the rest of types are already registered here 
```
internal/controllers/utils/utils_test.go:       suitescheme.AddKnownTypes(openshiftv1.SchemeGroupVersion, &openshiftv1.IngressController{})
internal/controllers/utils/utils_test.go:       suitescheme.AddKnownTypes(inventoryv1alpha1.GroupVersion, &inventoryv1alpha1.Inventory{})
internal/controllers/utils/utils_test.go:       suitescheme.AddKnownTypes(inventoryv1alpha1.GroupVersion, &inventoryv1alpha1.InventoryList{})
internal/controllers/utils/utils_test.go:       suitescheme.AddKnownTypes(appsv1.SchemeGroupVersion, &appsv1.Deployment{})
internal/controllers/utils/utils_test.go:       suitescheme.AddKnownTypes(appsv1.SchemeGroupVersion, &appsv1.DeploymentList{})
internal/controllers/utils/provision_test.go:   suitescheme.AddKnownTypes(clusterv1.SchemeGroupVersion, &clusterv1.ManagedCluster{})
internal/controllers/utils/provision_test.go:   suitescheme.AddKnownTypes(clusterv1.SchemeGroupVersion, &clusterv1.ManagedClusterList{})
```

internal/controllers/utils/utils_test.go
```diff
@@ -44,23 +45,15 @@ import (
 // Scheme used for the tests:
 var suitescheme = clientgoscheme.Scheme
 
+func init() {
+	suitescheme.AddKnownTypes(openshiftv1.SchemeGroupVersion, &openshiftv1.IngressController{})
```
Missxiaoguo: See comment https://github.com/openshift-kni/oran-o2ims/pull/2674#discussion_r3326958042

internal/controllers/suite_test.go
donpenney: [Comment authored by Claude]

Good suggestion — consolidated all type registrations into `fakeclient.init()` and removed the scheme parameter from `GetFakeClientFromObjects`. The `fakeclient.Scheme` variable is now the single source of truth for test scheme setup. Removed ~35 lines of redundant `AddKnownTypes` calls from suite_test.go, utils_test.go, and provision_test.go. Fixed in 66dc245.

internal/controllers/utils/utils_test.go
donpenney: [Comment authored by Claude]

Addressed — removed `suitescheme` and the inline type registrations. Both files now use `fakeclient.Scheme` and `fakeclient.GetFakeClientFromObjects()` directly.

test/fakeclient/fakeclient.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+const inventoryCRDName = "inventories.ocloud.openshift.io"
+
+// Scheme is the shared scheme for all unit tests that use GetFakeClientFromObjects.
+// All custom resource types are registered here so callers don't need to manage
+// scheme setup individually.
+var Scheme = clientgoscheme.Scheme
+
+func init() {
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch — switched to `runtime.NewScheme()` with `clientgoscheme.AddToScheme()` so we don't mutate the global singleton. Fixed in 863d5dc.

PR #2693, opened 2026-06-01
CNF-21985: [ORAN] Central TLS Profile consistency

rauhersu:
**Why these changes?** 
[CNF-21985](https://redhat.atlassian.net/browse/CNF-21985) implementation . Read `docs/dev/tls-configuration.md`.


rauhersu:
@coderabbit review

internal/cmd/operator/start_controller_manager.go
```diff
@@ -182,6 +185,27 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
 		return err //nolint:wrapcheck
 	}
 
+	// Create a cancellable context; used by the TLS watcher to trigger graceful shutdown.
+	ctx, cancel := context.WithCancel(ctx)
```
alegacy: [Comment authored by Claude]

**HIGH: Signal handling regression**

This replaces `ctrl.SetupSignalHandler()` (which was used in `mgr.Start()` below) with a plain cancellable context derived from `cmd.Context()`. The cobra command context does NOT include OS signal handling (confirmed in `internal/tool.go:176`).

`ctrl.SetupSignalHandler()` registers a handler for SIGTERM/SIGINT with graceful-then-forced shutdown semantics. Without it, when Kubernetes sends SIGTERM during pod termination, the manager will **not** initiate graceful shutdown — the container will be hard-killed after `terminationGracePeriodSeconds`.

This also affects the TLS-watcher restart flow: the process needs to exit cleanly so the Deployment brings up a new pod with the updated profile.

**Fix:** Compose the signal-handling context with the TLS watcher's cancel:
```go
signalCtx := ctrl.SetupSignalHandler()
ctx, cancel := context.WithCancel(signalCtx)
defer cancel()
```

This preserves both: SIGTERM triggers graceful shutdown, and TLS profile changes trigger graceful restart.

internal/cmd/operator/start_controller_manager.go
```diff
@@ -195,6 +219,7 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
 			c.NextProtos = []string{"http/1.1"}
 		})
 	}
+	tlsOpts = append(tlsOpts, ctlrutils.NewTLSConfiguratorFromProfile(tlsProfile))
```
alegacy, line 228: [Comment authored by Claude]

**MEDIUM: `tlsOpts` ordering consideration**

The TLS profile configurator is appended after the http/2 disabling function. The vendored `NewTLSConfigFromProfile` does NOT set `NextProtos` today, so ordering is safe. But if the library ever starts setting `NextProtos`, it would silently override the http/2 disabling above.

Worth either adding a brief comment noting the ordering dependency, or reversing the order (profile first, then http/2 override on top).

internal/controllers/utils/utils.go
```diff
@@ -901,28 +878,11 @@ func loadDefaultCABundles(config *tls.Config) error {
 }
 
 // GetDefaultTLSConfig sets the TLS configuration attributes appropriately to enable communication between internal
-// services and accessing the public facing API endpoints.
+// services and accessing the public facing API endpoints. TLS version and cipher suites are inherited from the
+// cluster TLS security profile (via operator-injected environment variables).
 func GetDefaultTLSConfig(config *tls.Config) (*tls.Config, error) {
```
alegacy: [Comment authored by Claude]

**MEDIUM: Behavioral change for existing callers**

Previously, `GetDefaultTLSConfig` preserved caller-provided `CipherSuites` (`if len(config.CipherSuites) == 0`). The new code always overwrites `MinVersion` and `CipherSuites` via `NewTLSConfiguratorFromProfile`, which unconditionally sets both.

The updated test explicitly validates this new behavior, so it's clearly intentional — but it's a silent semantic change for any existing caller that passes a pre-configured `tls.Config`. Worth verifying no callers rely on the old preserve-if-set behavior.

internal/controllers/utils/tls_profile_test.go
```diff
@@ -0,0 +1,328 @@
...
+	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certDER})
+	keyDER, err := x509.MarshalECPrivateKey(key)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "EC PRIVATE KEY", Bytes: keyDER})
+
+	ExpectWithOffset(1, os.WriteFile(certPath, certPEM, 0o600)).To(Succeed())
+	ExpectWithOffset(1, os.WriteFile(keyPath, keyPEM, 0o600)).To(Succeed())
+}
+
+func bytesEqual(a, b []byte) bool {
```
alegacy: [Comment authored by Claude]

**LOW:** `bytesEqual` duplicates `bytes.Equal` from the standard library.

internal/controllers/utils/tls_profile.go
```diff
@@ -0,0 +1,162 @@
...
+func newTLSProfileFromEnv() configv1.TLSProfileSpec {
+	minVersion := os.Getenv(TLSProfileMinVersionEnvName)
+	ciphersStr := os.Getenv(TLSProfileCiphersEnvName)
+
+	if minVersion == "" {
+		slog.Warn("TLS profile env vars not set, falling back to Intermediate profile")
+		return *configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
+	}
+
+	// Get "," separate ciphers
```
alegacy: [Comment authored by Claude]

**LOW:** Typo — `"separate"` should be `"separated"`.

internal/controllers/envtest/tls_profile_watcher_test.go
```diff
@@ -0,0 +1,165 @@
...
+						TLSProfileSpec: configv1.TLSProfileSpec{
+							MinTLSVersion: configv1.VersionTLS13,
+							Ciphers:       []string{"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384"},
+						},
+					},
+				}
+				Expect(k8sClient.Update(ctx, apiServer)).To(Succeed())
+			}
+			defer func() {
+				_ = k8sClient.Delete(ctx, apiServer)
```
alegacy, line 86: [Comment authored by Claude]

**LOW: Test isolation**

These tests create and delete `APIServer` resources named `"cluster"` but rely on `if err != nil` fallbacks to handle leftovers from previous tests. This works but is fragile — test ordering matters and cleanup isn't guaranteed if a test panics. Consider using `AfterEach` for deterministic cleanup or unique resource names per test.

alegacy (commented):
[Comment authored by Claude]

## Code Review: CNF-21985 — Central TLS Profile Consistency

### Summary

This PR adds support for the OpenShift central TLS security profile by vendoring `controller-runtime-common/pkg/tls`, fetching the cluster TLS profile from `apiservers.config.openshift.io/cluster` at operator startup, applying it to the operator's own metrics/webhook servers, propagating it to server pods via env vars + annotation-driven rolling restarts, and refactoring all existing TLS config functions to delegate to the central profile. RBAC, CSV annotation, and comprehensive tests are included.

All five commit messages accurately describe what the code does. Each claimed function, env var, RBAC change, and test is present and functional.

### Jira Compliance (CNF-21985)

| Requirement | Status | Notes |
|---|---|---|
| Remove hardcoded TLS configs | ✅ Done | `pfsCipherSuites`, `init()`, `PFSCipherSuites()`, hardcoded `MinVersion: tls.VersionTLS12` all removed |
| Fetch TLS policy from APIServer CR | ✅ Done | `FetchAPIServerTLSProfile` reads from `apiservers.config.openshift.io/cluster` |
| Explicitly respect all TLS profile settings | ✅ Done | Both operator and server pods get the profile; no reliance on Go defaults |
| Support Custom TLS profiles | ✅ Done | Handled in `controller-runtime-common` via `GetTLSProfileSpec` |
| CSV `tls-profiles: "true"` | ✅ Done | Updated in both base and bundle manifests |
| PQC readiness | ⚠️ Partial | MinVersion + Ciphers handled; curve preferences have a TODO in vendored lib |
| Functional TLS testing | ⚠️ Partial | Profile fetch/config/hash tests exist; no actual TLS handshake rejection tests |
| ML-KEM testing / tls-scanner re-scan | N/A | Operational validation, likely tracked separately |

### Architecture Assessment

The overall design is sound:
- **Operator reads profile once at startup** using a direct client (before cache is ready) — correct pattern
- **SecurityProfileWatcher triggers restart** on profile changes rather than trying to hot-reload — pragmatic and necessary since controller-runtime doesn't support changing TLSOpts on a running manager
- **Env var propagation** to server pods avoids giving them APIServer read access — good security practice
- **Pod template annotation** for rolling restarts on profile change — standard Kubernetes pattern
- **Existing wrapper functions** (`GetServerTLSConfig`, `GetClientTLSConfig`, `GetDefaultTLSConfig`) preserved as thin wrappers — zero impact on existing call sites

See inline comments for specific issues.

internal/cmd/operator/start_controller_manager.go
```diff
@@ -182,6 +185,27 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
 		return err //nolint:wrapcheck
 	}
 
+	// Create a cancellable context; used by the TLS watcher to trigger graceful shutdown.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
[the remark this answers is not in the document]
rauhersu: Valid finding.

cmd.Context() in this codebase is a plain context.Background() (no signal handling). Fixed by wrapping ctrl.SetupSignalHandler() with context.WithCancel().

This gives us both SIGINT/SIGTERM awareness (from the signal handler) and programmatic cancellation (the TLS profile watcher calls cancel() to trigger graceful restart on profile changes). Addressed in a new commit.

internal/cmd/operator/start_controller_manager.go
rauhersu: Already addressed as a previous crabbit comment: https://github.com/openshift-kni/oran-o2ims/pull/2693#discussion_r3350873570

internal/cmd/operator/start_controller_manager.go
```diff
@@ -182,6 +185,27 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
...
+		logger.ErrorContext(ctx, "Unable to create direct client for TLS profile fetch",
+			slog.String("error", err.Error()))
+		return exit.Error(1)
+	}
+	tlsProfile, err := ctlrutils.FetchAPIServerTLSProfile(ctx, directClient)
+	if err != nil {
+		logger.ErrorContext(ctx, "Unable to fetch cluster TLS profile, using Intermediate default",
+			slog.String("error", err.Error()))
+		tlsProfile = *openshiftv1.TLSProfiles[openshiftv1.TLSProfileIntermediateType]
+	}
```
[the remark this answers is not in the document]
rauhersu: Feedback rejected.

The fallback to Intermediate is a deliberate resilience choice. `FetchAPIServerTLSProfile` already returns the Intermediate profile when no profile is explicitly set, so the only errors are transient API failures or RBAC misconfig. Hard-failing here means operator CrashLoopBackOff on transient issues. The SecurityProfileWatcher will converge once the cache is running.

internal/controllers/utils/tls_profile.go
```diff
@@ -0,0 +1,162 @@
...
+	var ciphers []string
+	if ciphersStr != "" {
+		ciphers = strings.Split(ciphersStr, ",")
+	} else {
+		slog.Warn("TLS profile ciphers env var is empty, TLS 1.3 ciphers will be managed by Go internally",
+			slog.String("minVersion", minVersion))
+	}
```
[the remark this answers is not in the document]
rauhersu: Feedback accepted.

We will fall back to Intermediate profile ciphers when ciphersStr == "" and minVersion != VersionTLS13.

internal/service/alarms/internal/alertmanager/api.go
```diff
@@ -219,14 +218,16 @@ func (c *AMClient) createAlertmanagerClient() (*http.Client, string, error) {
 		return nil, "", fmt.Errorf("failed to append CA certificate")
 	}
 
-	// Create HTTP client with the CA certificate
+	// Create HTTP client with the CA certificate and cluster TLS profile
+	tlsConfig, err := ctlrutils.GetDefaultTLSConfig(nil)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to create TLS config: %w", err)
+	}
+	tlsConfig.RootCAs = rootCAs
```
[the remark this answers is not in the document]
rauhersu: Feedback accepted.

The alertmanager client already has its own explicit trust setup: it loads a specific service CA (ALARMS_SERVER_CA_FILE) and pins it via tlsConfig.RootCAs = rootCAs.  We have changed the GetDefaultTLSConfig signature to not to use the the global skip-verify flag in this caller.

internal/controllers/utils/tls_profile.go
rauhersu: Fixed.

internal/controllers/utils/tls_profile_test.go
rauhersu: Fixed.

internal/controllers/utils/tls_profile.go
```diff
@@ -0,0 +1,162 @@
...
+func NewOutboundTLSConfig(profile configv1.TLSProfileSpec, config *tls.Config) (*tls.Config, error) {
+	if config == nil {
+		config = &tls.Config{} //nolint:gosec
+	}
+
+	config.InsecureSkipVerify = GetTLSSkipVerify()
+	if !config.InsecureSkipVerify {
+		if err := loadDefaultCABundles(config); err != nil {
+			return nil, fmt.Errorf("error loading default CABundles: %w", err)
+		}
```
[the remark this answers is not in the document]
rauhersu: Feedback rejected.

The new loadCAs parameter design already makes it impossible for callers with pre-supplied trust anchors to hit this code path. A defensive nil-check in loadDefaultCABundles would be harmless but adds no value to current or foreseeable callers: all callers with custom RootCAs pass loadCAs=false.

internal/cmd/operator/start_controller_manager.go
rauhersu: Added a comment clarifying the ordering dependency: "profile configurator is appended last so http/2 disabling (NextProtos) is not overridden."

internal/controllers/utils/utils.go
rauhersu: Intentional as commented: the cluster TLS profile must override any caller-provided settings to satisfy CNF-21985's requirement that all TLS configuration inherits centrally. 

The test 'should override caller-provided cipher suites with profile' validates this contract. No existing caller was relying on preserve-if-set; they all passed empty CipherSuites.

internal/controllers/envtest/tls_profile_watcher_test.go
rauhersu: Fixed.

- Added a BeforeEach that deletes any leftover "cluster" APIServer before each test.
- Removed the fragile if err != nil → Get → Update fallback patterns from 3 tests.
- Each test now does a clean Expect(Create).To(Succeed()) since the BeforeEach guarantees no leftovers.
- Tests are now order-independent and robust against Ginkgo randomization.

internal/controllers/envtest/tls_profile_watcher_test.go
```diff
@@ -0,0 +1,145 @@
...
+				Expect(k8sClient.Delete(ctx, existing)).To(Succeed())
+			}
+		})
+
+		It("should fetch the default Intermediate profile when no APIServer exists", func() {
+			profile, err := ctlrutils.FetchAPIServerTLSProfile(ctx, k8sClient)
+			if err != nil {
+				// If APIServer CRD not installed, this is expected in envtest
+				Skip("APIServer CRD not available in this envtest: " + err.Error())
+			}
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/controllers/utils/tls_profile.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Full cipher list is available in the TLS_PROFILE_CIPHERS env var on the pod spec
+	slog.Info("TLS profile loaded from environment",
+		slog.String("minVersion", minVersion),
+		slog.Int("cipherCount", len(ciphers)))
+
+	return configv1.TLSProfileSpec{
+		MinTLSVersion: configv1.TLSProtocolVersion(minVersion),
+		Ciphers:       ciphers,
+	}
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/service/common/notifier/provider.go
```diff
@@ -42,7 +42,7 @@ func NewClientFactory(oauthConfig *ctlrutils.OAuthClientConfig, serviceTokenFile
 }
 
 func (f *ClientFactory) newClusterClient(ctx context.Context) (*http.Client, error) {
-	tlsConfig, _ := ctlrutils.GetDefaultTLSConfig(&tls.Config{MinVersion: tls.VersionTLS12})
+	tlsConfig, _ := ctlrutils.GetDefaultTLSConfig(&tls.Config{MinVersion: tls.VersionTLS12}, true)
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/controllers/utils/oauth.go
```diff
@@ -87,7 +87,7 @@ func SetupOAuthClient(ctx context.Context, logger *slog.Logger, config *OAuthCli
 	if logger == nil {
 		logger = slog.Default()
 	}
-	tlsConfig, _ := GetDefaultTLSConfig(&tls.Config{MinVersion: tls.VersionTLS12})
+	tlsConfig, _ := GetDefaultTLSConfig(&tls.Config{MinVersion: tls.VersionTLS12}, true)
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/controllers/utils/utils.go
```diff
@@ -901,28 +878,27 @@ func loadDefaultCABundles(config *tls.Config) error {
...
-		if err != nil {
-			return nil, fmt.Errorf("error loading default CABundles: %w", err)
+	if loadCAs {
+		tlsConfig.InsecureSkipVerify = GetTLSSkipVerify()
+		if !tlsConfig.InsecureSkipVerify {
+			if err := loadDefaultCABundles(tlsConfig); err != nil {
+				return nil, fmt.Errorf("error loading default CABundles: %w", err)
+			}
 		}
 	}
```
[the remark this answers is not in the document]
rauhersu: Fixed.

alegacy (commented):
[Comment authored by Claude]

Good work on this PR — the architecture is solid, the documentation is excellent, and the implementation aligns closely with the CNF-21985 Jira requirements. I have a few suggestions below.

### 1. Vestigial hardcoded `MinVersion` in callers of `GetDefaultTLSConfig` (Low)

Several callers still pass `&tls.Config{MinVersion: tls.VersionTLS12}` to `GetDefaultTLSConfig`, but `NewOutboundTLSConfig` applies the profile configurator which **overwrites** `MinVersion` unconditionally. These hardcoded values are effectively dead code:

- `internal/service/common/notifier/provider.go:45` — `&tls.Config{MinVersion: tls.VersionTLS12}`
- `internal/controllers/utils/oauth.go:90` — `&tls.Config{MinVersion: tls.VersionTLS12}`
- `internal/controllers/utils/utils.go:936` (`GetDefaultBackendTransport`) — same pattern

The test at `utils_test.go:1707-1716` even explicitly verifies this: *"should override caller-provided cipher suites with profile on GetDefaultTLSConfig"*. Consider passing `nil` instead to avoid misleading future readers into thinking the hardcoded value has any effect.

### 2. Failed profile fetch logged as WARN instead of ERROR (Medium — design tradeoff)

In `start_controller_manager.go:205-207`:
```go
logger.WarnContext(ctx, "Unable to fetch cluster TLS profile, using Intermediate default",
    slog.String("error", err.Error()))
tlsProfile = *openshiftv1.TLSProfiles[openshiftv1.TLSProfileIntermediateType]
```

Falling back to Intermediate on failure is the right call (prevents startup failures in degraded clusters), but logging at WARN could mask persistent RBAC misconfigurations. If the operator never successfully fetches the profile, the only evidence is a WARN log line that may be lost in noise. Consider:
- Logging at **ERROR** level (it *is* an error, just a recoverable one)
- Optionally emitting a status condition or metric so the degraded TLS state is observable via monitoring

### 3. `.NotTo()` vs `.ToNot()` style inconsistency (Low)

Per `AGENTS.md`, the project convention is `.ToNot()` for negated Gomega assertions. The new test files use `.NotTo()` throughout:

- `tls_profile_test.go`: ~30 instances of `.NotTo()` (e.g., `ExpectWithOffset(1, err).NotTo(HaveOccurred())`)
- `tls_profile_watcher_test.go`: ~7 instances of `.NotTo()`

These should be changed to `.ToNot()` to match the codebase standard.

rauhersu:
> ### 1. Vestigial hardcoded `MinVersion` in callers of `GetDefaultTLSConfig` (Low)
> ### 2. Failed profile fetch logged as WARN instead of ERROR (Medium — design tradeoff)
> ### 3. `.NotTo()` vs `.ToNot()` style inconsistency (Low)
All previous comments addressed, @alegacy

rauhersu:
One last thing, I raised [this concern at Ops](https://docs.google.com/document/d/1iJ3iK9vWc8VpPHgAO_UFFRKwoY9bQHjdMxMC2Tn_nlU/edit?disco=AAAB9-e9Q1A) about the 'curve ciphers' we (all operators actually) need on leveraging the controller runtime library for this feature.

I will elaborate on this in the standup next Monday.

internal/cmd/operator/start_controller_manager.go
```diff
@@ -186,6 +188,29 @@ func (c *ControllerManagerCommand) run(cmd *cobra.Command, argv []string) error
...
+	// programmatically trigger graceful shutdown on profile changes.
+	ctx, cancel := context.WithCancel(ctrl.SetupSignalHandler())
+	defer cancel()
+
+	// Fetch the cluster TLS security profile to configure MinVersion/CipherSuites.
+	// This uses a direct client since the manager cache isn't started yet.
+	directClient, err := client.New(ctrl.GetConfigOrDie(), client.Options{Scheme: scheme})
+	if err != nil {
+		logger.ErrorContext(ctx, "Unable to create direct client for TLS profile fetch",
+			slog.String("error", err.Error()))
```
alegacy: [Comment authored by Claude]

Per project logging conventions, errors should be logged with `slog.Any("error", err)` instead of `slog.String("error", err.Error())` to preserve the error type for structured logging handlers. This applies here and at lines 208 and 304 in this file.

```suggestion
			slog.Any("error", err))
```

alegacy (commented):
[Comment authored by Claude]

Logging convention: 3 places use `slog.String("error", err.Error())` instead of `slog.Any("error", err)`. Per project conventions (AGENTS.md), errors should be logged with `slog.Any("error", err)` to preserve the error type for structured logging handlers and avoid eager `.Error()` calls.

internal/cmd/operator/start_controller_manager.go
rauhersu: Fixed in the current run.

alegacy:
@coderabbitai review

internal/controllers/envtest/suite_test.go
```diff
@@ -49,77 +50,82 @@ var (
...
-// validateCRDContent checks that the downloaded content is a valid Kubernetes CRD YAML
-func validateCRDContent(data []byte) error {
-	content := string(data)
+// downloadCRD fetches a CRD from the given URL, validates it, and writes it to testdata/<filename>.
+func downloadCRD(url, filename string) error {
+	resp, err := http.Get(url) //nolint:gosec // URL is constructed from trusted template
+	if err != nil {
+		return fmt.Errorf("failed to download CRD: %w", err)
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/controllers/utils/tls_profile.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if certFile != "" {
+		getCert, err := startCertLoader(ctx, "tls-client", certFile, keyFile)
+		if err != nil {
+			return nil, err
+		}
+		tlsConfig.GetClientCertificate = func(_ *tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return getCert()
+		}
+	}
```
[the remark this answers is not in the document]
rauhersu: Fixed.

internal/service/common/notifier/notifier_test.go
```diff
@@ -337,10 +336,84 @@ var _ = Describe("Notifier integration", func() {
...
+		cancel()
+
+		notif := &Notification{NotificationID: uuid.New(), SequenceID: 1, Payload: "test"}
+		done := make(chan struct{})
+		go func() {
+			n.Notify(ctx, notif)
+			close(done)
+		}()
+		Eventually(done, 2*time.Second).Should(BeClosed())
+	})
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

The channel is buffered (DefaultBufferedChannelSize). In the test, the context is cancelled before Notify is called (cancel() is called before the goroutine starts). When a select has multiple ready cases, Go picks one non-deterministically — but in this test the channel buffer has capacity, so both cases are ready. However, the test is only verifying that the method returns (doesn't hang), which is the actual contract being tested. Adding assertions about whether the channel received the event would make the test flaky because of Go's non-deterministic select behavior. The test correctly validates the stated behavior: "aborts when context is canceled" means the method doesn't block indefinitely.

rauhersu:
@coderabbit review

rauhersu:
@coderabbitai why code reviews are not automatically run on a push basis for this PR ? I have had to explicitly ask for a review with a command here.

donpenney:
@rauhersu The tip about `auto_pause_after_reviewed_commits: 0` looks useful. We should do that, in a separate PR. cc @alegacy

donpenney:
lgtm... I'll let @alegacy have the final word

donpenney:
/override ci/prow/scorecard

Tested manually

PR #2761, opened 2026-06-05
CNF-24578: add BMH/node context enrichment and standardize field naming

donpenney:
Add logging.AppendCtx() for bmh and node names at function entry points throughout the hardware manager controller. This eliminates ~100 redundant explicit bmh/node attributes from individual log calls — the data is now automatically included via context in all downstream logs.

Context enrichment added to 30+ functions in:
- baremetalhost_manager.go: all functions receiving bmh parameter
- helpers.go: node processing functions (handleNodeInProgressUpdate, initiateNodeUpdate, executeNodeUpdates, etc.)
- allocatednode_controller.go: handleAllocatedNodeDeletion
- hostfirmwarecomponents_manager.go, hostfirmwaresettings_manager.go, hostupdatepolicy_manager.go: BMH processing functions

Field naming standardized to camelCase:
- "BMH" → "bmh" (baremetalhost_manager.go, helpers.go)
- "ClusterID" → "clusterId" (nodeallocationrequest_controller.go)
- "NodeClusterType ID" → "nodeClusterTypeId" (collector_alarms.go)
- "State" → "state", "Annotation" → "annotation"

### DIFF
--- a/internal/hardwaremanager/controller/baremetalhost_manager.go
+++ b/internal/hardwaremanager/controller/baremetalhost_manager.go
@@ -84,6 +84,7 @@ func updateBMHMetaWithRetry(
 	metaType string, // "label" or "annotation"
 	key, value, operation string,
 ) error {
+	ctx = logging.AppendCtx(ctx, slog.String("bmh", name.Name))
 	// nolint: wrapcheck
 	return retry.OnError(retry.DefaultRetry, k8serrors.IsConflict, func() error {
 		// Fetch the latest version of the BMH
--- a/internal/hardwaremanager/controller/baremetalhost_manager.go
+++ b/internal/hardwaremanager/controller/baremetalhost_manager.go
@@ -84,7 +84,6 @@ func updateBMHMetaWithRetry(
 	metaType string, // "label" or "annotation"
 	key, value, operation string,
 ) error {
-	ctx = logging.AppendCtx(ctx, slog.String("bmh", name.Name))
 	// nolint: wrapcheck
 	return retry.OnError(retry.DefaultRetry, k8serrors.IsConflict, func() error {
 		// Fetch the latest version of the BMH
@@ -413,8 +412,6 @@ func processHwProfile(ctx context.Context,
 	namespace string,
 	bmh *metal3v1alpha1.BareMetalHost, profileName string,
 	postInstall, validateOnly bool) (bool, error) {
-	ctx = logging.AppendCtx(ctx, slog.String("bmh", bmh.Name))
-
 	logger.DebugContext(ctx, "Processing hardware profile for BMH",
 		slog.String("hwProfile", profileName),
 		slog.Bool("postInstall", postInstall),
@@ -507,7 +504,6 @@ func processHwProfile(ctx context.Context,
 }
 
 func checkBMHStatus(ctx context.Context, logger *slog.Logger, bmh *metal3v1alpha1.BareMetalHost, state metal3v1alpha1.ProvisioningState) bool {
-	ctx = logging.AppendCtx(ctx, slog.String("bmh", bmh.Name))
 	// Check if the BMH is in  desired state
 	if bmh.Status.Provisioning.State == state {
 		logger.InfoContext(ctx, "BMH is now in desired state", slog.String("state", string(state)))
@@ -620,7 +616,6 @@ func handleTransitionNode(ctx context.Context,
 }
 
 func addRebootAnnotation(ctx context.Context, c client.Client, logger *slog.Logger, bmh *metal3v1alpha1.BareMetalHost) error {
-	ctx = logging.AppendCtx(ctx, slog.String("bmh", bmh.Name))
 	bmhName := types.NamespacedName{Name: bmh.Name, Namespace: bmh.Namespace}
 	logger.DebugContext(ctx, "Adding reboot annotation to BMH",
 		slog.String("annotation", BmhRebootAnnotation))
@@ -635,7 +630,6 @@ func addRebootAnnotation(ctx context.Context, c client.Client, logger *slog.Logg
 // firmware updates and adds the reboot annotation to the BMH to trigger node
 // reboot when required.
 func evaluateCRForReboot(ctx context.Context, c client.Client, logger *slog.Logger, bmh *metal3v1alpha1.BareMetalHost) error {
-	ctx = logging.AppendCtx(ctx, slog.String("bmh", bmh.Name))
 	// Check if both annotations are present
 	hasBiosAnnotation := bmh.Annotations[BiosUpdateNeededAnnotation] != ""
 	hasFirmwareAnnotation := bmh.Annotations[FirmwareUpdateNeededAnnotation] != ""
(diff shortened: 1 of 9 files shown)


internal/hardwaremanager/controller/baremetalhost_manager.go
```diff
@@ -636,34 +620,32 @@ func handleTransitionNode(ctx context.Context,
 }
 
 func addRebootAnnotation(ctx context.Context, c client.Client, logger *slog.Logger, bmh *metal3v1alpha1.BareMetalHost) error {
+	ctx = logging.AppendCtx(ctx, slog.String("bmh", bmh.Name))
```
Missxiaoguo: I noticed that the caller `evaluateCRForReboot` already appends bmh attribute to the ctx, and the caller of `evaluateCRForReboot` appends it again. Would this result in duplicate bmh fields in the log? It looks like the same pattern occurs in many other call chains as well.

internal/hardwaremanager/controller/hostfirmwaresettings_manager.go
Missxiaoguo: can be removed

internal/hardwaremanager/controller/baremetalhost_manager.go
donpenney: [Comment authored by Claude]

Good catch. Removed `AppendCtx` for bmh from leaf functions that are always called from a parent that already sets it in context. Only top-level entry points where the BMH is first identified now enrich the context. Fixed in e12bf38.

The remaining case of both caller and callee adding bmh (e.g., `handleAllocatedNodeDeletion` → `clearBMHAnnotation`) will be resolved by PR #2759 which fixes `AppendCtx` to replace existing keys instead of appending duplicates.

internal/hardwaremanager/controller/hostfirmwaresettings_manager.go
donpenney: [Comment authored by Claude]

Removed in e12bf38.

internal/hardwaremanager/controller/baremetalhost_manager.go
```diff
@@ -935,7 +897,8 @@ func handleBMHCompletion(ctx context.Context,
 		go func(node *hwmgmtv1alpha1.AllocatedNode) {
 			defer wg.Done()
 
-			updating, err := handleSingleNodeCompletion(ctx, c, noncachedClient, logger, namespace, node)
+			nodeCtx := logging.AppendCtx(ctx, slog.String("node", node.Name))
+			updating, err := handleSingleNodeCompletion(nodeCtx, c, noncachedClient, logger, namespace, node)
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Valid concern. The concurrent slice mutation is fixed by PR #2759 which updates `AppendCtx` to copy the slice before modification. The goroutine code already uses a local `nodeCtx` per goroutine, so once the copy fix lands, each goroutine operates on its own slice copy with no data race.

internal/service/cluster/collector/collector_alarms.go
```diff
@@ -106,15 +106,15 @@ func nodeClusterTypesWithAlarmDictionaryID(nodeClusterTypes []models.NodeCluster
 	var filteredNodeClusterTypes []models.NodeClusterType
 	for _, nodeClusterType := range nodeClusterTypes {
 		if nodeClusterType.Extensions == nil {
-			slog.Error("no extensions found for node cluster type", "NodeClusterType ID", nodeClusterType.NodeClusterTypeID)
+			slog.Error("no extensions found for node cluster type", "nodeClusterTypeId", nodeClusterType.NodeClusterTypeID)
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Aligned to `nodeClusterTypeID` (uppercase ID suffix) to match the convention used in context enrichment and the API model fields. Fixed in e12bf38.

alegacy:
/lgtm

@Missxiaoguo you can approve if you are satisfied with the fixes to your comments.

donpenney:
[Comment authored by Claude]

Rebased on main (which now includes the `AppendCtx` replace-instead-of-append fix from PR #2759) and force-pushed with an additional commit addressing review feedback and extending context enrichment to the resource collector:

**Review feedback fixes:**
- Removed `AppendCtx` from leaf functions that inherit bmh from caller (per Angie's feedback)
- Aligned `nodeClusterTypeId` → `nodeClusterTypeID` for consistency (per CodeRabbit)
- Removed leftover explicit `slog.String("bmh", ...)` in hostfirmwaresettings_manager.go (per Angie)

**Additional entry points enriched (same pattern, previously missed):**
- `updateBMHMetaWithRetry`: receives a NamespacedName rather than being called from a parent that already set bmh — adds bmh to context so logs inside the retry loop always identify the BMH
- `handleBMHEvent`, `handleHardwareDataEvent`, `handleAllocatedNodeEvent`: entry points for watch events in the resource collector — adds bmh/node to context at event entry, removes redundant explicit attributes from downstream log calls
- Removed redundant `HFC` attribute from HostFirmwareComponents creation log (bmh name already in context)

BMH context format is now consistently `bmh.Name` everywhere, with `bmhNamespace` as a separate field in the resource collector where BMHs come from different namespaces.

donpenney:
[Comment authored by Claude]

@coderabbitai Re: [CodeRabbit nitpick](https://github.com/openshift-kni/oran-o2ims/pull/2761#pullrequestreview-4449503130) on `IsFirmwareUpdateRequired` — this is a leaf function called from `processHwProfile`, which is called from `initiateNodeUpdate`/`handleNodeInProgressUpdate`. Both callers already set bmh in context before calling down, so `IsFirmwareUpdateRequired` inherits bmh via the passed ctx. Per the approach in this PR, we only add bmh to context at entry points where the BMH is first identified, not in every function that receives a bmh parameter.

internal/hardwaremanager/controller/helpers.go
```diff
@@ -1162,22 +1162,22 @@ func executeNodeUpdates(
...
 						namespace, nodeToProcess.node, true, nodeOps)
 				}
 			case actionInProgressUpdate:
 				if nodeToProcess.node.Spec.HwProfile != newHwProfile {
-					res, err = abandonNodeUpdate(ctx, c, noncachedClient, logger,
+					res, err = abandonNodeUpdate(nodeCtx, c, noncachedClient, logger,
 						newHwProfile, nodeToProcess.node, nodeOps)
 				} else {
-					res, err = handleNodeInProgressUpdate(ctx, c, noncachedClient, logger,
+					res, err = handleNodeInProgressUpdate(nodeCtx, c, noncachedClient, logger,
```
Missxiaoguo, line 1177: Since the `node` has already been appended to the ctx in executeNodeUpdates, the additional appending in each of these child functions can be removed.

donpenney: [Comment authored by Claude]

Good catch — removed the redundant `AppendCtx` for node from `handleNodeInProgressUpdate`, `abandonNodeUpdate`, and `initiateNodeUpdate`. The caller (`executeNodeUpdates`) already sets node in context via `nodeCtx` before calling these. Fixed in 30d7468.

PR #2770, opened 2026-06-08
CNF-24580: suppress duplicate alarm rule WARN logs

donpenney:
Replace per-rule WARN logging of duplicate Prometheus rules with a single DEBUG-level summary after filtering. Previously, each duplicate rule was logged individually at WARN level with the full rule body (PromQL expression, labels, annotations), producing ~48 WARN entries per 60-second sync cycle with identical duplicates every time.

Now logs a single DEBUG entry with the count and alert names of duplicates found. Duplicates are expected (same rules appear on multiple managed clusters) and not actionable, so WARN was the wrong level.


donpenney:
Rebase pushed

PR #2775, opened 2026-06-08
CNF-23716: Add scope-to-groups-to-RBAC observability and documentation

rauhersu:
## Summary
- Log user groups at DEBUG level in the Authorizer on successful authorization, making the OAuth scope → groups → RBAC chain observable for compliance auditing ([CNF-23717](https://redhat.atlassian.net/browse/CNF-23717))
- Document DD-003 in AGENTS.md explaining the indirect scope verification pattern, and add clarifying comments above the `security:` section in all five service OpenAPI specs ([CNF-23718](https://redhat.atlassian.net/browse/CNF-23718))

## Test plan
- [x] New test verifies groups appear in log output on successful authorization
- [x] Auth package tests pass (`go test ./internal/service/common/auth/`)

### DIFF
--- a/internal/service/common/auth/auth.go
+++ b/internal/service/common/auth/auth.go
@@ -98,14 +98,20 @@ func Authorizer(kubernetesAuthorizer authorizer.Authorizer) middleware.Middlewar
 			decision, reason, err := kubernetesAuthorizer.Authorize(req.Context(), attributes)
 			if err != nil {
 				msg := fmt.Sprintf("Authorization for user '%s' failed", attributes.User.GetName())
-				slog.ErrorContext(req.Context(), msg, slog.Any("user", user), slog.String("verb", attributes.Verb), slog.String("path", attributes.Path), slog.Any("error", err))
+				slog.ErrorContext(req.Context(), msg,
+					slog.String("user", user.GetName()), slog.Any("groups", user.GetGroups()),
+					slog.String("verb", attributes.Verb), slog.String("path", attributes.Path),
+					slog.Any("error", err))
 				middleware.ProblemDetails(w, msg, http.StatusInternalServerError)
 				return
 			}
 
 			if decision != authorizer.DecisionAllow {
 				msg := fmt.Sprintf("Authorization not allowed for user '%s'", attributes.User.GetName())
-				slog.DebugContext(req.Context(), msg, slog.Any("user", user), slog.String("verb", attributes.Verb), slog.String("path", attributes.Path), slog.Any("decision", decision), slog.String("reason", reason))
+				slog.DebugContext(req.Context(), msg,
+					slog.String("user", user.GetName()), slog.Any("groups", user.GetGroups()),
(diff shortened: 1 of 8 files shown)


internal/service/common/auth/auth.go
```diff
@@ -110,6 +110,10 @@ func Authorizer(kubernetesAuthorizer authorizer.Authorizer) middleware.Middlewar
 				return
 			}
 
+			slog.DebugContext(req.Context(), "authorization allowed",
+				"user", user.GetName(), "groups", user.GetGroups(),
```
donpenney: nit:

I thought the use of typed attrs (ie. `slog.String("user", user.GetName())`) was being enforced in golangci-lint, but it looks like I only had that in oran-hwmgr-plugin. I'll add another logging work item to deal with that (created CNF-24707), since there seems to be quite a few cases using untyped key-value pairs.

Should also note that the `slog.ErrorContext` calls above this are using `user`, whereas you're using `user.GetName()` and `user.GetGroups()` here. I guess the error log is just spitting out the entire user data structure. I wonder if that's too much data being logged?

donpenney:
We've enabled the check requiring all commits to be signed before merging. Please setup your git commit signing and update the commits

donpenney:
Commits must be signed and verified before merging:
https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

internal/service/common/auth/auth.go
donpenney: The pre-existing `slog.ErrorContext` calls have been updated in https://github.com/openshift-kni/oran-o2ims/pull/2788, which has now merged.  The golangci-lint check is enabled in https://github.com/openshift-kni/oran-o2ims/pull/2789

rauhersu:
@alegacy @donpenney , re-pushed and verified badge present now.

internal/service/common/auth/auth.go
```diff
@@ -110,6 +110,10 @@ func Authorizer(kubernetesAuthorizer authorizer.Authorizer) middleware.Middlewar
...
+				"verb", attributes.Verb, "path", attributes.Path)
```
donpenney:
```suggestion
			slog.DebugContext(req.Context(), "authorization allowed",
				slog.String("user", user.GetName()), slog.String("groups", user.GetGroups()),
				slog.String("verb", attributes.Verb), slog.String("path", attributes.Path))
```

rauhersu: @coderabbitai: can you check this has been addressed and close the conversation if so ?

internal/service/common/auth/auth.go
rauhersu: Replaced slog.Any("user", user) with targeted slog.String("user",                                                                                                                                                                                              
user.GetName()) and slog.Any("groups", user.GetGroups()) on the error                                                                                                                                                                                         
and denied authorization paths.

PR #2787, opened 2026-06-09
CNF-24707: use typed slog attributes in alarms internal packages

donpenney:
Replace untyped alternating key-value pairs with typed slog attribute constructors in the alarms infrastructure, alertmanager, listener, db/repo, serviceconfig, and notifier_provider packages.

### DIFF
--- a/internal/service/alarms/internal/alertmanager/api.go
+++ b/internal/service/alarms/internal/alertmanager/api.go
@@ -71,7 +71,7 @@ func (c *AMClient) RunAlertSyncScheduler(ctx context.Context, interval time.Dura
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
 
-	slog.InfoContext(ctx, "Alert sync scheduler started", "interval", interval.String())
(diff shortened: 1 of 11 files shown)


internal/service/alarms/internal/alertmanager/api.go
```diff
@@ -160,7 +160,7 @@ func (c *AMClient) getAlerts(ctx context.Context) ([]APIAlert, error) {
 	}
 	defer func(Body io.ReadCloser) {
 		if err := Body.Close(); err != nil {
-			slog.ErrorContext(ctx, "failed to close response body during AM get call", "error", err.Error())
+			slog.ErrorContext(ctx, "failed to close response body during AM get call", slog.String("error", err.Error()))
```
donpenney, line 163: We will need to standardize `slog.String("error", err.Error())` vs `slog.Any("error", err)`, but that can be a followup update. This PR (and #2786 and #2788) are about typing the existing untyped log fields, and not revising what's logged.

donpenney:
@coderabbitai There are a large number of calls throughout the code using `slog.String("error", err.Error())`, This will be done as a separate work item. I've created a Jira work item, CNF-24759, for this change

PR #2790, opened 2026-06-10
OCPBUGS-87974: fix non-hardware PR spec changes causing NAR configTransactionId mismatch

Missxiaoguo:
After a day-2 hardware firmware update completes (NAR Configured condition is present), subsequent non-hardware ProvisioningRequest spec changes (e.g., clusterInstance parameters) caused the PR to get stuck in "Hardware provisioning completed: Created". ConfigTransactionId, derived from PR.generation, was bumped on every spec change and included in the NAR spec comparison, triggering a NAR update even with no hardware-relevant changes. The NAR controller acknowledged ObservedGeneration but not ObservedConfigTransactionId when no HW profile changes were detected, causing the PR controller's stale guard to block indefinitely.

Exclude ConfigTransactionId from the NAR spec DeepEqual comparison so only hardware-relevant field changes trigger a NAR update. Add UpdateNodeAllocationRequestObservedStatus helper that acknowledges both ObservedGeneration and ObservedConfigTransactionId, and use it in the no-HW-profile-changes path.

### DIFF
--- a/internal/hardwaremanager/utils/node_allocation_request.go
+++ b/internal/hardwaremanager/utils/node_allocation_request.go
@@ -83,6 +83,34 @@ func UpdateNodeAllocationRequestObservedGeneration(
 	return nil
 }
 
+// UpdateNodeAllocationRequestObservedStatus acknowledges all spec changes by updating
+// both ObservedGeneration and ObservedConfigTransactionId in a single status update.
+func UpdateNodeAllocationRequestObservedStatus(
+	ctx context.Context,
+	c client.Client,
+	nodeAllocationRequest *hwmgmtv1alpha1.NodeAllocationRequest) error {
+
+	// nolint: wrapcheck
+	err := ctlrutils.RetryOnConflictOrRetriable(retry.DefaultRetry, func() error {
+		newNodeAllocationRequest := &hwmgmtv1alpha1.NodeAllocationRequest{}
+		if err := c.Get(ctx, client.ObjectKeyFromObject(nodeAllocationRequest), newNodeAllocationRequest); err != nil {
+			return err
+		}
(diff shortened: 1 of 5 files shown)


internal/hardwaremanager/utils/node_allocation_request.go
```diff
@@ -83,6 +83,34 @@ func UpdateNodeAllocationRequestObservedGeneration(
 	return nil
 }
 
+// UpdateNodeAllocationRequestObservedStatus acknowledges all spec changes by updating
+// both ObservedGeneration and ObservedConfigTransactionId in a single status update.
+func UpdateNodeAllocationRequestObservedStatus(
```
donpenney, line 88: nit: Claude noticed this function is very similar to `UpdateNodeAllocationRequestObservedGeneration`, and maybe these could be consolidated

donpenney:
Commits must be signed and verified before merging:
https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Otherwise, lgtm

internal/hardwaremanager/utils/node_allocation_request.go
Missxiaoguo: Right, the `UpdateNodeAllocationRequestObservedGeneration` is used by day0 provisioning and other paths where `ObservedConfigTransactionId` should not be touched. `ObservedConfigTransactionId` is only relevant to day2 hardware updates. I thought about adding a parameter to control the behaviour but I ended up creating a separate function, which I think is cleaner.

Missxiaoguo:
> Commits must be signed and verified before merging: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification
> 
> Otherwise, lgtm

Fixed now!

PR #2800, opened 2026-06-10
CNF-24428: Update SNO upgrade examples to inline upgradeDefaults format

Missxiaoguo:
## Summary

Update SNO upgrade sample YAMLs and documentation to use the new inline `upgradeDefaults` format introduced in #2799. Add documentation for overriding upgrade parameters per ProvisioningRequest.

**Jira**: [CNF-24426](https://redhat.atlassian.net/browse/CNF-24426) — Simplify the upgrade template for SNO upgrade
**Depends on**: #2799

## Changes
**[CNF-24428](https://redhat.atlassian.net/browse/CNF-24428): Upgrade doc updates**
- `config/samples/v1alpha1_clustertemplate.yaml`: Replace `upgradeDefaults: upgrade-defaults-v1` with inline IBGU spec/plan, add `upgradeParameters` to `templateParameterSchema` with full imageBasedGroupUpgrade schema
- `docs/samples/.../sno-ran-du-v4-Y-Z+1-2.yaml`: Same inline conversion and schema addition
- Delete `docs/samples/.../upgrade-defaults-v1.yaml` (old ConfigMap no longer needed)
- Remove ConfigMap reference from `kustomization.yaml`
- `docs/user-guide/upgrade.md`: Rewrite to show inline format, add "Overriding Upgrade Parameters per ProvisioningRequest" section with deep-merge explanation and example, add `PreconditionChecksFailed` to monitoring table


donpenney:
[Comment authored by Claude]

Clean PR — inline format, schema definitions, and ConfigMap cleanup all look correct. Two minor observations:

**Override example: redundant Prep timeout**
The override example sets `timeout: 15` for the Prep step, which matches the default — making it a no-op. Changing it to a different value (e.g., 20) would better demonstrate that overrides actually change behavior.

**Suggestion: explicit statement about shorter arrays**
The deep-merge preserves unmatched default elements when the override array is shorter (nicely demonstrated in the example where 3 override items merge with 5 defaults, preserving items 3 and 4). A brief note like "If the override array is shorter than the defaults, unmatched default elements are preserved" would help users who might expect full array replacement — this is a subtle behavior that could catch people off guard.

Neither is blocking. The doc is well-structured and the samples are consistent across all three locations.

PR #2824, opened 2026-06-13
Fix panic in inventory controller when LoggerFromContext finds no logger

alegacy:
LoggerFromContext performs an unsafe type assertion on the context value. When no logger has been injected via LoggerIntoContext, ctx.Value() returns a nil interface{}, and the assertion panics with "interface conversion: interface {} is nil, not *slog.Logger" before the caller's nil-guard can execute.

This only occurs when the register on restart functionality is enabled by annotation.

### DIFF
--- a/internal/controllers/inventory_controller.go
+++ b/internal/controllers/inventory_controller.go
@@ -37,7 +37,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	inventoryv1alpha1 "github.com/openshift-kni/oran-o2ims/api/inventory/v1alpha1"
-	"github.com/openshift-kni/oran-o2ims/internal"
 	"github.com/openshift-kni/oran-o2ims/internal/constants"
 	ctlrutils "github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 )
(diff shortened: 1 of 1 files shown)


internal/controllers/inventory_controller.go
```diff
@@ -111,14 +110,10 @@ const registerOnRestartAnnotation = "ocloud.openshift.io/register-on-restart"
 var registerOnRestart = false
 
 // setRegisterOnRestart initializes the `registerOnRestart` value from an annotation.
-func setRegisterOnRestart(ctx context.Context, object *inventoryv1alpha1.Inventory) {
+func setRegisterOnRestart(ctx context.Context, logger *slog.Logger, object *inventoryv1alpha1.Inventory) {
 	if annotation, ok := object.Annotations[registerOnRestartAnnotation]; ok {
 		if value, err := strconv.ParseBool(annotation); err == nil {
 			registerOnRestart = value
-			logger := slog.Default()
```
alegacy, line 118: @donpenney  Thought about adding the logger to the context at the beginning of the reconciler, but given that `LoggerFromContext` is barely used anywhere else I figured might as well just use the logger directly.

donpenney (approved):

PR #2829, opened 2026-06-14
CNF-24707: standardize error log key to "error"

donpenney:
Rename inconsistent error attribute keys:
- slog.Any("err", err) → slog.Any("error", err) (21 calls)
- slog.Any("Error", err) → slog.Any("error", err) (3 calls)
- slog.String("err", err.Error()) → slog.Any("error", err) (4 calls)


donpenney:
Updated for rebase

PR #2832, opened 2026-06-14
CNF-24707: add String() method to AsyncEventType

donpenney:
- Add fmt.Stringer implementation so event types render as "Updated", "Deleted", or "SyncComplete" instead of raw integers in log output
- Update call sites from slog.Any("type", eventType) to slog.String("type", eventType.String()) since slog.Any on int-based types uses the numeric value, bypassing Stringer


donpenney:
[Comment authored by Claude]

@coderabbitai Good catch — fixed in the latest push. The `slog.Any("event", o.eventType)` at line 146 of `store.go` has been converted to `slog.String("event", o.eventType.String())` for consistency with the rest of the PR.

donpenney:
Updated for rebase

PR #2867, opened 2026-06-16
CNF-24707: fix duplicate bmh log key in resource collector

donpenney:
The buildAndSendResource and findAllocatedNodeForBMH functions added a "bmh" attribute to log calls when the context already contained one from handleBMHEvent, producing duplicate JSON keys with different formats. Remove the redundant inline attributes and add per-BMH context enrichment in the BuildResourcesForPool loop so both call paths provide consistent "bmh" and "bmhNamespace" via context.


donpenney:
Updated for rebase

PR #2868, opened 2026-06-16
chore: disable CodeRabbit auto-pause after 5 commits

donpenney:
The default auto_pause_after_reviewed_commits threshold of 5 causes CodeRabbit to silently stop reviewing PRs that go through many review iterations, requiring manual `@coderabbitai review` to resume. Set to 0 so reviews run on every push.


donpenney:
Per the following comment:
https://github.com/openshift-kni/oran-o2ims/pull/2693#issuecomment-4719844649

PR #2874, opened 2026-06-16
CNF-24432: Spoke client lifecycle and upgrade type detection for MNO upgrade

Missxiaoguo:
## Summary

Implements the spoke client lifecycle and upgrade type detection for the standard MNO upgrade flow ([CNF-24432](https://redhat.atlassian.net/browse/CNF-24432), Step 1). Adds a reusable `spokeclient` package that creates scoped spoke clients via ManagedServiceAccount + ManifestWork, and refactors `handleUpgrade` to dispatch between ClusterVersion and IBGU upgrade types.

**Jira**: [CNF-24432](https://redhat.atlassian.net/browse/CNF-24432) — Standard MNO upgrade flow

## Changes

**[CNF-24443](https://redhat.atlassian.net/browse/CNF-24443): Upgrade type detection and dispatch**
- Add `detectUpgradeType` that inspects top-level keys in CT `upgradeDefaults` and PR `upgradeParameters` to determine the upgrade method: `imageBasedGroupUpgrade` key dispatches to the IBGU handler, `clusterVersion` key dispatches to the ClusterVersion handler
- Extract existing IBGU logic into `handleIBGUUpgrade`
- Add stub `handleClusterVersionUpgrade` for the MNO upgrade path
- Fail with `PreconditionChecksFailed` on mixed types, missing config, or malformed parameters

**[CNF-24440](https://redhat.atlassian.net/browse/CNF-24440): Spoke client lifecycle**
- New `internal/controllers/utils/spokeclient/` package with `EnsureSpokeClient`, `CleanupSpokeAccess`, `BuildRBACManifestWork`
- Full MSA lifecycle: check addon → create MSA (24h token, auto-rotation) → wait for token sync → deliver RBAC ManifestWork → wait for `WorkAvailable` → build client
- Client caching with token `resourceVersion`-based freshness checks; cache invalidation on token secret deletion
- `EnsureSpokeClient` returns `(client, ready, error)` — `!ready` for transient wait states (no error), `InputError` for precondition failures (addon missing), standard errors for API failures
- `handleClusterVersionUpgrade` calls `EnsureSpokeClient` and sets `Pending` condition when not ready, `PreconditionChecksFailed` on `InputError`, and returns the error to the reconciler for transient API failures
- Add `Pending` condition reason, `UpgradeDefaultsClusterVersionKey` constant
- RBAC markers and scheme registration for MSA, ManifestWork, ManagedClusterAddOn
- Register types in controller manager scheme (`start_controller_manager.go`)

## Acceptance Criteria

- [x] `detectUpgradeType` returns correct type from CT defaults and/or PR params
- [x] Mixed upgrade types (both clusterVersion and imageBasedGroupUpgrade) fail with `PreconditionChecksFailed`
- [x] Missing upgrade configuration fails with `PreconditionChecksFailed`
- [x] Spoke client created via MSA + ManifestWork
- [x] Token rotation detected and client rebuilt automatically
- [x] ManifestWork `WorkAvailable` condition checked before proceeding
- [x] `PreconditionChecksFailed` set when MSA addon is missing
- [x] `Pending` condition set during spoke client preparation
- [x] Existing IBGU upgrade tests pass after refactor

## Test Plan

**Unit tests — spokeclient package (10 tests):**
Covers the full `EnsureSpokeClient` lifecycle: addon not found, token sync waiting, ManifestWork not yet available, happy path (MSA + MW creation through to client ready), client caching and token rotation, cache invalidation on missing secret, cleanup, and `BuildRBACManifestWork` output verification.

**Unit tests — upgrade (16 tests):**
Covers `detectUpgradeType` (11 cases: type detection from CT/PR, mixed types, missing config, malformed inputs), `handleUpgrade` dispatch (3 cases: missing config, missing CT, successful dispatch), and `handleClusterVersionUpgrade` (2 cases: addon missing, token not synced).

**Real cluster testing:**
- Verified MSA creation, token sync, ManifestWork delivery, RBAC applied on spoke
- Verified token rotation with short validity period
- Verified spoke ClusterVersion read via scoped token

## Notes

- `spokeclient` package is designed for reuse by the hardware manager when it adopts MSA-based spoke access
- The `handleClusterVersionUpgrade` stub reads the spoke ClusterVersion but does not trigger upgrades — full upgrade logic (precondition checks, trigger, monitoring) comes in Step 2


test/fakeclient/fakeclient.go
```diff
@@ -65,6 +68,12 @@ func init() {
 	Scheme.AddKnownTypes(assistedservicev1beta1.GroupVersion,
 		&assistedservicev1beta1.Agent{}, &assistedservicev1beta1.AgentList{})
 	Scheme.AddKnownTypes(ibguv1alpha1.SchemeGroupVersion, &ibguv1alpha1.ImageBasedGroupUpgrade{})
+	Scheme.AddKnownTypes(msav1beta1.SchemeGroupVersion,
+		&msav1beta1.ManagedServiceAccount{}, &msav1beta1.ManagedServiceAccountList{})
+	Scheme.AddKnownTypes(workv1.SchemeGroupVersion,
+		&workv1.ManifestWork{}, &workv1.ManifestWorkList{})
+	Scheme.AddKnownTypes(addonv1alpha1.SchemeGroupVersion,
+		&addonv1alpha1.ManagedClusterAddOn{}, &addonv1alpha1.ManagedClusterAddOnList{})
```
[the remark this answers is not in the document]
Missxiaoguo: No spokeclient tests are using the `GetFakeClientFromObjects`. The spokeclient tests that need status subresource handling for these types have added `WithStatusSubresource` explicitly in their own clients. We can add them here when we add the intergration tests later.

bundle/manifests/oran-o2ims.clusterserviceversion.yaml
```diff
@@ -1848,6 +1866,18 @@ spec:
...
+          resources:
+          - manifestworks
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
```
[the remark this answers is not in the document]
Missxiaoguo: Addressed.

Missxiaoguo:
Rebased and resolved conflicts.

internal/controllers/utils/spokeclient/spokeclient.go
```diff
@@ -0,0 +1,410 @@
...
+		return nil, false, nil
+	}
+
+	// 7. Build spoke client.
+	apiServerURL, err := getAPIServerURL(ctx, hubClient, clusterName)
+	if err != nil {
+		return nil, false, err
+	}
+	spokeClient, err := newSpokeClientFunc(apiServerURL, string(tokenSecret.Data["token"]), tokenSecret.Data["ca.crt"], spokeScheme)
+	if err != nil {
```
donpenney, line 215: [Comment authored by Claude]

`tokenSecret.Data["token"]` and `tokenSecret.Data["ca.crt"]` are used without checking they exist. If the secret is malformed (e.g., MSA controller bug, partial sync), this silently passes empty values to `buildSpokeClient`. Consider validating:
```go
tokenData, ok := tokenSecret.Data["token"]
if !ok || len(tokenData) == 0 {
    return nil, false, fmt.Errorf("token secret %s missing 'token' key", tokenSecretName)
}
```
Same pattern applies to the second usage in `refreshCachedSpokeClient` around line 264.

internal/controllers/utils/spokeclient/spokeclient.go
```diff
@@ -0,0 +1,410 @@
...
+func EnsureSpokeClient(
+	ctx context.Context,
+	hubClient client.Client,
+	logger *slog.Logger,
+	clusterName, msaName, mwName string,
+	rules []rbacv1.PolicyRule,
+	spokeScheme *runtime.Scheme,
+) (client.Client, bool, error) {
+	spokeClientsMu.RLock()
+	entry, cached := spokeClients[msaName]
```
donpenney, line 109: [Comment authored by Claude]

The cache check-then-act here isn't atomic — two concurrent reconcilers for different PRs on the same cluster can both detect a token rotation and redundantly rebuild the client. Not a correctness bug (last write wins with identical data), but wasteful. A simple fix would be to hold the write lock for the full rebuild, or add a "rebuild in progress" flag to `spokeClientEntry` to coalesce concurrent rebuilds.

Non-blocking, but worth considering given that multiple PRs can target the same cluster.

internal/controllers/provisioningrequest_upgrade_test.go
```diff
@@ -307,3 +311,311 @@ var _ = Describe("prepareIBGU", func() {
...
+
+		result, proceed, err := task.handleClusterVersionUpgrade(ctx, ct, clusterName)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(proceed).To(BeFalse())
+		Expect(result.RequeueAfter).To(BeNumerically(">", 0))
+
+		condition := meta.FindStatusCondition(task.object.Status.Conditions,
+			string(provisioningv1alpha1.PRconditionTypes.UpgradeCompleted))
+		Expect(condition).ToNot(BeNil())
+		Expect(condition.Reason).To(Equal(string(provisioningv1alpha1.CRconditionReasons.Pending)))
```
donpenney, line 630: [Comment authored by Claude]

The `handleClusterVersionUpgrade` tests cover addon-missing and token-not-synced, but there's no test for the spoke client ready path (where `EnsureSpokeClient` returns a client and the code reads the ClusterVersion). The `SetTestSpokeClientCreator` seam makes this straightforward to add — inject a fake client that returns a ClusterVersion object and verify the log output or condition state.

Also missing: a test where `addon.Status.Namespace` is set to a custom value (vs. the empty-string fallback to `defaultAddonInstallNamespace`).

internal/controllers/provisioningrequest_upgrade_test.go
```diff
@@ -307,3 +311,311 @@ var _ = Describe("prepareIBGU", func() {
...
+		}
+		_, err := detectUpgradeType(ct, pr)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("only one upgrade type is allowed"))
+	})
+
+	It("should return error when no upgrade configuration is provided", func() {
+		_, err := detectUpgradeType(ct, pr)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("no upgrade configuration found"))
```
donpenney, line 414: [Comment authored by Claude]

Minor gap: no test for when both CT and PR specify the *same* upgrade type (e.g., both have `clusterVersion`). The current union logic should handle this correctly (returns `clusterVersion`, not an error), but an explicit test would document that behavior.

Also worth testing: empty `upgradeParameters` map `{}` — should return "no upgrade configuration found".

donpenney (commented):
[Comment authored by Claude]

Nice work on the spoke client architecture — the MSA lifecycle, caching with token rotation detection, and the clean upgrade type dispatch are well-designed. A few items to consider:

**Summary:**
- Token secret data should be validated before use (keys may not exist in a malformed secret)
- Cache check-then-rebuild has a race window between concurrent reconcilers
- A couple of test gaps for the spoke client ready path and addon namespace fallback
- Overall: solid design, RBAC is minimal/correct, logging follows conventions, test coverage is good

internal/controllers/utils/spokeclient/spokeclient.go
Missxiaoguo: You are right, thanks for suggestion. Will address.

internal/controllers/utils/spokeclient/spokeclient.go
donpenney: I don't know how we'd have "different PRs on the same cluster"... I missed this when Claude made these comments

Missxiaoguo: Thanks for the comment, but one ProvisioningRequest targets one cluster. Two concurrent reconcilers for different PRs on the same cluster should not happen. Even if it did happen, there wouldn’t be a race because the cached client is per PR (<pr-name>-upgrade), so each PR would use its own client.

donpenney: [Comment authored by Claude]

You are right on both counts. One PR creates one cluster, so there is no concurrent access to the same cache entry. And even if there were, the per-PR cache key isolates entries. Withdrawing this comment.

internal/controllers/provisioningrequest_upgrade_test.go
Missxiaoguo: > The handleClusterVersionUpgrade tests cover addon-missing and token-not-synced, but there's no test for the spoke client ready path (where EnsureSpokeClient returns a client and the code reads the ClusterVersion).

The happy path was not added because the handleClusterVersionUpgrade is stub and the functionality is not fully implemented. The happy path will be added in the next PR which implements that function. 

>Also missing: a test where addon.Status.Namespace is set to a custom value (vs. the empty-string fallback to defaultAddonInstallNamespace).

will add one

internal/controllers/provisioningrequest_upgrade_test.go
Missxiaoguo: >Minor gap: no test for when both CT and PR specify the same upgrade type (e.g., both have clusterVersion). The current union logic should handle this correctly (returns clusterVersion, not an error), but an explicit test would document that behavior.

Will add one for this.

>Also worth testing: empty upgradeParameters map {} — should return "no upgrade configuration found".

This one is already covered - line399.

internal/controllers/utils/spokeclient/spokeclient.go
Missxiaoguo: Updated!

internal/controllers/provisioningrequest_upgrade_test.go
Missxiaoguo: Updated!

internal/controllers/provisioningrequest_upgrade_test.go
Missxiaoguo: Updated!

internal/controllers/utils/spokeclient/spokeclient_test.go
```diff
@@ -0,0 +1,413 @@
...
+			"test-pr-upgrade", "test-pr-upgrade-rbac", testRules, NewSpokeScheme())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ready).To(BeTrue())
+		Expect(result).ToNot(Equal(oldClient))
+
+		spokeClientsMu.RLock()
+		entry := spokeClients["test-pr-upgrade"]
+		spokeClientsMu.RUnlock()
+		Expect(entry.tokenResourceVersion).To(Equal("100"))
+	})
```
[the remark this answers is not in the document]
Missxiaoguo: [Comment authored by Claude]

This finding is incorrect. The reviewer misread the return semantics of `refreshCachedSpokeClient`.

When a token rotation is detected (resourceVersion changed), `refreshCachedSpokeClient` rebuilds the client **inline** (lines 260-281) and returns `(spokeClient, false, nil)` — where `false` means `needsFullSetup=false`. The caller in `EnsureSpokeClient` then returns the rebuilt client directly without falling through to the full lifecycle setup:

```go
if !needsFullSetup {
    return spokeClient, true, nil
}
```

So the addon check is never reached in this path. The test is correct — it only needs `ManagedCluster` (for API server URL) and the token secret (for the new token), which are both provided.

The full lifecycle setup (addon → MSA → MW) only runs on a true cache miss (no entry at all, or entry invalidated due to missing token secret), not on token rotation.

donpenney (commented):
[Comment authored by Claude]

All review findings addressed:
- Token validation via `extractTokenData()` with tests for missing/empty keys
- Addon namespace fallback test added
- Same upgrade type (CT + PR) test added
- RBAC verbs tightened per CodeRabbit suggestion
- Race condition comment withdrawn (invalid)

No further concerns from my side.

PR #2916, opened 2026-06-18
CNF-24962: remove nonResourceURL permissions from manager-role

alegacy:
## Summary
- Removes all `nonResourceURL` permissions from the `manager-role` ClusterRole. The controller-manager does not call any O2IMS REST API endpoints or the alertmanager webhook — these permissions were historical artifacts from before the service-specific ClusterRoles were separated into dedicated files.
- All service RBAC is deployed by OLM via the CSV bundle, so RBAC escalation prevention does not require the controller-manager to hold these permissions.
- The original security concern ([CNF-24962](https://redhat.atlassian.net/browse/CNF-24962)) that any SA granted the alarms-webhook verb could inject arbitrary alarms is already mitigated by: (1) the ClusterRoleBinding restricting access to only the `alertmanager` SA in `open-cluster-management-observability`, and (2) NetworkPolicy restricting network access to only alertmanager pods.

## Test plan
- [x] Verify `make generate && make manifests && make bundle` produces a clean tree
- [x] Verify `make ci-job` passes
- [x] Deploy operator and confirm alertmanager webhook delivery still works (alertmanager SA has its own dedicated ClusterRole)
- [x] Confirm service-to-service REST API calls still work (each service has its own dedicated ClusterRole)


donpenney (approved):

alegacy:
@coderabbitai review

PR #3002, opened 2026-06-25
CNF-23722: Add Prometheus metrics and alerts for authentication failures

rauhersu:
## Why these changes?

The O2IMS operator and its API service pods had no Prometheus observability. Security-related
events (authentication, authorization, RFC 8705 certificate-binding failures) were only
visible in logs, with no way to alert or trend on them. This PR builds the full metrics
pipeline from instrumentation to alerting.

### 1. Controller-manager metrics scraping

* Uncomment the `../prometheus` kustomize component in `config/default/kustomization.yaml`.
* Add `openshift.io/cluster-monitoring: "true"` label to the operator namespace so the
  platform Prometheus discovers ServiceMonitors.
* Add a ServiceMonitor for the controller-manager pod (port 8443, `/metrics`) with
  TLS verification (`serverName` includes the kustomize `oran-o2ims-` prefix).
* Grant RBAC for `monitoring.coreos.com` resources (servicemonitors, prometheusrules).

### 2. `/metrics` endpoint on all API service pods

* Register `promhttp.Handler()` on each service's existing TLS mux at `/metrics`
  (alarms, artifacts, cluster, provisioning, resources), wrapped with
  Authenticator and Authorizer middleware (TokenReview + SubjectAccessReview).
* Exempt `/metrics` from audience-scoped token validation so the Prometheus
  ServiceAccount's default-audience token is accepted.
* Introduce the `o2ims_auth_failures_total` counter with bounded labels
  (`service`, `type`, `method`, `path`). UUID path segments are normalized to `{id}`
  to prevent cardinality explosion.
* Instrument the three auth failure paths: `Authenticator`, `Authorizer`, and
  RFC 8705 `withClientVerification`.
* Add per-service ServiceMonitors with full TLS certificate verification
  (`serverName` matching the service-ca SAN, `caFile` pointing to the
  service-ca CA bundle mounted in Prometheus).
* Add the `oran-o2ims/metrics` label to API service `Service` objects
  (excludes database and hardware-manager).
* Add NetworkPolicy ingress rules allowing the `openshift-monitoring` namespace
  to scrape service pods.
* Programmatically ensure the namespace monitoring label via the
  `InventoryController` reconciler.
* Bind the `prometheus-k8s` ServiceAccount to the `metrics-reader` ClusterRole
  (grants `GET` on the `/metrics` non-resource URL).
* Unit tests for `NormalizePath`, `RegisterMetricsHandler` middleware ordering,
  `AuthFailures` counter, `Authenticator` metric increments, and
  `withClientVerification` metric increments.

### 3. PrometheusRule alert definitions

* `O2IMSHighAuthFailureRate` — fires when auth failures exceed 10/min sustained over 5 minutes
  (severity: warning).
* `O2IMSCertificateBindingFailure` — fires on any RFC 8705 certificate-binding rejection
  within a 10-minute window (severity: critical), as it may indicate a stolen-token replay.

### 4. Developer documentation

* `docs/dev/prometheus-metrics.md` covering architecture, TLS/authentication/authorization
  flow (with sequence diagram), ServiceMonitor config, metric definitions, alert rules,
  RBAC/NetworkPolicy setup, OLM verification, and alert lifecycle testing steps.

## Test plan

* `make test` — unit tests pass for the new `metrics` package, `auth` instrumentation, and `inventory_controller` changes
* `make test-e2e` — all 21 e2e specs pass (fixed port conflict)
* `make ci-job` — full CI pipeline passes
* Deploy to an OCP cluster and verify:
  * `oc get servicemonitor -n oran-o2ims` shows all monitors
  * `oc get prometheusrule -n oran-o2ims` shows `o2ims-auth-alerts`
  * Prometheus targets page shows all service pods as UP
  * `o2ims_auth_failures_total` metric appears after sending an unauthenticated request
  * `O2IMSHighAuthFailureRate` alert transitions to pending/firing when generating sustained failures
  * TLS scanner confirms all metrics endpoints are TLS-only with cluster-compliant cipher suites

### DIFF
--- a/bundle/manifests/oran-o2ims-artifacts-server-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/oran-o2ims-artifacts-server-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -4,11 +4,9 @@ metadata:
   labels:
     app.kubernetes.io/component: metrics
     app.kubernetes.io/created-by: oran-o2ims
-    app.kubernetes.io/instance: o2ims-services-metrics-monitor
     app.kubernetes.io/managed-by: kustomize
-    app.kubernetes.io/name: servicemonitor
     app.kubernetes.io/part-of: oran-o2ims
-  name: oran-o2ims-o2ims-services-metrics-monitor
+  name: oran-o2ims-artifacts-server-metrics-monitor
 spec:
   endpoints:
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
@@ -16,7 +14,8 @@ spec:
     port: api
     scheme: https
     tlsConfig:
-      insecureSkipVerify: true
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      serverName: artifacts-server.oran-o2ims.svc
   selector:
     matchLabels:
-      oran-o2ims/metrics: "true"
+      app: artifacts-server
--- a/bundle/manifests/oran-o2ims-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/oran-o2ims-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -1,0 +1,23 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/instance: controller-manager-metrics-monitor
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: servicemonitor
+    app.kubernetes.io/part-of: oran-o2ims
+    control-plane: controller-manager
+  name: oran-o2ims-controller-manager-metrics-monitor
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
+  selector:
+    matchLabels:
+      control-plane: controller-manager
--- a/bundle/manifests/oran-o2ims-o2ims-services-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/oran-o2ims-o2ims-services-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -1,0 +1,22 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/instance: o2ims-services-metrics-monitor
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: servicemonitor
+    app.kubernetes.io/part-of: oran-o2ims
+  name: oran-o2ims-o2ims-services-metrics-monitor
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: api
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
+  selector:
+    matchLabels:
+      oran-o2ims/metrics: "true"
--- a/config/prometheus/services-monitor.yaml
+++ b/config/prometheus/services-monitor.yaml
@@ -1,0 +1,24 @@
+# Prometheus Monitor for O2IMS service pods (alarms, cluster, resource, provisioning, artifacts)
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/name: servicemonitor
+    app.kubernetes.io/instance: o2ims-services-metrics-monitor
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: o2ims-services-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
+  selector:
+    matchLabels:
+      oran-o2ims/metrics: "true"
--- a/config/prometheus/services-monitor.yaml
+++ b/config/prometheus/services-monitor.yaml
@@ -1,15 +1,16 @@
-# Prometheus Monitor for O2IMS service pods (alarms, cluster, resource, provisioning, artifacts)
+# Per-service Prometheus ServiceMonitors for O2IMS API pods.
+# Each service gets its own monitor so that tlsConfig.serverName matches
+# the serving certificate SAN (<service-name>.<namespace>.svc).
+---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    app.kubernetes.io/name: servicemonitor
-    app.kubernetes.io/instance: o2ims-services-metrics-monitor
     app.kubernetes.io/component: metrics
     app.kubernetes.io/created-by: oran-o2ims
     app.kubernetes.io/part-of: oran-o2ims
     app.kubernetes.io/managed-by: kustomize
-  name: o2ims-services-metrics-monitor
+  name: alarms-server-metrics-monitor
   namespace: system
 spec:
   endpoints:
@@ -18,7 +19,100 @@ spec:
       scheme: https
       bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
       tlsConfig:
-        insecureSkipVerify: true
+        serverName: alarms-server.oran-o2ims.svc
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
   selector:
     matchLabels:
-      oran-o2ims/metrics: "true"
+      app: alarms-server
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: artifacts-server-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        serverName: artifacts-server.oran-o2ims.svc
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+  selector:
+    matchLabels:
+      app: artifacts-server
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: cluster-server-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        serverName: cluster-server.oran-o2ims.svc
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+  selector:
+    matchLabels:
+      app: cluster-server
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: provisioning-server-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        serverName: provisioning-server.oran-o2ims.svc
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+  selector:
+    matchLabels:
+      app: provisioning-server
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: resource-server-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        serverName: resource-server.oran-o2ims.svc
+        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+  selector:
+    matchLabels:
+      app: resource-server
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -267,6 +267,18 @@ rules:
   resources:
   - prometheusrules
   verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  verbs:
   - get
   - list
   - watch
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -266,17 +266,6 @@ rules:
   - monitoring.coreos.com
   resources:
   - prometheusrules
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - monitoring.coreos.com
-  resources:
   - servicemonitors
   verbs:
   - get
--- a/docs/dev/prometheus-metrics.md
+++ b/docs/dev/prometheus-metrics.md
@@ -1,0 +1,722 @@
+# Prometheus Metrics
+
+This document describes how Prometheus metrics are configured and collected
+for the O-Cloud Manager operator **and** its O2IMS service pods.
+
+## Table of Contents
+
+- [Architecture](#architecture)
+- [Configuration Files](#configuration-files)
+- [TLS, Authentication and Authorization](#tls-authentication-and-authorization)
+  - [Audience-scoped token exemption for /metrics](#audience-scoped-token-exemption-for-metrics)
+- [Network Policies](#network-policies)
+- [RBAC](#rbac)
+- [Operator Metrics (controller-manager)](#operator-metrics-controller-manager)
+- [Service Pod Metrics (O2IMS API Servers)](#service-pod-metrics-o2ims-api-servers)
+- [Alert Rules (PrometheusRule)](#alert-rules-prometheusrule)
+- [End-to-End Verification](#end-to-end-verification)
+- [Adding New Metrics](#adding-new-metrics)
+
+## Architecture
+
+The O-Cloud Manager exposes Prometheus metrics at two layers:
+
+1. **Operator layer** — The `controller-manager` pod uses controller-runtime's
+   built-in metrics server on port 6443, serving reconciliation and workqueue
+   metrics.
+2. **Service layer** — Each O2IMS API service pod (`alarms-server`,
+   `cluster-server`, `provisioning-server`, `resource-server`,
+   `artifacts-server`) exposes application-level metrics on its existing TLS
+   port 8443 at `/metrics`.
+
+Both layers are scraped by the OpenShift platform Prometheus via dedicated
+`ServiceMonitor` resources.
+
+```mermaid
+flowchart LR
+    subgraph oran_ns [oran-o2ims namespace]
+        CM[controller-manager :6443 /metrics]
+        RS[resource-server :8443 /metrics]
+        CS[cluster-server :8443 /metrics]
+        AS[alarms-server :8443 /metrics]
+        PS[provisioning-server :8443 /metrics]
+        ARS[artifacts-server :8443 /metrics]
+        SM1[ServiceMonitor operator]
+        SM2[ServiceMonitors per-service]
+    end
+
+    subgraph monitoring [openshift-monitoring]
+        Prom[Prometheus]
+    end
+
+    Prom -->|scrapes| SM1
+    Prom -->|scrapes| SM2
+    SM1 --> CM
+    SM2 --> RS
+    SM2 --> CS
+    SM2 --> AS
+    SM2 --> PS
+    SM2 --> ARS
+```
+
+For the platform Prometheus instance (in `openshift-monitoring`) to discover
+ServiceMonitors in the operator namespace, the namespace must carry the label
+`openshift.io/cluster-monitoring: "true"`. The Inventory controller ensures
+this label is present on the operator namespace at every reconciliation cycle.
+
+## Configuration Files
+
+| File | Purpose |
+|------|---------|
+| `config/prometheus/monitor.yaml` | ServiceMonitor for the operator (controller-manager) |
+| `config/prometheus/services-monitor.yaml` | Per-service ServiceMonitors for O2IMS API service pods |
+| `config/prometheus/auth-alerts.yaml` | PrometheusRule with authentication failure alerts |
+| `config/prometheus/kustomization.yaml` | Kustomize resource list for the prometheus overlay |
+| `config/rbac/metrics_service.yaml` | Service exposing port 8443 with OCP serving-cert annotation |
+| `config/rbac/metrics_service_clusterrole.yaml` | ClusterRole granting GET on `/metrics` non-resource URL |
+| `config/rbac/clusterrolebindings_services.yaml` | ClusterRoleBinding for Prometheus SA to `metrics-reader` |
+| `config/default/kustomization.yaml` | Includes `../prometheus` to deploy both ServiceMonitors and PrometheusRule |
+
+---
+
+## TLS, Authentication and Authorization
+
+The service pod `/metrics` endpoints are served over TLS on the same port
+(8443) as the API endpoints. They inherit the cluster-wide TLS security
+profile from
+`apiservers.config.openshift.io/cluster`, ensuring cipher suites and minimum
+TLS version comply with the cluster administrator's policy.
+
+### How TLS is applied to metrics
+
+The `/metrics` handler shares the HTTP server's `tls.Config`, which is
+configured at startup via `GetServerTLSConfig()`. This means:
+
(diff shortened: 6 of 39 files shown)


rauhersu:
@coderabbitai I already skipped some comments in https://github.com/openshift-kni/oran-o2ims/pull/2910/changes. Please, don't show them again if I decided not to address them previously.

rauhersu:
I will assess coderabbit comments

/hold

bundle/manifests/oran-o2ims-controller-manager-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -0,0 +1,23 @@
...
+    control-plane: controller-manager
+  name: oran-o2ims-controller-manager-metrics-monitor
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: https
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — this `insecureSkipVerify: true` comes from the pre-existing `config/prometheus/monitor.yaml`, which was scaffolded with the operator (MGMT-16144) and was not modified in this PR. It appears in the bundle diff only because uncommenting the `../prometheus` kustomize overlay now pulls it in. This is pre-existing tech debt, not something introduced here.

bundle/manifests/oran-o2ims-o2ims-services-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app.kubernetes.io/part-of: oran-o2ims
+  name: oran-o2ims-o2ims-services-metrics-monitor
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: api
+    scheme: https
+    tlsConfig:
+      insecureSkipVerify: true
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — this is an intentional design choice, documented in `docs/dev/prometheus-metrics.md` (lines 128-138). The connection is encrypted via OCP serving certs (`service.beta.openshift.io/serving-cert-secret-name`), and access is gated at the network layer by NetworkPolicy restricting ingress to the `openshift-monitoring` namespace. The risk is limited to in-cluster MITM between those two namespaces, which is already constrained. We acknowledge this as a potential future hardening item but consider it out of scope for this PR.

config/prometheus/services-monitor.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: o2ims-services-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: true
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — same rationale as the bundle manifest comment. This is the source file and the design is intentional and documented. See the reply on comment #2.

config/rbac/role.yaml
```diff
@@ -267,6 +267,18 @@ rules:
   resources:
   - prometheusrules
   verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — the write verbs on the PrometheusRule RBAC marker are intentional. While no controller code currently creates/updates PrometheusRule objects directly (the resource is deployed via the kustomize bundle), the write permissions are included deliberately to support future reconciler-managed alert rules without requiring an RBAC change. The `role.yaml` is auto-generated from `kubebuilder:rbac` markers via `make manifests`.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,551 @@
...
+## Alert Rules (PrometheusRule)
+
+A `PrometheusRule` resource (`config/prometheus/auth-alerts.yaml`) defines
+alerting rules for authentication failures. These are evaluated by the
+platform Prometheus and route through AlertManager.
+
+| Alert | Severity | Condition | Description |
+|-------|----------|-----------|-------------|
+| `O2IMSHighAuthFailureRate` | warning | `rate(o2ims_auth_failures_total[5m]) > 0.1667` sustained for 5 min (>10/min) | A service is experiencing a high rate of authentication or authorization failures |
+| `O2IMSCertificateBindingFailure` | critical | Any `certificate_binding` failure in the last 10 min | RFC 8705 client certificate mismatch — potential stolen token replay |
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Accepting — the documentation table was showing a simplified expression that didn't match the actual PrometheusRule. The deployed alert in `config/prometheus/auth-alerts.yaml` already uses the correct `sum(rate(o2ims_auth_failures_total[5m])) by (service, type) > 0.1667`. Fixed the docs table to match the real expression.

internal/controllers/inventory_controller.go
```diff
@@ -648,6 +650,11 @@ func (t *reconcilerTask) run(ctx context.Context) (nextReconcile ctrl.Result, er
 			return
 		}
 	}
+
+	err = t.ensureNamespaceMonitoringLabel(ctx)
+	if err != nil {
+		ctlrutils.LogError(ctx, t.logger, "Failed to ensure namespace monitoring label; continuing reconciliation", err)
+	}
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — this is an intentional design choice. The monitoring namespace label is a best-effort, non-critical operation. The log message explicitly says "continuing reconciliation" to make the intent clear. Returning the error here would block deployment of critical infrastructure (databases, services, deployments) over a non-critical monitoring label failure — that's disproportionate. The label will be retried on the next reconciliation cycle (~5 minutes).

internal/controllers/inventory_controller.go
```diff
@@ -1106,12 +1128,16 @@ func (t *reconcilerTask) createServiceAccount(ctx context.Context, resourceName
 func (t *reconcilerTask) createService(ctx context.Context, resourceName string, port int32, targetPort string) error {
 	t.logger.DebugContext(ctx, "[createService]")
 	// Build the Service object.
+	labels := map[string]string{
+		"app": resourceName,
+	}
+	if resourceName != ctlrutils.InventoryDatabaseServerName && resourceName != ctlrutils.HardwareManagerServerName {
+		labels["oran-o2ims/metrics"] = "true"
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — `HasMetrics()` and the `oran-o2ims/metrics: "true"` label serve different purposes. `HasMetrics()` returns true for the hardware manager because it exposes controller-runtime metrics on its own port (6443). However, the `oran-o2ims/metrics` label is specifically for the services ServiceMonitor, which scrapes the application `/metrics` endpoint on port 8443. The hardware manager uses a different metrics pattern (controller-runtime, like the main controller-manager) and is not scraped by this ServiceMonitor. Using `HasMetrics()` here would make the services ServiceMonitor try to scrape the hardware manager on the wrong port. This is explained in `docs/dev/prometheus-metrics.md` (lines 195-197).

internal/service/common/metrics/metrics.go
```diff
@@ -0,0 +1,43 @@
...
+
+var uuidSegmentRe = regexp.MustCompile(
+	`/[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`,
+)
+
+// NormalizePath replaces UUID path segments with {id} to keep metric
+// cardinality bounded. Paths without UUIDs pass through unchanged.
+func NormalizePath(path string) string {
+	return uuidSegmentRe.ReplaceAllString(path, "/{id}")
+}
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — the only non-UUID path parameter is `{hwVendorName}` on the internal endpoint `/internal/v1/hardware-alerts/{hwVendorName}`, and the set of vendor names is inherently tiny (dell, hpe, supermicro, etc.), so cardinality is bounded in practice. Using route templates from the router context (e.g., `req.Pattern()` in Go 1.22+) would be a nice improvement but is out of scope for this PR. We may revisit this as a follow-up.

rauhersu:
Ready for review, ptal @alegacy @donpenney 

/unhold

config/prometheus/services-monitor.yaml
alegacy: Sorry, I'm a bit late to the party, but could you explain why skip verify is required here?  wouldn't it be hitting our cluster generated tls cert and wouldn't it have that CA in its trusted list so shouldn't this pass?

config/prometheus/services-monitor.yaml
```diff
@@ -0,0 +1,24 @@
...
+    app.kubernetes.io/part-of: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+  name: o2ims-services-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: api
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
alegacy, line 20: I assume this is telling prometheus to use its own sa token file to access our api?  Beware, we've recently made changes to our auth handlers and we'll be looking for specific audiences.  I haven't yet looked at how this /metric endpoint has been wired in, but this may be an issue so needs to be retested on latest in main.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,551 @@
...
+### RBAC for metrics
+
+The **operator metrics** (controller-manager) are protected by controller-runtime's
+built-in authn/authz middleware, which validates bearer tokens against the
+kube-apiserver. The `metrics-reader` ClusterRole grants GET on the `/metrics`
+non-resource URL so that Prometheus's ServiceAccount token is authorized.
+
+The **service pod metrics** do **not** require additional RBAC. The `/metrics`
+handler is registered on the base mux before the OAuth/SubjectAccessReview
+middleware chain, so no application-level authentication or authorization
```
alegacy: We need to tighten this up so that the endpoint is protected by authn/authz otherwise it will get flagged on our next scan.

internal/controllers/utils/utils.go
```diff
@@ -210,11 +210,10 @@ func HasApiEndpoints(serverName string) bool {
 		serverName == InventoryProvisioningServerName
 }
 
-// HasMetrics determines whether a server exposes a metrics endpoint that
-// requires TLS. Servers with API endpoints also serve metrics, but this
-// function identifies servers that serve metrics without API endpoints.
+// HasMetrics determines whether a server exposes a /metrics endpoint.
+// All API servers and the hardware manager server expose metrics.
```
alegacy: [Comment authored by Claude]

`HasApiEndpoints` includes `InventoryDatabaseServerName`, so after this change `HasMetrics("postgres-server")` returns `true` — but the database doesn't expose a `/metrics` endpoint.

No functional impact today since `HasMetrics` is only called by `NeedsServingCert` (which already covered the database via `HasApiEndpoints`), but a future caller relying on `HasMetrics` to mean "exposes `/metrics`" would get a wrong answer for the database.

The updated comment ("All API servers and the hardware manager server expose metrics") reinforces the incorrect assumption.

Consider excluding the database explicitly:

```go
func HasMetrics(serverName string) bool {
    return (HasApiEndpoints(serverName) && serverName != InventoryDatabaseServerName) ||
        serverName == HardwareManagerServerName
}
```

Or alternatively, just add a note to the comment acknowledging the database exception.

internal/service/common/metrics/metrics_test.go
```diff
@@ -0,0 +1,64 @@
+/*
+SPDX-FileCopyrightText: Red Hat
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package metrics
```
alegacy, line 7: [Comment authored by Claude]

Nit: This file uses standard Go `testing` while the rest of the test files in this PR and codebase use Ginkgo (`Describe`/`It`/`Expect`). Consider converting to Ginkgo for consistency.

config/rbac/role.yaml
```diff
@@ -267,6 +267,18 @@ rules:
   resources:
   - prometheusrules
   verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
```
alegacy: [Comment authored by Claude]

The PrometheusRule is deployed statically via kustomize/OLM — the operator doesn't appear to create or modify PrometheusRules programmatically. If that's the case, `get/list/watch` (matching the `servicemonitors` verbs below) would be sufficient here, and would follow the principle of least privilege.

If there's a plan for dynamic rule management, then the broader verbs make sense — just worth confirming the intent.

config/prometheus/services-monitor.yaml
rauhersu: [Comment authored by Claude]

Good question. In principle Prometheus _could_ validate the server cert since it's signed by OpenShift's service-ca. However, the platform Prometheus instance doesn't automatically trust the service-ca signer — its CA bundle doesn't include the service-ca root by default. To do proper verification we'd need to inject the service-ca CA bundle into the ServiceMonitor via a \`tlsConfig.ca\` reference (e.g., from the \`service-ca.crt\` ConfigMap) and set the \`serverName\`.

That said, the bigger concern you and @alegacy raised (comments #2 and #3) was about authn/authz on the endpoint itself, not just TLS verification. We've addressed that: the \`/metrics\` handler is now wrapped with the same TokenReview + SubjectAccessReview middleware chain as API endpoints, and the Prometheus SA is bound to the \`metrics-reader\` ClusterRole via a new \`prometheus-metrics-reader\` ClusterRoleBinding.

TLS certificate verification (replacing \`insecureSkipVerify\`) is a separate hardening item we can follow up on.

config/prometheus/services-monitor.yaml
rauhersu: [Comment authored by Claude]

Good catch on the audience concern. To clarify the wiring: previously the \`/metrics\` endpoint was registered on the base mux _before_ the authn/authz middleware chain, so the bearer token was sent but never validated.

We've now fixed this — the \`/metrics\` handler is wrapped with the same \`Authenticator\` and \`Authorizer\` middleware used for API endpoints. The Kubernetes authenticator performs a TokenReview which validates the token, and the authorizer performs a SubjectAccessReview against the \`metrics-reader\` ClusterRole. The \`bearerTokenFile\` in the ServiceMonitor is now functionally required, not just cosmetic.

Regarding the audience changes you mentioned: after rebasing on latest main, we confirmed that all services now configure audience-scoped TokenReview (e.g., \`--audience=alarms-server\`). The Prometheus SA token uses the default \`api\` audience, so \`/metrics\` is added to \`AudienceExemptPaths\` in each service. This bypasses the audience constraint while still requiring valid authentication (TokenReview) and authorization (SubjectAccessReview) — same pattern used for the alertmanager endpoints marked with \`x-skip-audience-validation\`.

docs/dev/prometheus-metrics.md
rauhersu: [Comment authored by Claude]

Accepted. The \`/metrics\` endpoint is now protected by the same TokenReview + SubjectAccessReview middleware chain as API endpoints. The \`RegisterMetricsHandler\` function wraps \`promhttp.Handler()\` with the \`authn\` and \`authz\` middleware. The Prometheus ServiceAccount (\`prometheus-k8s\`) is bound to the existing \`metrics-reader\` ClusterRole via a new \`prometheus-metrics-reader\` ClusterRoleBinding.

Updated the documentation accordingly — the "Bypasses auth middleware" language is replaced with the new auth-protected design.

internal/controllers/utils/utils.go
rauhersu: [Comment authored by Claude]

Good catch. Fixed — \`HasMetrics\` now excludes \`InventoryDatabaseServerName\` explicitly:

\`\`\`go
func HasMetrics(serverName string) bool {
    return (HasApiEndpoints(serverName) && serverName != InventoryDatabaseServerName) ||
        serverName == HardwareManagerServerName
}
\`\`\`

No functional impact on \`NeedsServingCert\` since the database still returns \`true\` from the \`HasApiEndpoints\` short-circuit. Updated the comment and added a test asserting \`HasMetrics(InventoryDatabaseServerName)\` returns \`false\`.

internal/service/common/metrics/metrics_test.go
rauhersu: [Comment authored by Claude]

Fair point. Converted to Ginkgo with a \`DescribeTable\`/\`Entry\` structure and added a \`suite_test.go\` bootstrap file, matching the pattern used throughout the codebase.

config/rbac/role.yaml
rauhersu: [Comment authored by Claude]

Agreed. The PrometheusRule is deployed statically via kustomize — the operator doesn't manage it programmatically. Reduced the kubebuilder marker to \`verbs=get;list;watch\` and regenerated manifests. The controller-gen merged the \`prometheusrules\` and \`servicemonitors\` entries in \`role.yaml\` since they now share the same verbs.

config/prometheus/services-monitor.yaml
rauhersu: [Raul's comments]

@alegacy , thanks for the feedback. For this specific item I had previously assumed that being this an in-cluster communication and with the proper policies (allow only traffic from openshift-monitoring namespace) it might be all, relaxing our security for metrics. Coderabbit had also detected this issue and that was my previous  answer:

> this is an intentional design choice, documented in docs/dev/prometheus-metrics.md (lines 128-138). The connection is encrypted via OCP serving certs (service.beta.openshift.io/serving-cert-secret-name), and access is gated at the network layer by NetworkPolicy restricting ingress to the openshift-monitoring namespace. The risk is limited to in-cluster MITM between those two namespaces, which is already constrained. We acknowledge this as a potential future hardening item but consider it out of scope for this PR.

We might indeed want to remove the 'insecureSkipVerify' and strengthen our security but have just seen Claude insisted on managing this separately (new PR). Let me know if we want to fix this here.

I will hold this PR until this is clarified and I am able to test metrics in my cluster with all the new code pushed after your comments.

/hold

config/prometheus/services-monitor.yaml
rauhersu: @alegacy the codebase you commented here had all your previous commits included but the July 6th commit:

https://github.com/openshift-kni/oran-o2ims/commits/main/?author=alegacy

Apparently there was no problem when I deployed this in my cluster and checked the counters with that codebase. Is there any commit from the link above I should take a look specially ? Note that this metrics logic was included previous to authn/authz in the middleware chain as Claude commented above so that might be the reason not to have any feature impact here previously.

config/prometheus/services-monitor.yaml
rauhersu: All metrics endpoints are now TLS secured.

docs/dev/prometheus-metrics.md
rauhersu: All metrics endpoints are now wrapped up by authn/authz.

config/rbac/role.yaml
rauhersu: fixed to 'read only'.

internal/service/common/auth/auth.go
```diff
@@ -119,6 +123,7 @@ func Authenticator(oauthHandler, kubernetesHandler authenticator.Request) middle
 				args := []any{slog.Any("error", err), slog.String("method", req.Method), slog.String("path", req.URL.Path)}
 				args = append(args, tokenClaimsAttrs(req)...)
 				slog.WarnContext(req.Context(), "authentication failed", args...)
+				metrics.AuthFailures.WithLabelValues(ServiceName, "authentication", req.Method, metrics.NormalizePath(req.URL.Path)).Inc()
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]


This is a false positive. CodeRabbit correctly identifies that `authn`/`authz` run before `OpenAPIValidation` in the middleware chain, but misses that the **oapi-codegen generated route wrappers** validate path parameters (`Format: "uuid"`) *before* the middleware chain is even constructed:

```go
// inventory.generated.go — runs BEFORE middleware chain
err = runtime.BindStyledParameterWithOptions(..., Format: "uuid")
if err != nil {
    siw.ErrorHandlerFunc(w, r, &InvalidParamFormatError{...})
    return  // middleware chain never executes
}
```

Non-UUID values get a 400 from the generated wrapper and never reach `authn`/`authz`. The only paths that can hit `AuthFailures.Inc()` are static collection paths or paths with valid UUIDs — both already bounded by `NormalizePath` (which replaces UUIDs with `{id}`). Verified on a live cluster.

config/prometheus/services-monitor.yaml
rauhersu: > I haven't yet looked at how this /metric endpoint has been wired in, but this may be an issue so needs to be retested on latest in main.


@alegacy , have included a detailed description of how the metrics logic is integrated in the middleware chain, including the 'audience' management. Ptal at `prometheus-metrics.md`, section 'TLS, Authentication and Authorization'. I was able to successfully scrape metrics and generate alerts with that setup in  my cluster.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,721 @@
...
+```
+
+Expected output (rate > 0.1667):
+
+```json
+{
+  "service": "resource-server",
+  "type": "authentication",
+  "rate_per_sec": "0.29"
+}
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — the \`0.29\` value was observed on an actual cluster. The \`rate()\` function computes a per-second average over a 5-minute window, not an instantaneous rate. At the ~2 minute mark (when the doc says to check), the 5-minute window is only partially filled with data points, so \`rate()\` extrapolates over the full range, producing a value well below the instantaneous 0.5 req/s. The doc already states the pass criterion as "rate > 0.1667" and the JSON block is example output, not an exact expected value.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,721 @@
...
+```
+
+Expected lifecycle (assuming no prior failures in the 5-minute rate window):
+
+| Time after load starts | Expected state | Explanation |
+|------------------------|---------------|-------------|
+| 0–2 min | `inactive` | Rate is still below threshold (building up in the 5m window) |
+| ~2 min | `pending` | Rate crosses 0.1667/s; the `for: 5m` timer starts |
+| ~7 min (2 + 5) | `firing` | Rate has been above threshold for 5 consecutive minutes |
+| ~12 min after load stops | `inactive` | The 5m rate window clears and drops below threshold |
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Accepting — the \`~12 min after load stops\` was incorrect. The 5-minute rate window clears ~5–7 minutes after the last failure, consistent with Step 5 (line 549). Fixed to \`~5–7 min after load stops\`.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,721 @@
...
+| `config/default/kustomization.yaml` | Includes `../prometheus` to deploy both ServiceMonitors and PrometheusRule |
+
+---
+
+## TLS, Authentication and Authorization
+
+All `/metrics` endpoints are served over TLS on the same port (8443) as the
+API endpoints. They inherit the cluster-wide TLS security profile from
+`apiservers.config.openshift.io/cluster`, ensuring cipher suites and minimum
+TLS version comply with the cluster administrator's policy.
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Accepting — the paragraph incorrectly implied all \`/metrics\` endpoints use port 8443, but the controller-manager uses port 6443. Scoped the statement to "service pod \`/metrics\` endpoints" to match the architecture section.

internal/controllers/inventory_controller_test.go
```diff
@@ -340,6 +340,19 @@ var _ = Describe("Inventory Controller", func() {
...
+						for _, peer := range rule.From {
+							if peer.NamespaceSelector != nil && peer.PodSelector == nil {
+								if peer.NamespaceSelector.MatchLabels["network.openshift.io/policy-group"] == "monitoring" {
+									return true
+								}
+							}
+						}
+					}
+					return false
+				}
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — \`hasMonitoringRule\` intentionally checks only the namespace selector ("does this NetworkPolicy allow traffic from the monitoring namespace?"). The port is correct by construction in \`createNetworkPolicy()\`, and the full NetworkPolicy spec (including ports) is validated elsewhere. Adding port assertions to this helper would conflate two concerns and make tests brittle if port constants change.

bundle/manifests/oran-o2ims-artifacts-server-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    path: /metrics
+    port: api
+    scheme: https
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      serverName: artifacts-server.oran-o2ims.svc
+  selector:
+    matchLabels:
+      app: artifacts-server
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — the ServiceMonitor and the target Services are deployed into the same namespace. The \`namespace: system\` in the source YAML is a kustomize placeholder replaced by the namespace transformer; in the bundle manifest (what OLM deploys) there is no \`metadata.namespace\` — OLM installs it into the operator namespace (\`oran-o2ims\`). When a ServiceMonitor omits \`namespaceSelector\`, Prometheus discovers matching Services in the ServiceMonitor's own namespace, which is correct here. Adding \`namespaceSelector.matchNames\` would be redundant.

docs/dev/prometheus-metrics.md
```diff
@@ -0,0 +1,722 @@
...
+
+This is safe because:
+
+1. The Prometheus SA token is only mounted inside the `openshift-monitoring`
+   namespace (pod-level isolation).
+2. Network access is restricted by NetworkPolicy (only `openshift-monitoring`
+   can reach port 8443).
+3. RBAC limits the token to `GET /metrics` only (minimal privilege via
+   `metrics-reader` ClusterRole).
+4. The token is still validated by TokenReview (cannot be forged).
```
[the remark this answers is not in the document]
rauhersu: [Comment authored by Claude]

Rejecting — using the label selector \`network.openshift.io/policy-group: monitoring\` rather than a hardcoded namespace name is the standard OpenShift pattern for allowing monitoring traffic. In a default OCP cluster, only \`openshift-monitoring\` carries this label. The doc accurately describes this on lines 231-233: "any namespace labeled... In a standard OpenShift cluster, the \`openshift-monitoring\` namespace carries this label." There is no inconsistency — the doc describes the label-based selector and identifies which namespace matches it.

rauhersu:
Ready for review again

/unhold

alegacy (commented):
/lgtm

internal/controllers/utils/utils.go
```diff
@@ -209,11 +209,11 @@ func HasApiEndpoints(serverName string) bool {
 		serverName == InventoryProvisioningServerName
 }
 
-// HasMetrics determines whether a server exposes a metrics endpoint that
-// requires TLS. Servers with API endpoints also serve metrics, but this
-// function identifies servers that serve metrics without API endpoints.
+// HasMetrics determines whether a server exposes a /metrics endpoint.
+// All API servers except the database and the hardware manager server expose metrics.
```
donpenney, line 213: This comment change seems inaccurate. The original comment is showing this function is to identify a server that serves metrics, but without API endpoints, and that was just the hwmgr, and was introduced to decouple metrics TLS from the HasApiEndpoints function (PR #2451). The function has been expanded here to include those servers with API endpoints, but the comment is claiming the hwmgr doesn't expose metrics - while at the same time, it explicitly includes the hwmgr as a service with metrics

rauhersu: [Comment authored by Claude]

Good catch — the comment was ambiguous. "except the database and the hardware manager" reads as if both are excluded, but the hardware manager IS included (the `||` branch on line 216 explicitly adds it).

The function now answers a broader question than its original PR #2451 version: "does this server expose `/metrics`?" — which is true for the 5 API servers + hardware manager, and false for the database.

Fixed the comment to: "This includes all API servers except the database, plus the hardware manager server."

See https://github.com/openshift-kni/oran-o2ims/pull/3412

PR #3029, opened 2026-06-29
fix: pin delve debugger to v1.27.0 in Dockerfile

donpenney:
Replace @latest with a pinned version to ensure reproducible builds and address security scan findings about unpinned dependencies.


donpenney:
We'll leave this as `latest`, and only pin it when necessary.

PR #3234, opened 2026-07-06
CNF-22692: support AllNamespaces install mode for OLMv1

donpenney:
Enable AllNamespaces install mode in the CSV to support OLMv1, which requires operators to run in AllNamespaces mode. The controller-runtime Manager cache is already cluster-scoped (no namespace filter), and RBAC uses ClusterRoleBindings, so no code changes are needed beyond the CSV and the Makefile catalog-deploy target.

Both OwnNamespace and AllNamespaces are now supported, maintaining backward compatibility with existing OLMv0 installations.


donpenney:
/test scorecard

The failure appears to be a networking issue, or possibly an issue with the go registry

donpenney:
/override ci/prow/scorecard

Test passes when run manually

PR #3389, opened 2026-07-15
CNF-25886: add MNO worker node scale-out support (Phase 1)

donpenney:
Add support for scaling out worker nodes on provisioned MNO clusters by updating the ProvisioningRequest nodes array.

Webhook validation:
- Reject node scaling while a cluster upgrade is in progress

Controller validation:
- Reject scaling on single-node clusters (after CI rendering where node count is known)
- Reject adding or removing control-plane nodes (workers only)

Hardware manager:
- Detect NodeGroup.Size increases on already-provisioned NARs in handleNodeAllocationRequestSpecChanged and re-enter the allocation flow to provision additional nodes

E2E test:
- Provision a 3-master + 2-worker cluster with 1 extra BMH in pool
- Scale out by adding a third worker via new CT version
- Verify NAR worker Size increases, new AllocatedNode is created, and master Size is unchanged

Documentation:
- Add "Scaling Worker Nodes" section to the Day-2 configuration guide

### DIFF
--- a/api/provisioning/v1alpha1/provisioningrequest_webhook.go
+++ b/api/provisioning/v1alpha1/provisioningrequest_webhook.go
@@ -15,6 +15,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/openshift-kni/oran-o2ims/internal/constants"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -234,11 +235,14 @@ func (v *provisioningRequestValidator) validateCreateOrUpdate(ctx context.Contex
 		}
 
 		if len(scalingNodes) > 0 {
-			// Reject scaling while an upgrade is in progress
+			// Reject scaling while an upgrade is active or in a non-healthy state.
+			// Check Status != True (rather than Reason == InProgress) to also block
+			// scaling during Pending, Unknown, Failed, TimedOut, and
+			// PreconditionChecksFailed states when the cluster may not be healthy.
 			upgradeCond := meta.FindStatusCondition(
 				newPr.Status.Conditions, string(PRconditionTypes.UpgradeCompleted))
-			if upgradeCond != nil && upgradeCond.Reason == string(CRconditionReasons.InProgress) {
-				return fmt.Errorf("node scaling is not supported while a cluster upgrade is in progress")
+			if upgradeCond != nil && upgradeCond.Status != metav1.ConditionTrue {
+				return fmt.Errorf("node scaling is not supported while a cluster upgrade is in progress or incomplete")
 			}
 		}
 
--- a/api/provisioning/v1alpha1/provisioningrequest_webhook.go
+++ b/api/provisioning/v1alpha1/provisioningrequest_webhook.go
@@ -226,11 +226,20 @@ func (v *provisioningRequestValidator) validateCreateOrUpdate(ctx context.Contex
 	} else if crProvisionedCond.Reason == string(CRconditionReasons.Completed) {
 		// Allow specific fields and node scaling after installation completes
 		// This enables Day 2 operations like annotation/label updates and cluster scaling
-		disallowedFields, _, err = FindClusterInstanceImmutableFieldUpdates(
+		disallowedFields, scalingNodes, err = FindClusterInstanceImmutableFieldUpdates(
 			oldPrClusterInstanceInput.(map[string]any), newPrClusterInstanceInput.(map[string]any),
 			[][]string{}, AllowedClusterInstanceFields)
 		if err != nil {
 			return fmt.Errorf("failed to find immutable field updates for ClusterInstance (%s): %w", newPr.Name, err)
+		}
+
+		if len(scalingNodes) > 0 {
+			// Reject scaling while an upgrade is in progress
+			upgradeCond := meta.FindStatusCondition(
+				newPr.Status.Conditions, string(PRconditionTypes.UpgradeCompleted))
+			if upgradeCond != nil && upgradeCond.Reason == string(CRconditionReasons.InProgress) {
+				return fmt.Errorf("node scaling is not supported while a cluster upgrade is in progress")
+			}
 		}
 
 		// Only reject disallowed field changes; node scaling is explicitly allowed
--- a/internal/controllers/provisioningrequest_clusterinstall.go
+++ b/internal/controllers/provisioningrequest_clusterinstall.go
@@ -97,6 +97,13 @@ func (t *provisioningRequestReconcilerTask) buildClusterInstance(
 		}
 	}
 
+	// Validate that node scaling only affects worker nodes
+	if ciExists && ctlrutils.IsClusterProvisionCompleted(t.object) {
+		if err := validateScaleWorkerOnly(existingCIUnstructured, renderedCIUnstructured); err != nil {
+			return nil, err
+		}
+	}
+
 	// Validate the rendered ClusterInstance with dry-run. The defaults defined in the
 	// ClusterInstance CRD will be applied by the APIserver after the dry-run.
 	// NOTE: ClusterInstance immutable field validation is handled by ACM 2.13+
@@ -577,6 +584,87 @@ func (t *provisioningRequestReconcilerTask) handleClusterInstanceUpgrade(
 	}
 
 	return nil
+}
+
+// validateScaleWorkerOnly ensures that node scaling only adds worker nodes.
+// It compares the existing and rendered ClusterInstance node lists and rejects
+// any operation that removes nodes (scale-in is not yet supported), adds
+// non-worker nodes, or scales a single-node cluster.
+func validateScaleWorkerOnly(existingCI, renderedCI *unstructured.Unstructured) error {
+	existingNodes := getNodeRolesByHostname(existingCI)
+	renderedNodes := getNodeRolesByHostname(renderedCI)
+
+	// No scaling detected
+	if len(existingNodes) == len(renderedNodes) {
+		same := true
+		for hostname := range existingNodes {
+			if _, exists := renderedNodes[hostname]; !exists {
+				same = false
+				break
+			}
+		}
+		if same {
+			return nil
+		}
+	}
+
+	// Reject scaling on single-node clusters
+	masterCount := 0
+	for _, role := range existingNodes {
+		if role == "master" || role == "control-plane" {
+			masterCount++
+		}
+	}
+	if len(existingNodes) == 1 || (len(existingNodes) == masterCount && masterCount <= 1) {
+		return typederrors.NewInputError("node scaling is not supported on single-node clusters")
+	}
+
+	// Reject any node removals — scale-in is not yet supported
+	for hostname, role := range existingNodes {
+		if _, exists := renderedNodes[hostname]; !exists {
+			return typederrors.NewInputError(
+				"node removal is not yet supported: cannot remove %s node %q",
+				role, hostname)
+		}
+	}
+
+	// Check for added nodes with non-worker role
+	for hostname, role := range renderedNodes {
+		if _, exists := existingNodes[hostname]; !exists {
+			if role != "worker" {
+				return typederrors.NewInputError(
+					"node scaling is restricted to worker nodes: cannot add %s node %q",
+					role, hostname)
+			}
+		}
+	}
+
+	return nil
+}
+
+// getNodeRolesByHostname extracts a map of hostname → role from a ClusterInstance.
+func getNodeRolesByHostname(ci *unstructured.Unstructured) map[string]string {
+	result := make(map[string]string)
+	spec, ok := ci.Object["spec"].(map[string]any)
+	if !ok {
+		return result
+	}
+	nodes, ok := spec["nodes"].([]any)
+	if !ok {
+		return result
+	}
+	for _, node := range nodes {
+		nodeMap, ok := node.(map[string]any)
+		if !ok {
+			continue
+		}
+		hostname, _ := nodeMap["hostName"].(string)
+		role, _ := nodeMap["role"].(string)
+		if hostname != "" && role != "" {
+			result[hostname] = role
+		}
+	}
+	return result
 }
 
 // extractNodeDetails extracts necessary node details from the existing ClusterInstance
--- a/internal/controllers/provisioningrequest_clusterinstall.go
+++ b/internal/controllers/provisioningrequest_clusterinstall.go
@@ -8,10 +8,16 @@ package controllers
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"log/slog"
 	"slices"
+	"strings"
 
+	certificatesv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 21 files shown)


internal/hardwaremanager/controller/nodeallocationrequest_controller.go
```diff
@@ -302,6 +302,34 @@ func (r *NodeAllocationRequestReconciler) handleNewNodeAllocationRequestCreate(
...
+		hwmgmtv1alpha1.Provisioned, hwmgmtv1alpha1.InProgress,
+		metav1.ConditionFalse, "Allocating additional nodes for scale-out"); err != nil {
+		return hwmgrutils.RequeueWithShortInterval(), false,
+			fmt.Errorf("failed to update status for scale-out: %w", err)
+	}
+	if err := hwmgrutils.UpdateNodeAllocationRequestObservedGeneration(ctx, r.Client, nodeAllocationRequest); err != nil {
+		return hwmgrutils.RequeueWithShortInterval(), false,
+			fmt.Errorf("failed to update ObservedGeneration for scale-out: %w", err)
+	}
+	return hwmgrutils.RequeueImmediately(), true, nil
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch. Removed the `UpdateNodeAllocationRequestObservedGeneration` call from `handleScaleOut`. Setting `Provisioned=InProgress` routes the FSM directly to Processing for allocation. After allocation completes and Provisioned returns to True, the FSM re-enters SpecChanged (since `ObservedGeneration != Generation`) to process any other spec changes in the same generation, such as HwProfile updates.

Added a test assertion verifying ObservedGeneration is not advanced.

Fixed in 8f571566.

test/e2e/mno_scale_test.go
```diff
@@ -0,0 +1,484 @@
...
+				Name:      "clustertemplate-sample.v1.0.0-extramanifests",
+				Namespace: ctNamespace,
+			},
+			Data: map[string]string{},
+		}
+		for _, r := range []client.Object{pullSecret, extraManifests, clusterImageSet} {
+			if err := K8SClient.Create(testCtx, r); err != nil && !errors.IsAlreadyExists(err) {
+				Expect(err).ToNot(HaveOccurred())
+			}
+		}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Valid point. Added a `createdClusterImageSet` flag that tracks whether this test successfully created the ClusterImageSet. The AfterAll now only deletes it if the flag is set, avoiding removal of a pre-existing shared resource.

Fixed in 8f571566.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -579,6 +671,286 @@ func (t *provisioningRequestReconcilerTask) handleClusterInstanceUpgrade(
...
+		if err := spokeClient.SubResource("approval").Update(ctx, csr); err != nil {
+			t.logger.WarnContext(ctx, "Failed to approve CSR",
+				slog.String("csr", csr.Name), slog.String("hostname", hostname),
+				slog.Any("error", err))
+			continue
+		}
+		t.logger.InfoContext(ctx, "Approved scale-out CSR",
+			slog.String("csr", csr.Name), slog.String("hostname", hostname),
+			slog.String("signerName", csr.Spec.SignerName))
+	}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Acknowledged. This is a temporary workaround for ACM-38075 (Assisted Service not approving day-2 CSRs for scale-out workers with pre-existing BMHs). The CN is validated against the rendered ClusterInstance hostnames, limiting approval to nodes we explicitly provisioned. Full SAN validation for kubelet-serving CSRs would require querying the spoke node's actual addresses, which adds complexity beyond what's appropriate for a workaround. The code is clearly marked for removal once the Assisted Service bug is fixed.

internal/controllers/provisioningrequest_controller.go
```diff
@@ -721,7 +730,7 @@ func (t *provisioningRequestReconcilerTask) checkClusterDeployConfigState(ctx co
 		return requeueWithLongInterval(), nil
 	}
 
-	err = t.finalizeProvisioningIfComplete(ctx)
+	err = t.finalizeProvisioningIfComplete(ctx, nil)
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch. Fixed — when `renderedClusterInstance` is nil (monitoring path via `checkClusterDeployConfigState`), the fallback now fetches the existing ClusterInstance via `getExistingClusterInstance` and converts it to extract the node count. This ensures `FulfilledNodeCount` is recorded even when the PR first reaches Fulfilled through the monitoring path.

Fixed in 6cc32a33.

internal/controllers/provisioningrequest_clusterinstall.go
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Updated all workaround comments to explicitly reference ACM-38075 as the removal condition.

Fixed in b23b3c87.

donpenney:
Branch updated for rebase

internal/controllers/provisioningrequest_clusterinstall_test.go
```diff
@@ -744,3 +744,172 @@ func TestAssignNodeDetails(t *testing.T) {
...
+		existingCI := makeCI([]map[string]any{
+			{"hostName": "master1", "role": "master"},
+			{"hostName": "worker1", "role": "worker"},
+		})
+		renderedCI := makeCI([]map[string]any{
+			{"hostName": "master1", "role": "master"},
+			{"hostName": "worker1", "role": "worker"},
+		})
+		Expect(validateScaleWorkerOnly(existingCI, renderedCI)).To(Succeed())
+	})
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Good catch. Added a defensive role-change check in the equal-count/same-hostname path. It now compares roles for each existing hostname and rejects with "in-place role change is not supported" if any differ. Two tests added covering worker→master and master→worker changes.

Note: this is also blocked upstream by `FindClusterInstanceImmutableFieldUpdates` which treats `role` as an immutable field, but the belt-and-suspenders check ensures the invariant is enforced even if the upstream check is bypassed.

Fixed in c1974287.

donpenney:
PR updated after rebase

api/provisioning/v1alpha1/provisioningrequest_webhook.go
```diff
@@ -226,13 +226,22 @@ func (v *provisioningRequestValidator) validateCreateOrUpdate(ctx context.Contex
...
 			[][]string{}, AllowedClusterInstanceFields)
 		if err != nil {
 			return fmt.Errorf("failed to find immutable field updates for ClusterInstance (%s): %w", newPr.Name, err)
 		}
 
+		if len(scalingNodes) > 0 {
+			// Reject scaling while an upgrade is in progress
+			upgradeCond := meta.FindStatusCondition(
+				newPr.Status.Conditions, string(PRconditionTypes.UpgradeCompleted))
+			if upgradeCond != nil && upgradeCond.Reason == string(CRconditionReasons.InProgress) {
```
Missxiaoguo: Maybe it's more reliable to check the upgrade condition's status is not True when the condition is present? For the MNO upgrade, the reason can also be pending or unknown, indicating that the upgrade intent has been initiated. And when the conditon is Failed, Timedout or PrecondionChecksFailed, the cluster might not be in a health/ready state. Maybe it's safer to prevent scaling while the cluster is in these uncertain states?

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -579,6 +671,293 @@ func (t *provisioningRequestReconcilerTask) handleClusterInstanceUpgrade(
...
+	if !strings.HasPrefix(cn, nodePrefix) {
+		return "", fmt.Errorf("csr %s CN %q does not have expected prefix %q", csr.Name, cn, nodePrefix)
+	}
+	return cn[len(nodePrefix):], nil
+}
+
+// checkAllNodesJoined verifies that all hostnames in the rendered ClusterInstance
+// exist as nodes on the spoke cluster. Returns true only if every rendered
+// hostname has a corresponding node.
+func (t *provisioningRequestReconcilerTask) checkAllNodesJoined(
```
Missxiaoguo, line 885: Is it intentional to check only for Node existence? Should the Node reaching the Ready state be considered the completion of the scale-out?

test/e2e/mno_scale_test.go
```diff
@@ -0,0 +1,494 @@
...
+		}, narObj); err == nil {
+			narObj.Finalizers = nil
+			_ = K8SClient.Update(testCtx, narObj)
+			_ = K8SClient.Delete(testCtx, narObj)
+		}
+
+		// Delete AllocatedNodes
+		anList := testNonCachingListAllocatedNodesForNAR(testCtx, prName)
+		for i := range anList.Items {
+			_ = K8SClient.Delete(testCtx, &anList.Items[i])
```
Missxiaoguo, line 305: nit: I recently updated all the e2e tests to remove the finalizer from the allocated node as well. Otherwise, they could be left behind on the cluster, which might potentionaly affect other tests that use the same PR name and BMHs.
```suggestion
			anList.Items[i].Finalizers = nil
			_ = K8SClient.Update(testCtx, &anList.Items[i])
			_ = K8SClient.Delete(testCtx, &anList.Items[i])
```

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -207,13 +207,26 @@ func (t *provisioningRequestReconcilerTask) updateClusterInstance(ctx context.Co
...
-		ctlrutils.SetStatusCondition(&t.object.Status.Conditions,
-			provisioningv1alpha1.PRconditionTypes.HardwareNodeConfigApplied,
-			provisioningv1alpha1.CRconditionReasons.NotApplied,
-			metav1.ConditionFalse,
-			msg)
-		ctlrutils.SetProvisioningStateFailed(t.object, msg)
+		if ctlrutils.IsClusterProvisionCompleted(t.object) {
+			// During scale-out, the new AllocatedNode may not exist yet when the
+			// first reconcile runs. Treat this as in-progress rather than failed
+			// so the user sees "progressing" while the HW manager allocates the node.
```
Missxiaoguo, line 213: It looks like this happens because after patching the NAR, we immediately read it back and still see the Provisioned=True condition from the previous provisioning. We previously added a guard in `checkNodeAllocationRequestStatus` to avoid acting on a stale Configured condition. I wonder if we could apply a similar guard there to avoid reading this stale Provisioned condition as well.

Missxiaoguo (commented):
Looks good to me! Just a few comments/questions.

api/provisioning/v1alpha1/provisioningrequest_webhook.go
donpenney: [Comment authored by Claude]

Good point. Updated to check `Status != True` instead of `Reason == InProgress`. This now also blocks scaling during Pending, Unknown, Failed, TimedOut, and PreconditionChecksFailed states.

Fixed in fe05db76.

internal/controllers/provisioningrequest_clusterinstall.go
donpenney: [Comment authored by Claude]

Good catch. Updated `checkAllNodesJoined` to verify nodes are Ready (not just present). Added an `isNodeReady` helper that checks for `NodeReady=True`.

Fixed in fe05db76.

test/e2e/mno_scale_test.go
donpenney: [Comment authored by Claude]

Applied — AllocatedNodes now have finalizers stripped before deletion in AfterAll.

Fixed in fe05db76.

internal/controllers/provisioningrequest_hwprovision.go
donpenney: [Comment authored by Claude]

Good suggestion. Added a guard matching the pattern used for the Configured condition: when `Generation != ObservedGeneration`, the Provisioned=True condition is stale and the check returns `false` (not provisioned) so the caller requeues. This prevents the transient "failed to find matches" error during scale-out.

Fixed in fe05db76.

api/provisioning/v1alpha1/provisioningrequest_webhook.go
```diff
@@ -226,13 +227,25 @@ func (v *provisioningRequestValidator) validateCreateOrUpdate(ctx context.Contex
...
 
+		if len(scalingNodes) > 0 {
+			// Reject scaling while an upgrade is active or in a non-healthy state.
+			// Check Status != True (rather than Reason == InProgress) to also block
+			// scaling during Pending, Unknown, Failed, TimedOut, and
+			// PreconditionChecksFailed states when the cluster may not be healthy.
+			upgradeCond := meta.FindStatusCondition(
+				newPr.Status.Conditions, string(PRconditionTypes.UpgradeCompleted))
+			if upgradeCond != nil && upgradeCond.Status != metav1.ConditionTrue {
+				return fmt.Errorf("node scaling is not supported while a cluster upgrade is in progress or incomplete")
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

The current logic is intentional. The `UpgradeCompleted` condition only exists after an upgrade has been initiated. On a freshly provisioned cluster that has never been upgraded, there is no `UpgradeCompleted` condition at all. Requiring it to be present and True would block scaling on any cluster that hasn't been upgraded, which is the common case.

The behavior is:
- No `UpgradeCompleted` condition → no upgrade has ever been initiated → scaling allowed
- `UpgradeCompleted` present with `Status == True` → upgrade completed → scaling allowed
- `UpgradeCompleted` present with `Status != True` → upgrade active/failed → scaling blocked

internal/controllers/provisioningrequest_hwprovision.go
donpenney: [Comment authored by Claude]

We added a stale Provisioned guard in fe05db76 but it broke the day-2 hwProfile e2e test — the guard couldn't distinguish scale-out (Size increase) from hwProfile changes since both cause `Generation != ObservedGeneration` with `Provisioned=True`. We tried narrowing it to only fire when `Provisioned.Status == True` (a5e84230) but that still blocked day-2 because Provisioned IS validly True during hwProfile changes.

Removed the guard entirely in 304600ef. The transient scale-out race (stale `Provisioned=True` before `handleScaleOut` runs) is already handled gracefully: `updateClusterInstance` fails with "node not found" which the `IsClusterProvisionCompleted` check catches and shows as "Waiting for hardware allocation" rather than a failure.

api/provisioning/v1alpha1/provisioningrequest_webhook.go
```diff
@@ -226,13 +227,25 @@ func (v *provisioningRequestValidator) validateCreateOrUpdate(ctx context.Contex
...
+			}
+		}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Role validation and scale-in rejection are handled in the controller's `validateScaleWorkerOnly` function, not the webhook. The webhook doesn't have role information — roles come from the CI defaults merge which happens during controller reconciliation, not at admission time. The webhook only checks for scale/upgrade mutual exclusion.

`validateScaleWorkerOnly` rejects:
- Adding non-worker nodes
- Removing non-worker nodes
- Scaling on SNO clusters
- In-place role changes

Node removals (scale-in) are allowed for workers in the Phase 2 code.

internal/controllers/provisioningrequest_hwprovision.go
```diff
@@ -207,13 +207,26 @@ func (t *provisioningRequestReconcilerTask) updateClusterInstance(ctx context.Co
...
+			ctlrutils.SetProvisioningStateInProgress(t.object, msg)
+		} else {
+			failMsg := "Failed to apply node configuration to the rendered ClusterInstance: " + configErr.Error()
+			ctlrutils.SetStatusCondition(&t.object.Status.Conditions,
+				provisioningv1alpha1.PRconditionTypes.HardwareNodeConfigApplied,
+				provisioningv1alpha1.CRconditionReasons.NotApplied,
+				metav1.ConditionFalse,
+				failMsg)
+			ctlrutils.SetProvisioningStateFailed(t.object, failMsg)
+		}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Acknowledged. On a completed cluster, the most likely `applyNodeConfiguration` failure during a spec change is the unmatched-node case (new AllocatedNode not yet created). The other error types (missing spec.nodes, invalid node data, boot-MAC lookup) would have failed during initial provisioning and are unlikely to surface during scale-out on an already-provisioned cluster. We'll consider adding a sentinel error if this becomes a practical issue.

internal/controllers/provisioningrequest_hwprovision.go
Missxiaoguo: yeah, it looks like there isn't a realiable guard to prevent reading from stale for this case. 

Another thought is probably the most realiable way is to have `createOrUpdateNodeAllocationRequest` return whether it actually changed the NAR spec, if so, requeue without checking the NAR status. The next reconcile will either be triggered by the NAR watch or by the requeue interval as a fallback.  That way we avoid reading stale status for any day-2 hw operation (scale-out, hw profile, future ones), without needing per-operation guards. I guess the tradeoff is one extra reconcile cycle after every NAR patch, but it should be negligible.

Nothing to do for this PR though, the current `IsClusterProvisionCompleted`catch is a reasonable mitigation.  Just a thought for us to discuss as a further improvement.

PR #3489, opened 2026-07-22
Run make fmt after generate

alegacy:
The `make generate` command has an unfortunate side-effect of leaving an empty import statement in generate files.  Running a `make fmt` afterwards 'repairs' the file.


Makefile
```diff
@@ -211,6 +211,7 @@ manifests: deps-update controller-gen ## Generate WebhookConfiguration, ClusterR
 .PHONY: generate
 generate: deps-update controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	make fmt
```
donpenney: Should be `$(MAKE)` I think?

donpenney (approved):

PR #3535, opened 2026-07-24
CNF-25887: add MNO worker node scale-in support (Phase 2)

donpenney:
Add scale-in support for removing worker nodes from a provisioned MNO cluster. When a user removes worker node entries from the ProvisioningRequest, the controller:

1. Cordons and drains the removed nodes on the spoke cluster
2. SSA-applies an intermediate ClusterInstance with pruneManifests to instruct siteconfig to delete per-node resources (InfraEnv, NMStateConfig) while the node entries are still present
3. Waits for siteconfig to process the pruning, then deletes the Agent CRs (not managed by siteconfig)
4. SSA-applies the reduced ClusterInstance (removing node entries)
5. Waits for the removed nodes to leave the spoke cluster
6. Deletes AllocatedNode CRs (triggering BMH deallocation) and cleans up AllocatedNodeHostMap

Includes RBAC for infraenvs/nmstateconfigs (get/list/watch) and agents (delete), scale-in e2e test, and documentation updates.

### DIFF
--- a/docs/user-guide/cluster-configuration.md
+++ b/docs/user-guide/cluster-configuration.md
@@ -790,8 +790,26 @@ nodes) is planned for a future release.
 
 #### Removing a worker node (scale-in)
 
-> [!IMPORTANT]
-> Scale-in is not yet supported. Do not remove worker nodes from the
-> ProvisioningRequest nodes array. When scale-in support is added in a
-> future release, the controller will drain nodes before removal to
-> minimize workload disruption.
+To remove a worker node, remove its entry from the
+`clusterInstanceParameters.nodes` array in the ProvisioningRequest and
+update the `templateName` to reference a ClusterTemplate with matching
+CI defaults (one fewer worker node template).
+
+The controller automatically:
+
+1. Cordons and drains the removed node on the spoke cluster
+2. Applies an intermediate ClusterInstance with `pruneManifests` to
+   instruct siteconfig to delete per-node resources (InfraEnv,
+   NMStateConfig) while the node entry is still present
+3. Waits for siteconfig to process the pruning, then deletes the
+   Agent CR
+4. Applies the reduced ClusterInstance via Server-Side Apply
+   (removing the node entry)
+5. Waits for the removed node to leave the spoke cluster
+6. Deletes the AllocatedNode CR (triggering BMH deallocation)
+7. Cleans up the AllocatedNodeHostMap
+
+> [!NOTE]
+> Only worker nodes can be removed. Attempting to remove a control-plane
+> node is rejected. Scale-in and upgrade operations cannot run
+> concurrently.
--- a/docs/user-guide/cluster-configuration.md
+++ b/docs/user-guide/cluster-configuration.md
@@ -724,10 +724,9 @@ The following steps are required:
 
 ### Scaling Worker Nodes
 
-After a multi-node (MNO) cluster is provisioned, you can add worker
-nodes by updating the ProvisioningRequest's
-`clusterInstanceParameters.nodes` array. Scale-in (removing worker
-nodes) is planned for a future release.
+After a multi-node (MNO) cluster is provisioned, you can add or remove
+worker nodes by updating the ProvisioningRequest's
+`clusterInstanceParameters.nodes` array.
 
 > [!NOTE]
 > Scaling is only supported for worker nodes. Control plane (master)
--- a/docs/user-guide/cluster-configuration.md
+++ b/docs/user-guide/cluster-configuration.md
@@ -796,17 +796,19 @@ CI defaults (one fewer worker node template).
 
 The controller automatically:
 
-1. Cordons and drains the removed node on the spoke cluster
-2. Applies an intermediate ClusterInstance with `pruneManifests` to
+1. Signals the hardware manager to drain and decommission the
+   removed node (via a NAR annotation)
+2. The hardware manager cordons and drains the node on the spoke,
+   deletes the Node object, deletes the AllocatedNode CR
+   (triggering BMH deallocation), and cleans up node tracking
+3. Applies an intermediate ClusterInstance with `pruneManifests` to
    instruct siteconfig to delete per-node resources (InfraEnv,
    NMStateConfig) while the node entry is still present
-3. Waits for siteconfig to process the pruning, then deletes the
+4. Waits for siteconfig to process the pruning, then deletes the
    Agent CR
-4. Applies the reduced ClusterInstance via Server-Side Apply
+5. Applies the reduced ClusterInstance via Server-Side Apply
    (removing the node entry)
-5. Waits for the removed node to leave the spoke cluster
-6. Deletes the AllocatedNode CR (triggering BMH deallocation)
-7. Cleans up the AllocatedNodeHostMap
+6. Cleans up the AllocatedNodeHostMap
 
 > [!NOTE]
 > Only worker nodes can be removed. Attempting to remove a control-plane
--- a/internal/controllers/provisioningrequest_clusterinstall.go
+++ b/internal/controllers/provisioningrequest_clusterinstall.go
@@ -22,6 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -29,8 +31,11 @@ import (
 
 	hwmgmtv1alpha1 "github.com/openshift-kni/oran-o2ims/api/hardwaremanagement/v1alpha1"
 	provisioningv1alpha1 "github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1"
+	"github.com/openshift-kni/oran-o2ims/internal/constants"
 	ctlrutils "github.com/openshift-kni/oran-o2ims/internal/controllers/utils"
 	"github.com/openshift-kni/oran-o2ims/internal/controllers/utils/spokeclient"
+	hwmgrcontroller "github.com/openshift-kni/oran-o2ims/internal/hardwaremanager/controller"
+	k8sclients "github.com/openshift-kni/oran-o2ims/internal/service/common/clients/k8s"
 	typederrors "github.com/openshift-kni/oran-o2ims/internal/typed-errors"
 	siteconfig "github.com/stolostron/siteconfig/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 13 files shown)


docs/user-guide/cluster-configuration.md
```diff
@@ -790,8 +790,26 @@ nodes) is planned for a future release.
...
 
-> [!IMPORTANT]
-> Scale-in is not yet supported. Do not remove worker nodes from the
-> ProvisioningRequest nodes array. When scale-in support is added in a
-> future release, the controller will drain nodes before removal to
-> minimize workload disruption.
+To remove a worker node, remove its entry from the
+`clusterInstanceParameters.nodes` array in the ProvisioningRequest and
+update the `templateName` to reference a ClusterTemplate with matching
+CI defaults (one fewer worker node template).
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed in `d6fa0b91` — updated the intro to reflect that scale-in is now supported.

test/e2e/mno_scale_test.go
```diff
@@ -449,6 +449,44 @@ var _ = Describe("MNO Scale-Out test", Ordered, Label("mno-scale-out"), func() {
...
+				sizes := nodeGroupSizeMap(nar)
+				return sizes[worker] == workerCount && sizes[master] == masterCount
+			}, timeout, interval).Should(BeTrue(),
+				"Worker Size should be %d and master Size should remain %d", workerCount, masterCount)
+		})
+
+		// NOTE: AllocatedNode deletion is gated on spoke node removal
+		// (checkRemovedNodesGone), which requires a real spoke cluster.
+		// This cannot be verified in the e2e test environment. Lab testing
+		// validates the full flow including AllocatedNode cleanup.
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

This is acknowledged in the test comment at lines 485-488: AllocatedNode deletion is gated on `checkRemovedNodesGone`, which requires a real spoke cluster that the e2e test environment doesn't have. The full cleanup flow (AllocatedNode deletion, host-map cleanup, BMH deallocation) has been verified through lab testing on a real MNO cluster.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -936,6 +1393,98 @@ func isNodeReady(node *corev1.Node) bool {
...
+	narNS := ctlrutils.GetEnvOrDefault(constants.DefaultNamespaceEnvName, constants.DefaultNamespace)
+	for _, nodeID := range nodesToDelete {
+		an := &hwmgmtv1alpha1.AllocatedNode{}
+		an.Name = nodeID
+		an.Namespace = narNS
+		if err := t.client.Delete(ctx, an); err != nil {
+			if !errors.IsNotFound(err) {
+				return fmt.Errorf("failed to delete AllocatedNode %s: %w", nodeID, err)
+			}
+		}
```
Missxiaoguo: Would this be a separation of concerns issue? Our current design principle is that the NAR controller manages AllocatedNodes (create, update, deletion) and the PR controller sends requests through the NAR. Also, it looks like the NAR controller doesn't handle reduced group size — would the removed node's entry still remain in `status.properties.nodeNames` after completion? Might be worth considering delegating the BMH/AllocatedNode deprovisioning (probably drain as well) to the hardware manager 🤔 .

donpenney: [Comment authored by Claude]

You're right — this breaks separation of concerns. The NAR controller should manage AllocatedNode lifecycle, not the PR controller.

The plan is to delegate both drain and deallocation to the hardware manager via NAR annotations:

1. PR controller annotates the NAR with nodes to drain
2. NAR controller handles the drain (it already has NodeOps/spoke client capability from the hardware configuration workflow)
3. Drain status is reported via a condition on the AllocatedNode (e.g., `Drained` with reasons `Cordoned`, `DrainComplete`), following the existing per-node condition pattern
4. Once drain completes, PR controller annotates the NAR with nodes to deallocate
5. NAR controller deletes the AllocatedNodes, cleans up `status.properties.nodeNames`, and removes the annotation

The drain status condition should also be added to the existing hardware configuration (update) workflow for consistency — it currently drains/uncordons without any status visibility on the AllocatedNode.

This also addresses the stale `nodeNames` issue and keeps all node lifecycle operations in the hardware manager.

Missxiaoguo: Step4 might not be needed. Allocatednode has status.hostname which is the k8s node name, the NAR controller should be able to identify which nodes to deallocated by comparing the hostnames in the drain annotation (from step1) against status.hostname. But need to ensure status.hostname is populated before scale-in run - currently it’s only set during day2 hwprofile configuration.
I am also wondering if we want to introduce a new condition (e.g. Deprovisioned) to track the status or probably reuse the existing Configured condition with different reason/message to indicate deprovisioning operation? The latter might be simper since the PR controller already watches Configured condition status, so it would work without PR controller changes, but it feels sematically awkward …

donpenney: [Comment authored by Claude]

Good suggestions — we incorporated both in the latest push:

- Combined drain and deallocation into a single `clcm.openshift.io/scale-in-nodes` annotation. The NAR controller handles the full lifecycle: drain on spoke → set `Deprovisioned` condition → delete Node from spoke → delete AllocatedNode → clean up `status.properties.nodeNames` → remove annotation. The PR controller sets the annotation and waits for it to be removed.
- Added a `Deprovisioned` condition on AllocatedNode to track the scale-in lifecycle (went with a new condition rather than overloading `Configured`)
- `Status.Hostname` is now populated at provisioning completion and during spec changes, not just during day-2 config. The scale-in handler also calls `populateNodeHostnames` as a safety net before attempting drain.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -936,6 +1400,94 @@ func isNodeReady(node *corev1.Node) bool {
...
+	// The scale-in-nodes annotation was set by handleScaleInDrain. The NAR
+	// controller handles drain, Node deletion from spoke, AllocatedNode
+	// deletion, and nodeNames cleanup. Wait for it to finish.
+	nar, err := t.getNAR(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get NAR for scale-in cleanup: %w", err)
+	}
+	if _, ok := nar.GetAnnotations()[hwmgrcontroller.ScaleInNodesAnnotation]; ok {
+		return fmt.Errorf("waiting for NAR controller to complete scale-in")
+	}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — `handleScaleInCleanup` now returns `(done bool, err error)` instead of using an error for the wait state. The caller requeues cleanly when `done` is false.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -936,6 +1400,94 @@ func isNodeReady(node *corev1.Node) bool {
...
+	renderedHostnames := getNodeRolesByHostname(renderedCI)
+	for hostname := range spokeNodes {
+		if _, inRendered := renderedHostnames[hostname]; !inRendered {
+			t.logger.InfoContext(ctx, "Removed node still present on spoke",
+				slog.String("hostname", hostname))
+			return false, nil
+		}
+	}
+
+	return true, nil
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — `checkRemovedNodesGone` now only checks hostnames from `AllocatedNodeHostMap` that are absent from the rendered CI, rather than iterating all spoke nodes.

internal/hardwaremanager/controller/nodeallocationrequest_controller_test.go
```diff
@@ -478,3 +481,163 @@ var _ = Describe("handleScaleOut", func() {
...
+		Expect(updatedNAR.GetAnnotations()).ToNot(HaveKey(ScaleInNodesAnnotation))
+
+		// Verify nodeNames was pruned
+		Expect(updatedNAR.Status.Properties.NodeNames).To(ConsistOf("node-2"))
+
+		// Verify AllocatedNode was deleted
+		deletedNode := &hwmgmtv1alpha1.AllocatedNode{}
+		err = fakeClient.Get(ctx, client.ObjectKeyFromObject(node), deletedNode)
+		Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+	})
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed — `Status.Hostname` is now persisted via `Status().Update` after `Create`, matching how the fake client handles status subresources.

internal/hardwaremanager/controller/nodeallocationrequest_controller.go
```diff
@@ -269,6 +294,203 @@ func (r *NodeAllocationRequestReconciler) HandleNodeAllocationRequest(
...
+			if !errors.IsNotFound(err) {
+				return hwmgrutils.RequeueWithShortInterval(),
+					fmt.Errorf("failed to delete AllocatedNode %s: %w", nodeName, err)
+			}
+		}
+		r.Logger.InfoContext(ctx, "Deleted AllocatedNode",
+			slog.String("allocatedNode", nodeName))
+
+		remainingNodeNames = removeFromSlice(remainingNodeNames, nodeName)
+	}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Addressed — BMH identities are now resolved before `requestAndWaitForDrain` (while AllocatedNodes still exist) and passed through to the pruneManifests and Agent cleanup steps. This ensures the lookups succeed even after the NAR controller deletes the AllocatedNodes.

internal/hardwaremanager/controller/nodeallocationrequest_controller.go
```diff
@@ -517,6 +744,21 @@ func (r *NodeAllocationRequestReconciler) handleNodeAllocationRequestProcessing(
...
+		// for a day-2 config trigger.
+		nodelist, err := hwmgrutils.GetChildNodes(ctx, r.Logger, r.Client, nodeAllocationRequest)
+		if err != nil {
+			return hwmgrutils.RequeueWithShortInterval(),
+				fmt.Errorf("failed to list AllocatedNodes for hostname population: %w", err)
+		}
+		if err := populateNodeHostnames(ctx, r.Client, r.Logger, nodelist, nodeAllocationRequest); err != nil {
+			r.Logger.WarnContext(ctx, "Failed to populate node hostnames, will retry on next reconcile",
+				slog.Any("error", err))
+		}
```
Missxiaoguo: This doesn't seem to be useful. At this point, the PR's `AllocatedNodeHostMap` doesn't seem to be populated yet.

internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -204,6 +209,399 @@ func (t *provisioningRequestReconcilerTask) buildClusterInstanceUnstructured() (
...
+	ctlrutils.SetProvisioningStateInProgress(t.object, message)
+	if updateErr := ctlrutils.UpdateK8sCRStatus(ctx, t.client, t.object); updateErr != nil {
+		return false, fmt.Errorf("failed to update scale-in status: %w", updateErr)
+	}
+
+	// Resolve BMH identities before requesting drain. The NAR controller
+	// deletes AllocatedNodes during scale-in processing, so lookups must
+	// happen while the CRs still exist.
+	removedNodeIDs := t.hostnameToAllocatedNodeIDs(removedHostnames)
+	removedNodeBMHs, err := t.lookupRemovedNodeBMHs(ctx, removedHostnames)
```
Missxiaoguo, line 264: not sure if there is a potential race condition here where the allocated nodes may have already gone by the time of subsequent reconciles?

docs/user-guide/cluster-configuration.md
```diff
@@ -790,8 +789,26 @@ nodes) is planned for a future release.
...
+1. Cordons and drains the removed node on the spoke cluster
+2. Applies an intermediate ClusterInstance with `pruneManifests` to
+   instruct siteconfig to delete per-node resources (InfraEnv,
+   NMStateConfig) while the node entry is still present
+3. Waits for siteconfig to process the pruning, then deletes the
+   Agent CR
+4. Applies the reduced ClusterInstance via Server-Side Apply
+   (removing the node entry)
+5. Waits for the removed node to leave the spoke cluster
+6. Deletes the AllocatedNode CR (triggering BMH deallocation)
```
Missxiaoguo: outdated order for step6?

internal/hardwaremanager/controller/nodeallocationrequest_controller.go
donpenney: My goal here was to have the `status.Hostname` populated on the `AllocatedNode` as early as possible - if nothing else, at least for user info. I'll review this again

donpenney: [Comment authored by Claude]

You're right — `AllocatedNodeHostMap` isn't populated yet at provisioning completion. Removed this call. The SpecChanged handler and scale-in handler calls are the ones that work (they run after the PR controller sets the hostmap).

internal/controllers/provisioningrequest_clusterinstall.go
donpenney: [Comment authored by Claude]

Good catch. The lookup is now non-fatal (warns instead of erroring) since AllocatedNodes may already be deleted on subsequent reconciles. The BMH identities resolved on the first detection are used for pruneManifests checking and Agent CR cleanup — if they can't be resolved (AllocatedNodes gone), those steps degrade gracefully (empty list → no-op).

docs/user-guide/cluster-configuration.md
donpenney: [Comment authored by Claude]

Fixed — updated the docs to reflect the NAR controller handling drain, Node deletion, and AllocatedNode deallocation.

PR #3596, opened 2026-07-28
CNF-25888: add scale-in abort, skip drain for NotReady nodes, allow node group removal (Phase 3)

donpenney:
Harden the scale-in flow for production edge cases:

- When a user re-adds removed nodes during a scale-in, the controller detects the abort, removes the scale-in-nodes annotation from the NAR, and uncordons the specific nodes that were being drained (not all cordoned nodes). Abort errors are propagated for retry.

- Skip drain for nodes that are NotReady or absent from the spoke cluster. A dead/unreachable node cannot be drained — attempting to do so would retry indefinitely until the overall provisioning timeout. The NAR controller now checks IsNodeReady before draining and proceeds directly to deallocation if the node is down.

- Remove validateNodeGroupsMatchNAR, which required the NAR's node groups to exactly match the hwMgmt data. This blocked scaling to zero workers because the 3+0 template omits the worker group from hwMgmt defaults, while the existing NAR still has one. Node groups are naturally added/removed as part of scaling operations and do not require strict validation.

- Document the abort procedure and its limitations (abort window closes once deallocation starts).


internal/controllers/provisioningrequest_clusterinstall.go
```diff
@@ -307,6 +313,93 @@ func (t *provisioningRequestReconcilerTask) handleScaleInDrain(
...
+			}
+			return fmt.Errorf("failed to get spoke node %s: %w", hostname, err)
+		}
+		if spokeNode.Spec.Unschedulable {
+			t.logger.InfoContext(ctx, "Uncordoning node from aborted scale-in",
+				slog.String("hostname", hostname))
+			if err := nodeOps.UncordonNode(ctx, hostname); err != nil {
+				return fmt.Errorf("failed to uncordon node %s: %w", hostname, err)
+			}
+		}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed in d534a21. Reordered to uncordon nodes before removing the annotation, so failures leave it in place for retry on next reconcile.

internal/hardwaremanager/controller/nodeallocationrequest_controller.go
```diff
@@ -460,6 +448,47 @@ func (r *NodeAllocationRequestReconciler) handleScaleInNodesAnnotation(
...
+		r.Logger.InfoContext(ctx, "Node not found on spoke, skipping drain",
+			slog.String("hostname", an.Status.Hostname),
+			slog.Any("error", readyErr))
+		return nil
+	}
+	if !ready {
+		r.Logger.InfoContext(ctx, "Node is NotReady on spoke, skipping drain",
+			slog.String("hostname", an.Status.Hostname))
+		return nil
+	}
```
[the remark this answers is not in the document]
donpenney: [Comment authored by Claude]

Fixed in d534a21. Now only IsNotFound skips drain; other errors are propagated for retry.

internal/controllers/provisioningrequest_hwprovision_test.go
```diff
@@ -1516,15 +1515,9 @@ var _ = Describe("checkExistingNodeAllocationRequest", func() {
...
 
-		hwMgmtData := map[string]any{
-			"nodeGroupData": []any{
-				map[string]any{"name": "controller", "role": "master", "hwProfile": "profile-spr-single-processor-64G"},
-				map[string]any{"name": "worker", "role": "worker", "hwProfile": "profile-spr-dual-processor-128G"},
-			},
-		}
 		nodeAllocationRequestId := crName
 
-		response, err := task.checkExistingNodeAllocationRequest(ctx, hwMgmtData, nodeAllocationRequestId)
```
Missxiaoguo, line 1527: Since `validateNodeGroupsMatchNAR` has been removed from checkExistingNodeAllocationRequest, this test case doesn't seem very useful anymore. I wonder if we should just remove `checkExistingNodeAllocationRequest` function, since it's now just a thin wrapper around `getNAR`.

Missxiaoguo (commented):
Looks good to me! Just one suggestion

internal/controllers/provisioningrequest_hwprovision_test.go
donpenney: [Comment authored by Claude]

Good catch. `checkExistingNodeAllocationRequest` is indeed just a thin wrapper around `getNAR` now — the only difference is it swallows `IsNotFound`. The single caller at line 882 discards the returned NAR anyway. Will inline the logic and remove the function.

donpenney: Fixed

donpenney:
Also rebased to pick up konflux updates

PR #3620, opened 2026-07-29
Fix hardware-manager TLS and metrics

rauhersu:
**Why these changes ?** 

These changes are a follow-up to https://github.com/openshift-kni/oran-o2ims/pull/2693 : the implementation for the hardware-manager TLS was partially done there. That development included the generic env var passing mechanism for TLS enforcement in the hardware-manager but those variables were never applied and the TLS configuration was never enforced on that pod.

Unfortunately, neither Claude nor CodeRabbit detected those variables were never applied.

Similarly, the metrics development was partially done: the hardware-manager configured metrics but they were never correctly exposed.

**Backport ?** 

Yes, we will need to fix this present already in 4.22 too.

### DIFF
--- a/bundle/manifests/oran-o2ims-hardwaremanager-server-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/oran-o2ims-hardwaremanager-server-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -1,0 +1,21 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/component: metrics
+    app.kubernetes.io/created-by: oran-o2ims
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/part-of: oran-o2ims
+  name: oran-o2ims-hardwaremanager-server-metrics-monitor
+spec:
+  endpoints:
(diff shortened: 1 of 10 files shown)


rauhersu:
Need to assess coderabbit comments

/hold

rauhersu:
Changes ready for review. Ptal, @browsell @alegacy @donpenney

donpenney:
[Comment authored by Claude]

Looks good overall. The changes are well-scoped — fixing two gaps from the original TLS/metrics implementation:

1. **TLS enforcement on hardware manager** — `NewTLSProfileFromEnv()` + `NewTLSConfiguratorFromProfile()` applied to the controller-runtime metrics server. Clean and follows the same pattern as the API servers.

2. **Metrics ServiceMonitor** — adding the hardware-manager ServiceMonitor with the correct `app: hardwaremanager-server` selector and TLS config.

3. **Exporting `NewTLSProfileFromEnv`** — necessary for the hardware manager to call it from a different package. The rename is clean and all call sites are updated.

**One minor observation:** The removal of the `oran-o2ims/metrics` label guard (lines 1134-1136 in `inventory_controller.go`) means the database service will now also get this label. The label isn't used by any selector so it's harmless, but the PR description doesn't mention why it was removed — might be worth a note if anyone asks.

Docs updates for both `prometheus-metrics.md` and `tls-configuration.md` are thorough.

No blocking issues.

PR #3630, opened 2026-07-30
CNF-26210: add MNO cluster node swap support (Phase 4)

donpenney:
Enable simultaneous worker node removal and addition in a single ProvisioningRequest update. The controller sequences the operation: scale-in (drain, prune, deallocate) completes first, then scale-out (provision, join) fires automatically.

Key changes:
- Webhook: detect in-place hostName changes as node swaps, not immutable field violations
- Filter new nodes from intermediate CI during swap so BMC secret validation doesn't fail before hardware provisioning runs
- Replace count-based scale-in detection with hostname-based check using AllocatedNodeHostMap, catching same-count swaps
- Set FulfilledNodeCount to oldCount - numRemoved after scale-in cleanup, so scale-out naturally triggers on next reconcile
- Force NAR ConfigTransactionId bump when allocation count is short after a swap's scale-in (spec unchanged but nodes are missing)
- Skip already-cleaned-up nodes in handleScaleInDrain by checking AllocatedNodeHostMap, preventing repeated drain processing
- Keep ClusterProvisioned=InProgress between scale-in and scale-out phases of a swap (no cycled state transitions)
- Update scale-out status message when transitioning from scale-in
- Block upgrades during swap via ClusterProvisioned status check
- Include BMH in pruneManifests to clean up stale manifestsRendered entries (siteconfig skips deletion when owned-by label is absent)

### DIFF
--- a/api/provisioning/v1alpha1/provisioningrequest_validation.go
+++ b/api/provisioning/v1alpha1/provisioningrequest_validation.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
 
 	"os"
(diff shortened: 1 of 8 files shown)


donpenney:
[Comment authored by Claude]

CodeRabbit's finding about the unpersisted scale-out status message during swap transition is fixed in 354ac0d. Added `UpdateK8sCRStatus` call in the `else` branch so the message updates from the stale scale-in text to the scale-out text. This only affects the swap scenario (where `ClusterProvisioned` is already `InProgress` when scale-out fires).

donpenney:
[Comment authored by Claude]

@coderabbitai Re: https://github.com/openshift-kni/oran-o2ims/pull/3630#pullrequestreview-4814321087

This is by design — the scale-out branch only fires after scale-in cleanup has already completed. The sequencing is:

1. `handleScaleInDrain` (runs before `handleClusterInstallation` in `executeProvisioningPhases`) completes all drain/prune work
2. The hostname-based scale-in block in `handleClusterInstallation` runs cleanup and sets `FulfilledNodeCount = oldCount - numRemoved`, then returns
3. On the next reconcile, `handleScaleInDrain` finds no removed nodes (AllocatedNodeHostMap is clean), so it returns `(false, nil)`
4. The scale-out block fires because `renderedNodeCount > fulfilledCount`

At step 4, there are no pending removals — the hostname-diff check would find `removedNodesExist = false`. The scale-out branch running first is the intended behavior.

PR #3638, opened 2026-07-31
Update dependabot config after go 1.26 upgrade

donpenney:


donpenney:
/override ci/prow/e2e

We may need to revise the filters for the e2e test. It's not needed for an update like this, whereas ci-job runs yamllint

PR #3687, opened 2026-08-03
Redefine FBC catalog to use per-z-stream channels (4.22)

by2waysprojects:
## Summary

- Replace the single `4.22` channel with individual z-stream channels (`4.22.0`, `4.22.1`) so each has a single head with no `replaces` edge
- Preserve the `stable` channel pointing to the latest z-stream (`v4.22.1`) for new installations
- Prevent OLM from offering automatic in-place upgrades between z-streams, which require manual uninstall and reinstall

## Details

OLM requires each channel to have exactly one head, and the `replaces` field both defines the upgrade edge and maintains the single-head hierarchy. This made it impossible to block automatic z-stream upgrades within the same channel.

With this change:
- **`stable`**: contains only `v4.22.1` (latest z-stream) — for new installations
- **`4.22.0`**: pinned to `v4.22.0` — no automatic upgrade
- **`4.22.1`**: pinned to `v4.22.1` — no automatic upgrade

Existing users subscribed to `stable` with `v4.22.0` installed will **not** be auto-upgraded. The running operator continues to function normally. Users must manually uninstall and reinstall to move between z-streams.

Ref: [CNF-26346](https://redhat.atlassian.net/browse/CNF-26346)

## Test plan

- [x] Validate FBC template renders correctly with `opm alpha render-template basic`
- [x] Verify OPM validation passes (no "multiple channel heads" errors)
- [x] Confirm new subscription to `stable` resolves to `v4.22.1`
- [x] Confirm subscription to `4.22.0` resolves to `v4.22.0`
- [x] Confirm no upgrade path exists between z-stream channels
- [ ] Verify existing `v4.22.0` installation on `stable` is not auto-upgraded

Made with [Cursor](https://cursor.com)


rauhersu:
Just holding this in case you want to test this yourself further. Feel free to unhold.

/hold
/lgtm
/approve

PR #3771, opened 2026-08-05
OCPBUGS-99224: fix IBI cluster deletion stall caused by BMO secret finalizer

donpenney:
During IBI ProvisioningRequest deletion, siteconfig tries to delete the network-data secret (named after the BMH) in the BMH pool namespace. BMO holds a baremetalhost.metal3.io/secret finalizer on this secret because the BMH has an ownerReference to it. Since the BMH outlives the cluster (it's pre-existing hardware), the finalizer is never released and ClusterInstance deletion stalls indefinitely.

Fix by deleting the IBI network-data secret during IBI post-provisioning (enableBMOManagementForIBINodes), alongside the existing DataImage cleanup. The secret is an IBI artifact that's no longer needed once we take over BMH management. Also restore the BMH's preprovisioningNetworkDataName to its original value and clear the PPI's networkDataName.

Key changes:
- Delete IBI network-data secret at allocation time, not deallocation
- Only delete if preprovisioningNetworkDataName differs from the saved original, preventing deletion of user's pre-existing secrets
- Restore BMH preprovisioningNetworkDataName from saved annotation
- Clear PPI networkDataName since the referenced secret is deleted
- Cleanup is resumable: BMH and PPI references are restored even if the secret is already gone (e.g. on retry after partial failure)
- Add delete verb to hardware manager ClusterRole for secrets


donpenney:
[Comment authored by Claude]

All three findings addressed in b9f4012:

1. **Require annotation proof** — `deleteIBINetworkDataSecret` now returns early if `OrigNetworkDataAnnotation` is absent, preventing deletion of user secrets on upgrades from older operator versions.
2. **Unconditional delete after finalizer removal** — Finalizer removal is conditional but `c.Delete` always executes, so retries after partial failure don't orphan the secret.
3. **PPI before BMH** — Reordered cleanup: PPI `networkDataName` is cleared before BMH `preprovisioningNetworkDataName` is restored. If PPI cleanup fails, the BMH still has the IBI value and the next reconcile retries everything.

donpenney:
[Comment authored by Claude]

@coderabbitai Re: https://github.com/openshift-kni/oran-o2ims/pull/3771#pullrequestreview-4870160109

Added `isIBIGeneratedSecret` validation in b433b23. Before removing the finalizer or deleting the secret, we now verify it has ownerReferences to both the BareMetalHost and PreprovisioningImage with the expected name. If either is missing, the secret is not IBI-generated and we skip deletion entirely.

donpenney:
Updated for rebase

Missxiaoguo (commented):
/lgtm
/approve

PR #3884, opened 2026-08-11
CNF-26438: Refactor DeepMergeMaps for reusable by-key slice merging

Missxiaoguo:
Today `DeepMergeMaps` only merges slices by index, and hwMgmt nodeGroupData needs a separate `MergeNodeGroupData` helper: peel the list out of the map, merge by name, then `DeepMergeMaps` the remaining fields. That split does not scale — ClusterInstance nodeGroups (and nested lists such as interfaces) need the same by-key behavior inside a single deep merge.

Refactor to one path-driven helper so callers declare which slices merge by key; everything else keeps index-based merge. This is preparatory work for [CNF-26438](https://redhat.atlassian.net/browse/CNF-26438) (provisioning with nodeGroups) and avoids more peel-and-merge special cases. Existing provisioning merge behavior is unchanged: hwMgmt nodeGroupData still merges by name, and other fields/slices continue to merge by index.

**Changes**:
- Add `SliceMergeRule` / `ByKeyMergeOptions` / `SliceMergeRules` (path → rule)
- Extend `DeepMergeMaps` with optional `SliceMergeRules` so nested slices can merge by either index or key (nil = by-index only)
- Support nested rule paths (e.g. nodeGroups[].nodeNetwork.interfaces)
- Replace the custom `MergeNodeGroupData` by-name merge with the shared `deepMergeSlicesByKey` helper via `SliceMergeRules`
- Wire nodeGroupData via `SliceMergeRules{Key: "name"}` and `HwMgmtNodeGroupDataKey`

Also move merge types, helpers, and tests from utils.go / utils_test.go into merge.go / merge_test.go.


Missxiaoguo:
addressing coderabbit nit comment

Missxiaoguo:
@coderabbitai Addresssed your Nitpick. To be consistent with DeepMergeMaps which performs merges in place, I updated the slice merging to be in-place as well.

donpenney:
[Comment authored by Claude]

Looks good — the merge logic is correct, backward compatibility is maintained, and test coverage is thorough. No blocking issues.

One suggestion: since this is foundational merge logic that agents (and developers) will need to use correctly, consider adding guidance to `AGENTS.md` about when to use by-key vs by-index merge and how to specify rules. Something like:

```markdown
### Deep Merge Semantics

`DeepMergeMaps` in `internal/controllers/utils/merge.go` merges two
`map[string]any` trees (typically defaults + user input). Source values
override destination values. The fourth argument (`SliceMergeRules`)
controls how slices are merged:

- **By index** (default, nil rules): elements at the same index are
  deep-merged. Use for ordered lists where position matters (e.g.,
  `nodeNetwork.config.interfaces`).
- **By key** (specify a `SliceMergeRule` with `Key` field): elements
  are matched by a named field (e.g., `name`), deep-merged when
  matched, and unmatched source entries are appended. Use for
  name-keyed lists (e.g., `hwMgmtDefaults.nodeGroupData` keyed by
  `name`).

Rules are path-based: `"nodeGroupData"` matches the top-level key,
`"nodeGroups[].interfaces"` matches `interfaces` inside each
`nodeGroups` element. Paths not in the rules map use index-based merge.

`DeepMergeMaps` mutates the destination map in place. Callers that
need to preserve the original must clone before calling.

Do not create domain-specific merge helpers (like the former
`MergeNodeGroupData`). Instead, use `DeepMergeMaps` with appropriate
`SliceMergeRules`.
```

Missxiaoguo:
@donpenney updated! The important guidance has been added to instruct agents to extend the helper's options when different merge behavior is needed, rather than introducing domain-specific merge helpers. Instead of duplicating the merge behavior and available options in AGENTS.md, it points agents to the relevant code, where they can learn the details when needed.

PR #4174, opened 2026-08-28
Ibi

browsell:
## Summary

Adds two design proposals under `docs/proposals/` covering the two halves of a fully automated, API-driven Image-Based Install (IBI) workflow for O-Cloud Manager. Both are documentation only — no controller or API code changes.

- `docs/proposals/automated-seed-image-generation.md`
- `docs/proposals/ibi-server-pre-provisioning.md`

## 1. Automated Seed Image and Live ISO Generation

Turns the manual, SSH-driven LCA seed-generation procedure into a first-class operation driven by the existing `ProvisioningRequest` API.

- New `seedGeneration` key under `upgradeDefaults` / `upgradeParameters`, reusing the existing merge-and-validate infrastructure rather than adding a new CRD or controller.
- Automates seed image generation and live installation ISO build + upload, matching the correct `openshift-install` version automatically.
- Reports progress via a typed `SeedGenerationCompleted` condition and `SeedGenerationStatus` fields (no more SSH-and-tail).
- Treats the seed cluster as a disposable factory: removes ACM agent state (klusterlet + addons) so it never contaminates the seed; retains spoke access during generation via the hub-held admin kubeconfig; never auto-deletes the `ManagedCluster` or the `ProvisioningRequest`.
- Disconnected environments are the primary use case: mirror mappings and registry trust are auto-derived from the hub.
- Fails fast with `PreconditionChecksFailed` (push access, release reachability, TLS/CA trust, upload credentials) before the lengthy generation begins.
- Surfaces the ISO-server CA cert in status for downstream (IBI) consumption.

**Security posture:** credential and outbound-target fields
  (`seedAuthSecretRef`, `recertImage`, `liveISO.*`, `uploadSecretRef`, `urlBase`,
  `pullSecretRef`, `imageDigestSources`, `additionalTrustBundleConfigMapRef`) are
  template-owned. The PR-facing `templateParameterSchema` exposes only the
  overridable field(s) with `additionalProperties: false`, so any template-owned
  key in PR input is rejected structurally; a separate controller-internal
  effective-object schema validates the merged document post-merge.

  ## 2. Automated IBI Server Pre-Provisioning

  Automates booting bare-metal servers from the live ISO and confirming readiness,
  integrated into the IBI Operator (which already owns BMH lifecycle and runs an
  HTTP image server).

  - HTTP **callback** monitoring (server → hub) instead of SSH: a systemd unit
    injected via `ignitionConfigOverride` calls back to an IBI Operator endpoint
    when the preparation service finishes.
  - A pre-provisioning state machine with explicit condition reasons
    (`PreProvisioningPending` → `Booting` → `InProgress` →
    `Succeeded`/`Failed`/`TimedOut`, with `Retrying`).
  - Per-attempt bearer-token auth, a `DataImage` carrier bound to the BMH,
    wall-clock callback budget, durable retry-pending marker, and
    compare-and-swap on the terminal result to accept only the first terminal
    callback.
  - Stream-and-verify of ISO bytes against a mandatory digest before boot;
    origin allowlist.

  ## Testing

  - Docs-only change. `make markdownlint CONTAINER_TOOL=podman`: 0 errors.
  - Both proposals kept under 100 KB.

### DIFF
--- a/docs/proposals/automated-seed-image-generation.md
+++ b/docs/proposals/automated-seed-image-generation.md
@@ -778,7 +778,7 @@ The controller creates a Kubernetes Job on the hub cluster that builds the live 
    The pull secret is constructed by merging the hub cluster pull secret (`openshift-config/pull-secret`) with the `seedAuthSecretRef` credentials. If `liveISO.pullSecretRef` is provided, its contents are used directly instead of the auto-merged secret. In disconnected environments,
    the pull secret must include credentials for both the local seed image mirror and the OCP release image mirror.
 
-4. **Build the ISO**: Run `openshift-install image-based create image --dir <workdir>`. This pulls the seed image, generates ignition configuration, and produces the `rhcos-ibi.iso` file. This step requires significant ephemeral storage (~15GB for seed image pull + ISO generation).
+4. **Build the ISO**: Run `openshift-install image-based create image --dir <workdir>`. This pulls the seed image, generates ignition configuration, and produces the `rhcos-ibi.iso` file. This step requires significant workspace storage (~15-20GB for seed image pull + ISO generation).
 
 5. **Upload the ISO**: SCP the ISO to the HTTPS server using the SSH credentials from `uploadSecretRef`, pinning the server host key from `knownHosts` with strict checking:
 
@@ -1067,7 +1067,11 @@ type SeedGenerationStatus struct {
     // ISOServerCACertRef is a typed reference to a ConfigMap the controller
     // materializes on completion (in the O-Cloud Manager namespace, named
     // deterministically from the ProvisioningRequest) holding the PEM CA bundle
-    // for the HTTPS ISO server under a well-known key. Populated only when
+    // for the HTTPS ISO server under the well-known key "ca-bundle.crt".
+    // Consumers that resolve the bundle in a different namespace (e.g. the IBI
+    // pre-provisioning flow, which reads ImageClusterInstall's name-only
+    // ISOServerCACertRef under "ca-bundle.crt" in the ICI namespace) copy this
+    // ConfigMap into the target namespace under the same key. Populated only when
     // uploadSecretRef.caCert is present. The controller creates the ConfigMap
     // rather than storing raw PEM here, so downstream consumers (e.g. BMC
     // virtual media configuration) get a stable object reference, not inline
--- a/docs/proposals/automated-seed-image-generation.md
+++ b/docs/proposals/automated-seed-image-generation.md
@@ -92,9 +92,10 @@ The manual seed generation workflow has several pain points:
   the seed, retain spoke access during generation via the spoke's admin
   kubeconfig retrieved from the hub (which survives ACM teardown and writes
   nothing to the spoke, so nothing is baked into the seed), and do **not**
-  restore ACM afterward. The controller
-  never deletes the `ManagedCluster` itself; the operator deletes the
-  ProvisioningRequest to reclaim the hardware once the seed is captured.
+  restore ACM afterward. The controller never deletes the `ManagedCluster`
+  itself, and it **never auto-deletes the ProvisioningRequest** — results
+  stay readable as conditions on the PR until the operator deletes it to
+  reclaim the hardware once the seed is captured.
 - Reuse the existing `upgradeDefaults` / `upgradeParameters`
   merge-and-validate infrastructure rather than introducing a new CRD or
   controller.
@@ -130,10 +131,15 @@ The manual seed generation workflow has several pain points:
 The seed SNO cluster must be provisioned with the following already in place:
 
 - **LifeCycle Agent (LCA) Operator**: Must be installed as part of the initial cluster configuration via the ClusterTemplate's `policyTemplateDefaults`. The controller validates LCA is present before proceeding with seed generation.
-- **`var-lib-containers` partition**: The ClusterTemplate's `clusterInstanceDefaults` must include the MachineConfig that creates a separate `/var/lib/containers` partition shared between stateroots.
+- **`var-lib-containers` partition**: The ClusterTemplate's `clusterInstanceDefaults` must include the MachineConfig that creates a separate `/var/lib/containers` partition. This is a **seed-generation** prerequisite on the seed cluster:
+  LCA validates a shared `/var/lib/containers` before it will generate the seed, independent of whether the resulting seed is later consumed by IBI or IBU. (The matching *target*-side shared partition is what is IBU-specific — for the dual-stateroot
+  pivot during upgrade — and is out of scope for this proposal, which only produces the seed and, optionally, the IBI live ISO.)
 - **OADP Operator**: Required by LCA for backup/restore operations during seed generation.
 
 These are the responsibility of the cluster template author. The controller validates their presence but does not install them.
+
+Live ISO generation (Phase 4) is **IBI-only and optional**: it is driven by the `liveISO` block and produces the bootable installation ISO that IBI needs to install bare-metal targets. A seed intended only for IBU (in-place upgrade of an existing
+cluster) needs no ISO — omit `liveISO` and the controller stops after the seed image is pushed.
 
 ## Proposed API Changes
 
@@ -1374,6 +1380,9 @@ reference when mounting the ISO via virtual media. The controller does not confi
 responsibility of the hardware manager or the BMC pre-configuration
 workflow. The controller's role is to validate the cert, use it during pre-flight checks, and surface it via the ConfigMap reference for downstream consumption.
 
+BMC-side certificate management proper — for example uploading CA certificates to BMCs via the Redfish `ImportCertificate` action as a day-0/day-2 operation — is intentionally **out of scope** for this proposal and is better handled as a separate
+certificate-management feature (consistent with the Non-Goals). This proposal deliberately stops at validating the ISO server certificate and surfacing it for downstream consumers, rather than taking on BMC trust provisioning.
+
 ### 10. Registry pre-flight check limitations
 
 **Challenge**: The registry validation in Phase 1 runs from the hub cluster, which may have different network access than the spoke (where the actual image push occurs).
@@ -1388,6 +1397,11 @@ may still fail. In disconnected environments, the `seedImage` pull-spec should r
 Detach the spoke entirely by deleting the ClusterInstance/ManagedCluster, as the manual workflow does today. Rejected because deleting the ClusterInstance risks triggering deprovisioning of the running spoke and discards the hub-side record of the cluster. The chosen approach instead
 keeps the `ManagedCluster` intact and removes only the spoke-side ACM agent state (klusterlet + addons), achieving the same seed cleanliness without touching the provisioning lifecycle; the operator reclaims the hardware by deleting the ProvisioningRequest when the seed is captured.
 (Spoke access is obtained the same way either way — the hub-held admin kubeconfig survives klusterlet teardown regardless — so the access credential is not what distinguishes this alternative.)
+
+Taking on spoke cleanup ourselves does add responsibility that deleting the `ManagedCluster` would otherwise leave to ACM, and that responsibility must be bounded so it does not become a maintenance burden as ACM adds features. The design
+deliberately leans on ACM's **own teardown primitives** rather than a hardcoded inventory of agent objects: it deletes the `Klusterlet` CR (so the klusterlet operator tears down *all* of its current and future registration/work agents and
+namespaces) and enumerates and deletes **all** `ManagedClusterAddOn` CRs (so new addons are covered without code changes). The residual risk is limited to spoke state that is neither klusterlet-managed nor modeled as a `ManagedClusterAddOn` — the
+observability secrets in [Challenge 1](#1-incomplete-acm-cleanup-contaminates-the-seed-image) are the known example, and the pattern there (prevent creation at the source via a label rather than chase deletions) is the template for any future such case.
 
 ### B. Separate SeedGenerationRequest CRD
 
--- a/docs/proposals/automated-seed-image-generation.md
+++ b/docs/proposals/automated-seed-image-generation.md
@@ -1084,11 +1084,14 @@ type SeedGenerationStatus struct {
     ISOServerCACertRef *ConfigMapKeyRef `json:"isoServerCACertRef,omitempty"`
 }
 
-// ConfigMapKeyRef references a single key in a ConfigMap.
+// ConfigMapKeyRef references a single key in a ConfigMap. Key is always
+// "ca-bundle.crt" for ISOServerCACertRef: the IBI pre-provisioning consumer
+// reads a name-only reference under that fixed key, so the producer must not
+// emit any other key or the consumer silently fails to resolve the bundle.
 type ConfigMapKeyRef struct {
     Name      string `json:"name"`
     Namespace string `json:"namespace"`
-    Key       string `json:"key"`
+    Key       string `json:"key"` // always "ca-bundle.crt"
 }
 ```
 
@@ -1141,6 +1144,9 @@ spec:
           urlBase: https://iso-server.example.com/ibi/
       clusterUpgradeTimeout: "3h"
   templateParameterSchema:
+    # Abbreviated: field types only. The `additionalProperties: false` guard
+    # (required at every object level) and some fields are elided — the canonical
+    # "Schema in templateParameterSchema" is authoritative; copy from there.
     properties:
       # ... standard provisioning parameters ...
       upgradeParameters:
--- a/docs/proposals/automated-seed-image-generation.md
+++ b/docs/proposals/automated-seed-image-generation.md
@@ -1,0 +1,1521 @@
+<!--
+SPDX-FileCopyrightText: Red Hat
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Proposal: Automated Seed Image and Live ISO Generation via ProvisioningRequest API
+
+## Table of Contents
+
+- [Background](#background)
+- [Problem Statement](#problem-statement)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [Prerequisites](#prerequisites)
+- [Proposed API Changes](#proposed-api-changes)
+  - [New key: `seedGeneration`](#new-key-seedgeneration)
+  - [ProvisioningRequest overrides](#provisioningrequest-overrides)
+  - [Schema in templateParameterSchema](#schema-in-templateparameterschema)
+- [Controller Workflow](#controller-workflow)
+  - [Functional Overview](#functional-overview)
+  - [Trigger and Dispatch](#trigger-and-dispatch)
+  - [Phase 1: Validation](#phase-1-validation)
+  - [Phase 2: ACM Agent Cleanup](#phase-2-acm-agent-cleanup)
+  - [Phase 3: Trigger Seed Generation](#phase-3-trigger-seed-generation)
+  - [Phase 4: Live ISO Generation](#phase-4-live-iso-generation)
+  - [Phase 5: Completion and Cleanup](#phase-5-completion-and-cleanup)
+- [New Status Fields](#new-status-fields)
+- [New Condition Type](#new-condition-type)
+- [Timeout](#timeout)
+- [Sample: End-to-End ClusterTemplate](#sample-end-to-end-clustertemplate)
+- [Challenges and Mitigations](#challenges-and-mitigations)
+  - [1. Incomplete ACM cleanup contaminates the seed image](#1-incomplete-acm-cleanup-contaminates-the-seed-image)
+  - [2. Spoke client availability during seed generation](#2-spoke-client-availability-during-seed-generation)
+  - [3. ACM self-healing during cleanup phase](#3-acm-self-healing-during-cleanup-phase)
+  - [4. Seed cluster afterlife and PR terminal state](#4-seed-cluster-afterlife-and-pr-terminal-state)
+  - [5. Seed SNO prerequisites](#5-seed-sno-prerequisites)
+  - [6. Standalone access token lifetime and cleanup](#6-standalone-access-token-lifetime-and-cleanup)
+  - [7. ISO generation Job storage requirements](#7-iso-generation-job-storage-requirements)
+  - [8. Disconnected environment registry configuration](#8-disconnected-environment-registry-configuration)
+  - [9. HTTPS ISO server certificate management](#9-https-iso-server-certificate-management)
+  - [10. Registry pre-flight check limitations](#10-registry-pre-flight-check-limitations)
+- [Alternatives Considered](#alternatives-considered)
+  - [A. Full ACM detachment (delete ManagedCluster)](#a-full-acm-detachment-delete-managedcluster)
+  - [B. Separate SeedGenerationRequest CRD](#b-separate-seedgenerationrequest-crd)
+  - [C. External job/pipeline](#c-external-jobpipeline)
+  - [D. Controller-managed LCA installation](#d-controller-managed-lca-installation)
+  - [E. ISO generation via external pipeline (Tekton, CI/CD)](#e-iso-generation-via-external-pipeline-tekton-cicd)
+  - [F. Default to public release image registry](#f-default-to-public-release-image-registry)
+- [Implementation Constants](#implementation-constants)
+- [Effort Estimate](#effort-estimate)
+- [Open Questions](#open-questions)
+
+## Background
+
+Today, generating a seed image and live ISO for IBI-based cluster provisioning is a multi-step manual process (documented in `docs/user-guide/ibi-based-cluster-provisioning.md`). An operator must:
+
+1. Provision a reference SNO cluster via ProvisioningRequest (with a specific `var-lib-containers` MachineConfig)
+2. Annotate the PR with `clcm.openshift.io/skip-cleanup` and delete it to detach the spoke from ACM
+3. SSH into the spoke, install the LifeCycle Agent (LCA) operator, create a `seedgen` Secret and a `SeedGenerator` CR
+4. Monitor the SeedGenerator until the seed image is pushed to the container registry
+5. Obtain the `openshift-install` binary matching the OCP version, create an `ImageBasedInstallationConfig`, and run `openshift-install image-based create image` to produce a live ISO
+6. Upload the ISO to an HTTPS server accessible from the BMC management network
+7. Pre-provision servers using the ISO via BMC virtual media
+
+This proposal automates steps 1 through 6 as a first-class operation within the ProvisioningRequest API.
+
+## Problem Statement
+
+The manual seed generation workflow has several pain points:
+
+- **Error-prone**: Requires precise ordering of annotate → delete → install LCA → create resources → monitor, plus matching the correct `openshift-install` binary version to build the ISO
+- **Untracked**: No status reporting through the O-Cloud Manager API; operators must SSH and tail logs
+- **Inconsistent**: Each team may follow a slightly different procedure
+- **Not scalable**: Generating seed images and ISOs for multiple hardware/OCP configurations requires repeating the full manual procedure
+- **Disconnected complexity**: In disconnected environments (the predominant use case), operators must also manage release image mirrors and ensure the `openshift-install` binary, seed image, and OCP release images are all accessible from the correct registries
+
+## Goals
+
+- Automate seed image generation (annotate → detach → install/create LCA
+  resources → monitor → push) as a first-class operation driven by the
+  existing ProvisioningRequest API, eliminating the manual SSH-based
+  procedure.
+- Automate live installation ISO generation and upload to the HTTPS server
+  as part of the same workflow, matching the correct `openshift-install`
+  version automatically.
+- Report progress and results through the O-Cloud Manager API via a typed
+  `SeedGenerationCompleted` condition and `SeedGenerationStatus` fields, so
+  operators no longer need to SSH and tail logs.
+- Treat the seed cluster as a disposable factory: remove all ACM agent
+  state (klusterlet and addons) from the spoke so it does not contaminate
+  the seed, retain spoke access during generation via a standalone
+  long-lived token, and do **not** restore ACM afterward. The controller
+  never deletes the `ManagedCluster` itself; the operator deletes the
+  ProvisioningRequest to reclaim the hardware once the seed is captured.
+- Reuse the existing `upgradeDefaults` / `upgradeParameters`
+  merge-and-validate infrastructure rather than introducing a new CRD or
+  controller.
+- Validate prerequisites and fail fast with a clear
+  `PreconditionChecksFailed` condition (registry push access, release image
+  reachability, TLS/CA trust, upload credentials) before the lengthy
+  generation begins.
+- Support disconnected environments as the primary use case, auto-deriving
+  mirror mappings and registry trust from the hub (local mirrors, private CA)
+  so the template author supplies only the required `releaseImage`.
+
+## Non-Goals
+
+- Installing or managing the prerequisite operators (LCA, OADP) or the
+  `var-lib-containers` MachineConfig — these remain the cluster template
+  author's responsibility; the controller only validates their presence.
+- Validating hardware-specific seed prerequisites the hub cannot inspect
+  (CPU topology alignment, FIPS, proxy, IP-version match with target SNOs);
+  see [Challenge 5](#5-seed-sno-prerequisites).
+- Configuring BMC virtual media or BMC CA trust to consume the generated
+  ISO — that belongs to the hardware manager / pre-provisioning workflow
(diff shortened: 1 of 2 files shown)


browsell:
@coderabbitai full review

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1849 @@
...
+      # prepared; marking it externallyProvisioned here would tell BMO the OS
+      # is already installed and skip/short-circuit the very boot we need.
+      # externallyProvisioned is set ONLY on a verified success callback,
+      # in transitionToIBIReady below.
+  → NEW: preProvisionHost (ordered so every write is recoverable — see
+    "Crash-safety of the boot handoff" below; each step is idempotent and the
+    next reconcile resumes from the first incomplete one):
+      # RECOVERY: if Status.PreProvisioningAttempt is already set, this is a
+      # resumed attempt. Do NOT mint a new id/token. Reuse the existing attempt
+      # and its token Secret, and re-drive the remaining steps idempotently.
```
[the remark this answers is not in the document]
browsell: [Comment authored by Claude]

Resolved in 52f8d759. Added a durable RetryNotBefore *metav1.Time status field. On a budgeted failure the operator now sets RetryNotBefore = now + RetryBackoff and clears PreProvisioningAttempt in the same status update, so the retired attempt (whose token is already invalidated) can never be resumed. Recovery classifies three post-failure states purely from persisted status: attempt set → resume; attempt unset + RetryNotBefore in the future → wait out the backoff; attempt unset + RetryNotBefore reached → mint the next attempt. The backoff remainder is recomputed from RetryNotBefore on every reconcile, so a restart that loses the in-memory RequeueAfter still honors it. Propagated to the status struct, the API checklist, the crash-safety invariants, the retry section, and a new fault-injection test ("retry recorded, process restarted mid-backoff").

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1849 @@
...
+      → 1. mint intent (only when no attempt is in flight):
+          → mint attempt id + per-attempt token
+          → write the token to its own Secret, named deterministically from
+            the attempt id (create-or-adopt). The token lives here, NOT in
+            status, so it must be persisted BEFORE the attempt id is recorded —
+            an attempt id in status with no matching token Secret is
+            unrecoverable.
+      → 2. record the attempt (ONE status update, flushed):
```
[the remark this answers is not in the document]
browsell: [Comment authored by Claude]

Resolved in 52f8d759. The per-attempt token Secret is now owned by the ICI (ownerReference) and labelled ibi.openshift.io/attempt=<id>. Recovery adds a case: when PreProvisioningAttempt is unset but a pending token Secret owned by this ICI with that label exists, the operator adopts that Secret's attempt id and continues at the record-attempt step, rather than minting a second token. This makes the "token Secret written, status not" window resume the same attempt with no orphaned Secret and no duplicate token. The corresponding fault-injection test now asserts the pending Secret is adopted by owner + label instead of being re-minted.

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1849 @@
...
+    `After=ibi-callback-data.service`. It is a mistake to rely on
+    `RequiredBy=`/`WantedBy=` in the *data* unit's `[Install]` section for
+    this: `[Install]` directives take effect only when the unit is
+    `enable`d, so if enablement is ever missed the reporter would run before
+    (or without) its data files. Both units are additionally marked
+    `enabled: true` in the ignition so they are active on first boot, but the
+    correctness guarantee comes from the runtime `Requires=`/`After=`, not
+    from enablement. With `Requires=`, if the data mount fails (device
+    absent) the reporter is not started, surfacing as a timeout rather than a
+    callback with an empty URL/token. This enablement-plus-runtime-dependency
```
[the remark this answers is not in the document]
browsell: [Comment authored by Claude]

Resolved in 52f8d759. ibi-callback-data.service is now Type=oneshot with RemainAfterExit=yes, so it stays active (exited) after materializing the files and the reporter's Requires=ibi-callback-data.service stays satisfied for the whole boot (without it the oneshot would go inactive when ExecStart returns and systemd could stop the reporter). Documented in both the reporter unit's [Unit] comment and the Challenge 4 mount/materialize bullet, and the integration test now asserts the data unit stays active (exited) after materialization and the reporter actually runs.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+- [Effort Estimate](#effort-estimate)
+- [Open Questions](#open-questions)
+
+## Background
+
+Today, generating a seed image and live ISO for IBI-based cluster provisioning is a multi-step manual process (documented in `docs/user-guide/ibi-based-cluster-provisioning.md`). An operator must:
+
+1. Provision a reference SNO cluster via ProvisioningRequest (with a specific `var-lib-containers` MachineConfig)
+2. Annotate the PR with `clcm.openshift.io/skip-cleanup` and delete it to detach the spoke from ACM
+3. SSH into the spoke, install the LifeCycle Agent (LCA) operator, create a `seedgen` Secret and a `SeedGenerator` CR
```
donpenney, line 60: You don't need to ssh into the spoke for this, as long as you get the kubeconfig from the hub before you delete the cluster. You can then use that kubeconfig to remotely create the SeedGenerator.

Wouldn't the LCA already be installed on the spoke, or is it not part of the RDS?

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+7. Pre-provision servers using the ISO via BMC virtual media
+
+This proposal automates steps 1 through 6 as a first-class operation within the ProvisioningRequest API.
+
+## Problem Statement
+
+The manual seed generation workflow has several pain points:
+
+- **Error-prone**: Requires precise ordering of annotate → delete → install LCA → create resources → monitor, plus matching the correct `openshift-install` binary version to build the ISO
+- **Untracked**: No status reporting through the O-Cloud Manager API; operators must SSH and tail logs
```
donpenney, line 73: The SeedGenerator CR status can be monitored remotely with the spoke's kubeconfig. If the user wants to watch the logs of the seed generation process itself, however, they'd need to SSH in

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+  as part of the same workflow, matching the correct `openshift-install`
+  version automatically.
+- Report progress and results through the O-Cloud Manager API via a typed
+  `SeedGenerationCompleted` condition and `SeedGenerationStatus` fields, so
+  operators no longer need to SSH and tail logs.
+- Treat the seed cluster as a disposable factory: remove all ACM agent
+  state (klusterlet and addons) from the spoke so it does not contaminate
+  the seed, retain spoke access during generation via a standalone
+  long-lived token, and do **not** restore ACM afterward. The controller
+  never deletes the `ManagedCluster` itself; the operator deletes the
```
donpenney: This seems like it shifts a lot of responsibility to the O-Cloud Manager. By deleting the ManagedCluster, the onus is on ACM to clean up the spoke. Taking on that cleanup ourselves seems like it could cause some future maintenance issues, as new ACM features are added.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+  ProvisioningRequest to reclaim the hardware once the seed is captured.
```
donpenney: I interpreted the previous point as that we're reporting progress and results as conditions on the `ProvisioningRequest`, but here we're automatically deleting it when the job is complete.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+- Supporting ISO upload mechanisms beyond SCP/SFTP in the initial
+  implementation (see [Open Questions](#open-questions)).
+
+## Prerequisites
+
+The seed SNO cluster must be provisioned with the following already in place:
+
+- **LifeCycle Agent (LCA) Operator**: Must be installed as part of the initial cluster configuration via the ClusterTemplate's `policyTemplateDefaults`. The controller validates LCA is present before proceeding with seed generation.
+- **`var-lib-containers` partition**: The ClusterTemplate's `clusterInstanceDefaults` must include the MachineConfig that creates a separate `/var/lib/containers` partition shared between stateroots.
+- **OADP Operator**: Required by LCA for backup/restore operations during seed generation.
```
donpenney, line 137: I think this is only necessary if using the seed image for IBU, not for IBI? And along those lines, if only using the seed image for IBU, we don't need an ISO.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+   - **Pod read permissions are bound per-namespace, not cluster-wide:**
+     `get`/`list`/`watch` on `pods` is granted through `RoleBinding` objects
+     in the specific agent namespaces the poll watches
+     (`open-cluster-management-agent*`), so the standalone token cannot read
+     pod specs in unrelated namespaces.
+   - **Self-cleanup permissions (so Phase 5 can revoke this very
+     credential).** After ACM/MSA teardown, the standalone token is the *only*
+     path back to the spoke, so it must be able to delete its own artifacts.
+     The grants therefore also include `delete` on the `seedgen` Secret and
+     SeedGenerator CR (namespaced, via the LCA-namespace `RoleBinding`),
```
donpenney: nit: The `SeedGenerator` CR is cluster-scoped. Only the `seedgen` Secret is namespaced

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+     channel it already trusts (`sha256sum` on the remote `remotePath`) and
+     compares it to the locally computed digest; a mismatch or an inability
+     to obtain the remote digest is a terminal `Failed`. Because SSH already
+     gives an authenticated, integrity-checked channel to the exact file just
+     written, this comparison is always available and does not depend on the
+     server exposing anything extra.
+   - The `remotePath` → effective-ISO-URL relationship is **deterministic and
+     explicitly defined**, not merely asserted, so the artifact verified over
+     SSH is provably the same object addressed by the recorded URL. The upload
+     endpoint (`uploadSecretRef.host` + `remotePath`) and the HTTPS origin
```
donpenney: The https bit has me thinking about certificate management for the BMCs, and how we would update them

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+**Challenge**: The seed SNO has hardware-specific prerequisites (CPU topology alignment, FIPS, proxy, IP version match with target SNOs) that cannot be fully validated by the hub controller.
+
+**Mitigation**: The controller validates what it can (OCP version, `var-lib-containers` partition, LCA operator presence, OADP operator presence, registry credentials). Hardware prerequisites are the cluster template author's responsibility — they must ensure the ClusterTemplate's
+`clusterInstanceDefaults` and `policyTemplateDefaults` produce a seed SNO that meets the documented prerequisites.
+
+### 6. Standalone access token lifetime and cleanup
+
+**Challenge**: With ACM fully removed, spoke access depends entirely on the standalone credential minted in Phase 2. It must not expire mid-operation; and if the workflow is abandoned, a valid credential could be left behind on the spoke.
+
+**Mitigation**: The credential is a **non-expiring** `kubernetes.io/service-account-token` Secret (not a TTL-bounded `TokenRequest` token), so it cannot lapse mid-operation.
```
donpenney: This new secret would become part of the seed image. Do we need to include handling to find and delete this secret after a successful IBI or IBU?

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+
+**Mitigation**: The `uploadSecretRef` Secret carries an optional `caCert` field containing the PEM-encoded CA
+bundle for the ISO server. Phase 1 validates the certificate chain by performing a TLS handshake against the
+`urlBase` host. On completion, the controller **materializes the PEM bundle into a dedicated ConfigMap** (in the
+O-Cloud Manager namespace, named deterministically from the ProvisioningRequest) and records a typed reference to
+it in `SeedGenerationStatus.ISOServerCACertRef` — a stable object reference, not inline certificate data or a name
+for an object that is never created. Downstream consumers (BMC configuration, hardware manager) resolve that
+reference when mounting the ISO via virtual media. The controller does not configure BMC trust — that is the
+responsibility of the hardware manager or the BMC pre-configuration
+workflow. The controller's role is to validate the cert, use it during pre-flight checks, and surface it via the ConfigMap reference for downstream consumption.
```
donpenney: Maybe we need to consider certificate management as a separate feature, something that would use the `ImportCertificate` redfish command to upload certs as needed, as part of day0/day2 operations

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1493 @@
...
+Roughly 4 working sessions to implement:
+
+- **Session 1**: API types, constants, conditions, `parseUpgradeConfig` extension, CT validation (registry pre-flight, release image check, TLS handshake, upload Secret validation), sample YAMLs
+- **Session 2**: Core seed generation state machine (Phases 1-3, 5), ACM removal helpers (standalone token mint, `Klusterlet` CR delete, addon-CR delete), terminal-state / config-reconcile suppression, spoke and hub RBAC, wiring into the main reconciler
+- **Session 3**: ISO generation — hub mirror-config derivation, PVC creation, Job template (release image extract, config generation, ISO build, SCP upload, HTTPS post-upload verification), pull secret merging, Job lifecycle monitoring, cleanup
+- **Session 4**: Tests, edge cases, docs update, `make ci-job` pass
+
+## Open Questions
+
+1. **Is deleting the `Klusterlet` CR + all `ManagedClusterAddOn` CRs sufficient to fully remove ACM agent state?** Option B relies on the klusterlet operator's own teardown (triggered by deleting the `Klusterlet` CR) rather than a hardcoded list of Deployments/DaemonSets. This should
```
donpenney, line 1360: We may also need to delete some secrets related to observability. In particular, I believe the observability-addon creates `hub-alertmanager-router-ca-*` and `observability-alertmanager-accessor-*` secrets in the `openshift-monitoring` and `openshift-user-workload-monitoring` namespaces that we may need to cleanup.

browsell: [Comment authored by Claude]

Good catch — and it pushed me toward a cleaner fix than deleting those secrets after the fact.

Rather than enumerating and deleting the observability secrets during teardown, the proposal now prevents them from ever being created. RHACM's multicluster-observability-operator auto-deploys the observability-addon (metrics-collector + observability-endpoint-operator) to every managed cluster, and that addon is what creates hub-alertmanager-router-ca-* and observability-alertmanager-accessor-* in
openshift-monitoring / openshift-user-workload-monitoring. Since the seed cluster is a disposable, purpose-built cluster we provision, the seed ClusterTemplate sets the observability: disabled label on the ManagedCluster (via clusterInstanceDefaults) so it's present from import time. The operator then never deploys the addon, and those secrets are never created — so they can't contaminate the seed.

I went with this over the delete-during-cleanup approach for two reasons:

1. The exact secret set drifts across RHACM/MCE versions, so a hardcoded deletion list is fragile.
2. Spoke-side secret deletion would require widening the standalone credential's RBAC into openshift-monitoring /openshift-user-workload-monitoring, enlarging its blast radius — which the credential design is deliberately trying to minimize.

Updated in Challenge 1 ("Incomplete ACM cleanup contaminates the seed image"). I also added Open Question 4 to capture the one case the label doesn't cover — repurposing a cluster that already had observability deployed — where a bounded residual sweep (and the corresponding RBAC widening) would be the fallback, rather than relaxing the preventive default.

move on to finding Ref: RHACM Observability — excluding specific managed clusters with observability: disabled.

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

Agreed on both points, and the design now reflects them:

1. No SSH. Phase 2 retrieves the spoke's admin kubeconfig from the hub (adminKubeconfigSecretRef) and drives everything — creating the seedgen Secret and the SeedGenerator CR, monitoring status, deleting the Klusterlet — remotely via that client. The only thing SSH would add is tailing the live seed-generation logs on the node, which the automated flow surfaces through the SeedGenerationCompleted condition /SeedGenerationStatus instead. (See the finding on the standalone-token thread — using the hub kubeconfig also means nothing new is baked into the seed.)

2. LCA is a prerequisite, not installed by us. Yes — LCA must already be present on the spoke; it's listed under Prerequisites and validated in Phase 1 (installed via the ClusterTemplate's policyTemplateDefaults, per the RDS). The controller only checks for it and fails fast if iit's missing (PreconditionChecksFailed) rather than installing it just-in-time (see Alternative D).

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

Good catch — you're right that the minted secret lands in etcd at capture time and would be cloned into every cluster deployed from the seed. Rather than adding post-IBI/IBU handling to find and delete it, I removed the thing that creates the problem.

The proposal no longer mints a standalone ServiceAccount token on the spoke. Instead, Phase 2 retrieves the spoke's admin kubeconfig from the  hub (ClusterDeployment.spec.clusterMetadata.adminKubeconfigSecretRef) and builds the spoke client from it. That kubeconfig authenticates with an embedded system:admin client certificate, which:

- Survives ACM teardown — the cert is independent of the klusterlet and the ManagedServiceAccount, so deleting the Klusterlet CR doesn't affect it. It only stops working if the spoke's API server cert rotates under a new CA, which happens at restore time (recert on the target), not during generation.
- Writes nothing to the spoke's etcd — no ServiceAccount, token Secret, or RBAC is created, so no new credential is captured into the seed. There's nothing to find-and-delete afterward.

This also let me drop the whole spoke-side least-privilege RBAC block and the self-cleanup RBAC paradox — the controller now needs only a hub-side get on the admin-kubeconfig Secret, and Phase 5 spoke cleanup is just the transient seedgen Secret + SeedGenerator CR.

Updated in Phase 2/3/5, the hub RBAC contract, Challenge 6 (renamed "Spoke access after ACM removal"), Challenge 1/2, Alternatives A, and the effort estimate. This dovetails with your line-60 comment (use the hub kubeconfig remotely instead of SSH) — that's now the mechanism.

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

Fair concern. We deliberately kept the ManagedCluster intact rather than deleting it (that's Alternative A) — deleting the ClusterInstance/ManagedCluster risks triggering deprovisioning of the still-running seed cluster and discards the hub-side record — so yes, we take on the spoke cleanup ourselves.

To keep that from becoming a maintenance treadmill, the design leans on ACM's own teardown primitives instead of a hardcoded inventory of agent objects:

- Delete the Klusterlet CR → the klusterlet operator tears down all of its registration/work agents and namespaces, including ones added in future ACM versions.
- Enumerate and delete all ManagedClusterAddOn CRs → new addons are covered without code changes.

The residual risk is exactly the category that's neither klusterlet-managed nor a ManagedClusterAddOn — and the observability secrets you flagged on line 1486 are the known example of that. The pattern we landed on there (prevent creation at the source via the observability:  disabled label, rather than chase deletions) is the template for any future such case. I've written this rationale into Alternative A.

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

Good catch on the ambiguity — the controller does not auto-delete the ProvisioningRequest. Results stay on the PR as SeedGenerationCompleted /SeedGenerationStatus conditions for the operator to read; the PR is deleted only when the operator chooses to reclaim the hardware. Those two facts aren't in tension — the auto-behavior is reaching a terminal condition, and the human decides when to tear down. I've made this
explicit in the Goals bullet ("never auto-deletes the ProvisioningRequest — results stay readable as conditions … until the operator deletes it").

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

  Split answer, since the partition and the ISO land on different sides:

- var-lib-containers partition — this one is a seed-generation prerequisite on the seed cluster itself: LCA validates a shared /var/lib/containers before it will generate the seed at all, independent of whether the seed is later consumed by IBI or IBU (OKD IBI prep lists the shared container dir under seed prerequisites). What's IBU-specific is the target-side shared partition (for the dual-stateroot pivot during upgrade), which is out of scope here since this proposal only produces the seed.
-  ISO — agreed, the live ISO is IBI-only. That's why Phase 4 (liveISO) is optional: for an IBU-only seed you omit liveISO and the controller stops after pushing the seed image.

  I've clarified both in the Prerequisites section.

docs/proposals/automated-seed-image-generation.md
browsell: [Comment authored by Claude]

Agreed — BMC-side certificate management (e.g. uploading CA certs via the Redfish ImportCertificate action as a day-0/day-2 operation) is out of scope for this proposal and fits better as a separate cert-management feature. This proposal deliberately stops at validating the ISO server certificate and surfacing it (via the ISOServerCACertRef ConfigMap) for downstream consumers; it doesn't take on BMC trust provisioning. I've added a note to Challenge 9 pointing at that boundary, consistent with the existing Non-Goals.

docs/proposals/automated-seed-image-generation.md
donpenney: Ok, so in this statement, the "operator" being referred to is the admin/user, as opposed to the O-Cloud Manager operator - I read it as the latter.

docs/proposals/automated-seed-image-generation.md
donpenney: My comment was on the "OADP Operator" item, not the "var-lib-containers" item

docs/proposals/automated-seed-image-generation.md
donpenney: If that works, that's a really interesting solution to this issue.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1357 @@
...
+      additionalProperties: false
+      properties: { name: { type: string } }
+      required: [name]
+    recertImage:
+      type: string
+      description: >
+        Template-owned; when set it must be an allowlisted, immutable digest
+        reference (repo@sha256:<digest>), because Phase 3 passes it to the
+        SeedGenerator CR where it runs as executable code during seed
+        generation.
```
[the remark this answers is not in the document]
browsell: [Comment authored by Codex]

I reviewed this finding and confirmed that LCA does not enforce a digest or registry allowlist for recertImage. This is intentional: the field is template-owned, selected only by trusted template authors, and excluded from the PR-facing schema.

I clarified this trust boundary in commit af006cfe. Deployments requiring digest pinning or registry restrictions can enforce them through platform or admission policy. Therefore, this is not an untrusted PR override path; it is an explicit trusted-input decision.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1428 @@
...
+     SeedGenerator CR (namespaced, via the LCA-namespace `RoleBinding`), and
+     `delete` on the credential's own `ServiceAccount` and token `Secret`
+     (namespaced) plus its `RoleBinding`s, `ClusterRole`, and
+     `ClusterRoleBinding` — the cluster-scoped deletes **restricted by
+     `resourceNames`** to exactly the objects this run created (deterministic
+     names), so the token cannot delete unrelated RBAC. `delete` is not
+     escalation, so this needs no additional `escalate`/`bind`. To avoid
+     cutting off its own access mid-cleanup, Phase 5 deletes the
+     credential-bearing `ServiceAccount` and token `Secret` **last**, after all
+     other spoke artifacts and RBAC objects are gone.
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The proposal now defines create-or-adopt handling for every generated resource using per-run ownership labels and UID checks. Matching resources are adopted; unrelated same-name resources fail closed. Restart-after-create tests are required. The design also clarifies that Phase 2 uses the admin kubeconfig and creates no ServiceAccount, token Secret, or RBAC resources.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1428 @@
...
+- `batch/jobs`: `create`, `get`, `list`, `watch`, `delete` (create and
+  monitor the ISO-build Job; delete it in Phase 5).
+- `core/persistentvolumeclaims`: `create`, `get`, `delete` (the ~20GB build
+  workspace PVC).
+- `core/secrets`: `create`, `get`, `delete` (per-run credential copies) and
+  `get` on the source Secrets in the ClusterTemplate namespace and
+  `openshift-config/pull-secret`.
+- `core/configmaps`: `create`, `get`, `delete` (the per-run mirror/CA
+  ConfigMap) and `get` on the hub image-config `additionalTrustedCA`.
+- `core/pods` and `core/pods/log`: `get`, `list` (read the build pod and its
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The RBAC contract now explicitly grants the controller namespaced get access to the ConfigMap referenced by liveISO.additionalTrustBundleConfigMapRef in the ClusterTemplate namespace. Authorization tests cover both the controller namespace and the referenced ClusterTemplate namespace.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1399 @@
...
+     is only a "served over HTTPS" liveness probe, never the integrity check.
+
+   The verified SHA-256 digest and the resolved per-run effective ISO URL are recorded
+   in status (`ISODigest`, `ISOURL`) so downstream consumers can pin the
+   exact artifact.
+
+**Job pod spec considerations:**
+
+- The Job image uses `oc` (available from the OCP CLI tools image) and `openshift-install` (extracted at runtime)
+- **PVC for build workspace**: The controller creates a dedicated PVC (~20GB), mounted into the Job pod as the working directory, because `openshift-install` pulls the full seed image and generates an ISO, exceeding typical ephemeral storage limits. It is created before the Job and deleted
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The proposal now requires an operator-owned ISO-build image pinned by immutable digest containing oc, POSIX shell, ssh, scp, and sha256sum. openshift-install is extracted from the pinned release image. A capability-check Pod runs from the same image before the PVC or Job is created and fails closed if a required tool is unavailable.

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1805 @@
...
+# not python3 (NOT present in the RHCOS live ISO — invoking it would fail and,
+# under `set -e`, exit before curl, turning every callback into a timeout).
+# --arg passes each value as a literal string, so no field injection. The
+# attempt id ties this callback to the exact boot the operator waits on (see
+# PreProvisioningAttempt); the operator rejects any mismatch.
+PAYLOAD=$(jq -nc \
+  --arg attempt "${CALLBACK_ATTEMPT}" \
+  --arg status "${STATUS}" \
+  --arg message "${MSG}" \
+  '{attempt: $attempt, status: $status, message: $message}')
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. Failure callbacks now use a fixed message (IBI preparation failed; inspect host-local logs) instead of forwarding journal output. Detailed logs remain local to the host, and tests verify that secret-like journal content is not persisted in the callback or ICI status.

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1805 @@
...
+
+```go
+// POST /callbacks/{namespace}/{name}/status
+// The bearer token is carried ONLY in the Authorization header; the
+// endpoint accepts no ?token= query parameter (see auth notes below).
+type CallbackPayload struct {
+    Attempt string `json:"attempt"` // must equal status.preProvisioningAttempt
+    Status  string `json:"status"`  // "success" or "failure"
+    Message string `json:"message"` // detail message
+}
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The callback endpoint now enforces a 16 KiB request-body limit before decoding or persisting the payload, rejects oversized requests with 413, and authenticates the bearer token before JSON decoding. Tests cover oversized and invalid-token requests.

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1805 @@
...
+media boot:
+
+```go
+patch := client.MergeFrom(bmh.DeepCopy())
+bmh.Spec.Image = &bmh_v1alpha1.Image{
+    URL:        preProvisioning.ISOURL,
+    DiskFormat: pointer.String("live-iso"),
+    // Populated from PreProvisioningConfig.ISODigest when set.
+    Checksum:     preProvisioning.ISODigest, // "sha256:<hex>"
+    ChecksumType: bmh_v1alpha1.SHA256,
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The proposal no longer treats Checksum or ChecksumType as an integrity control for live-iso, since BMO clears those fields before passing boot_iso to Ironic. Integrity is instead based on operator-side digest verification plus an immutable, allowlisted URL that returns the same bytes directly. Deployments unable to guarantee that contract are rejected.

docs/proposals/ibi-server-pre-provisioning.md
```diff
@@ -0,0 +1,1809 @@
...
+server-side request forgery (SSRF) vector. The operator therefore enforces two
+checks **before any byte of `ISOURL` is fetched**, failing closed with
+`PreProvisioningConfigInvalid`:
+
+- **Origin allowlist.** `ISOURL` must resolve to an operator-owned or
+  allowlisted ISO origin (seed-generation upload origins plus operator-configured
+  additions); a URL outside that set is rejected before the fetch, so the
+  operator never requests a caller-chosen host. The allowlist is operator
+  configuration, not template/`ProvisioningRequest` input, so a request author
+  cannot widen it.
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. Operator digest, content, and sidecar requests now disable redirects and fail closed on any 3xx response. Because BMO/Ironic receives only the URL for the later BMC fetch, the proposal also requires the immutable per-run origin to return a direct 200 without redirects, with tests covering off-allowlist redirects.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1388 @@
...
+  templateParameterSchema:
+    # PR-facing schema: exposes only the PR-overridable field(s). Every
+    # template-owned field lives in upgradeDefaults and is validated post-merge
+    # by the controller's effective-object schema — see "Schema: PR-facing and
+    # effective-object". additionalProperties: false rejects any other override.
+    properties:
+      # ... standard provisioning parameters ...
+      upgradeParameters:
+        type: object
+        additionalProperties: false
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. The duplicate type: object declaration was removed from the upgradeParameters sample schema; only the root declaration remains.

docs/proposals/automated-seed-image-generation.md
```diff
@@ -0,0 +1,1399 @@
...
+   exact artifact.
+
+**Job pod spec considerations:**
+
+- **Pinned ISO-build toolchain.** The Job uses an operator-owned image by
+  immutable digest (never a tag) containing `oc`, POSIX shell, `ssh`, `scp`, and
+  `sha256sum`; `openshift-install` is extracted from the pinned release image.
+  Before creating the PVC or Job, Phase 1 runs a short-lived capability-check
+  Pod from the same digest and fails `PreconditionChecksFailed` if a tool or
+  image pull is unavailable. Tests cover complete and missing-tool images.
```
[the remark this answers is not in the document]
browsell: [Comment authored by OpenAI Codex]

Addressed. Phase 1 now runs a short-lived preflight Pod before creating the PVC or Job. It uses the pinned ISO-build image, persisted ReleaseImageDigest, derived mirror/CA configuration, and effective pull secret to run oc adm release extract --command=openshift-install. The extraction must produce an executable; failures result in PreconditionChecksFailed, the Pod is deleted, and no PVC or Job is created. Tests cover extraction failure and a missing openshift-install binary.

PR #4268, opened 2026-09-03
CNF-26808: Drain scale-in nodes concurrently across worker MCPs

Missxiaoguo:
Extract per-node scale-in into scaleInNode and drain listed Ready nodes in parallel.

Extend mno-scale-out e2e to two named worker pools and scale both in one PR. Docs cover appending hosts under the target nodeGroup(s) and adding workers to a new MCP at day-2.

Lab-tested on a fulfilled MNO: 3 masters and two worker pools (2 hosts and 3 hosts). Sequence:
- Scale-in: remove 1 host from each pool in one ProvisioningRequest.
- Scale-out: add 2 hosts back to each pool in one ProvisioningRequest.
- Scale-in: remove 1 host from the first pool and 2 from the second in one ProvisioningRequest.
- Swap: remove 1 host from each pool and add 1 host to each in the same ProvisioningRequest.
- Day-2 new MCP: add a new worker MachineConfigPool through PolicyGenerator and scale-out 1 host into it.

Verified that scale-in or scale-out across multiple worker pools is processed concurrently. Scale-in and scale-out across mulitple pools can also be performed in a single ProvisioningRequest, with scale-in completing first, followed by scale-out.

### DIFF
--- a/docs/dev/e2e-test-overview.md
+++ b/docs/dev/e2e-test-overview.md
@@ -27,8 +27,8 @@ feature, which parses test structure without executing tests.
   - [Handles day2 hardware configuration update with BMH error](#handles-day2-hardware-configuration-update-with-bmh-error)
   - [Handles mid-flight profile change by abandoning stale updates](#handles-mid-flight-profile-change-by-abandoning-stale-updates)
 - [MNO Scale-Out test [mno-scale-out]](#mno-scale-out-test-mno-scale-out)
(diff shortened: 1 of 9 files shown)


Missxiaoguo:
@coderabbitai replying to your nit comment: the error we care about is already logged inside`drainIfReady`

donpenney (approved):

PR #4352, opened 2026-09-09
CNF-26914: Treat unchanged nodes without Configured as complete

Missxiaoguo:
Treat AllocatedNodes without a Configured condition as complete when their current hardware profile matches the desired group profile. This prevents the first day-2 hardware configuration update from selecting unchanged nodes or consuming maxUnavailable capacity.

Include these nodes in single- and multi-node NAR status aggregation, and report when configuration is already up to date because no node required an update.

Tested on cluster.

### DIFF
--- a/api/hardwaremanagement/v1alpha1/conditions.go
+++ b/api/hardwaremanagement/v1alpha1/conditions.go
@@ -51,6 +51,7 @@ type ConditionMessage string
 const (
 	AwaitConfig      ConditionMessage = "Spec updated; awaiting configuration application by the hardware manager"
 	ConfigSuccess    ConditionMessage = "Configuration has been applied successfully"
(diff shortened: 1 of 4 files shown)


api/hardwaremanagement/v1alpha1/conditions.go
```diff
@@ -51,6 +51,7 @@ type ConditionMessage string
 const (
 	AwaitConfig      ConditionMessage = "Spec updated; awaiting configuration application by the hardware manager"
 	ConfigSuccess    ConditionMessage = "Configuration has been applied successfully"
+	ConfigUpToDate   ConditionMessage = "Configuration is already up to date"
```
donpenney: nit: I would drop the "already", as it adds certain connotations to the message.

PR #4551, opened 2026-09-21
CNF-27072: Support MNO upgrades with multiple worker pools

Missxiaoguo:
## Summary

Adds controller-managed worker MachineConfigPool rollout for MNO ClusterVersion upgrades. Multiple worker pools can be upgraded serially or in parallel after the control plane, with optional non-EUS pools upgrading alongside the control plane.

**Jira**: [[CNF-27072](https://redhat.atlassian.net/browse/CNF-27072)]— Support MNO upgrades with multiple worker pools  

### Upgrade configuration

- Add `workerPoolUpgrade` with three strategies:
  - `OpenShiftDefault`: control plane and workers follow the default OpenShift rollout.
  - `Serial`: remaining worker pools upgrade individually in alphabetical order.
  - `Parallel`: remaining worker pools upgrade concurrently.
- Default to `OpenShiftDefault` for non-EUS upgrades and `Parallel` for EUS upgrades.
- Add `poolsWithControlPlane` for non-EUS Serial and Parallel upgrades.
- Validate strategy combinations and referenced MachineConfigPool names.

### Controller behavior

- Pause worker pools before triggering Serial and Parallel upgrades.
- Keep `poolsWithControlPlane` unpaused during non-EUS upgrades.
- Keep workers paused across both EUS upgrade hops.
- Start worker rollout only after the target ClusterVersion upgrade completes.
- Wait for each rollout wave to report current-generation `Updated=True`.
- Persist the resolved rollout configuration and pause-state ownership in ProvisioningRequest status.

### Robustness

- Require MachineConfigPools to be unpaused and updated before taking pause-state ownership.
- Persist pause-state ownership before modifying MachineConfigPools.
- Keep controller-managed pools paused while transient operations retry.
- Restore pools after terminal pre-start failures when the control plane has not advanced.
- Leave pools paused after EUS target-hop failures for administrator investigation.
- Check MCP `observedGeneration` before trusting `Updated=True`.
- Report all worker pools still updating in the status message.

### Tests and documentation

- Add unit coverage for validation, pool filtering, rollout strategies, recovery, timeout, and status cleanup.
- Update MNO E2E coverage to use multiple realistic worker pools.
- Add upgrade parameter schemas and regenerate CRD and bundle manifests.
- Document strategy selection, rollout behavior.

## Acceptance Criteria

- [x] Support clusters with multiple worker MachineConfigPools.
- [x] Upgrade worker pools serially after the control plane.
- [x] Upgrade worker pools in parallel after the control plane.
- [x] Allow selected non-EUS pools to upgrade with the control plane.
- [x] Support Serial and Parallel rollouts for EUS upgrades.
- [x] Preserve the existing OpenShift default behavior when appropriate.
- [x] Expose clear upgrade progress through ProvisioningRequest status.

### Automated testing

- Unit tests cover:
  - Worker pool strategy validation.
  - Serial and Parallel rollout combinations.
  - Rollouts with and without `poolsWithControlPlane`.
  - Alphabetical Serial ordering.
  - Stale MCP `observedGeneration`.
  - Pause-state recovery and terminal failures.
  - Upgrade timeout and status cleanup.
- MNO E2E tests cover multi-pool Parallel and Serial upgrades.
- `golangci-lint` passes.

### Live-cluster validation

Validated successfully on a cluster with three worker MachineConfigPools:

- [x] Y-stream upgrade using Serial.
- [x] Y-stream upgrade using Parallel.
- [x] EUS upgrade using Serial.
- [x] EUS upgrade using Parallel.


test/e2e/mno_cv_upgrade_test.go
```diff
@@ -436,14 +448,26 @@ var _ = Describe("MNO Standard ClusterVersion Upgrade", Ordered, Label("mno-cv-u
...
+				string(provisioningv1alpha1.CRconditionReasons.InProgress),
+				"Waiting for worker pools ["+workerR740+", "+workerXR8620t+"] to finish updating",
+				provisioningv1alpha1.StateProgressing,
+			)
+			assertSpokeMCPPaused(testCtx, spokeClient, workerR740, false)
+			assertSpokeMCPPaused(testCtx, spokeClient, workerXR8620t, false)
+
+			// Complete worker-dell-r740 and worker-dell-xr8620t pool updates.
+			ensureSpokeMCP(testCtx, spokeClient, workerR740, corev1.ConditionTrue)
+			ensureSpokeMCP(testCtx, spokeClient, workerXR8620t, corev1.ConditionTrue)
```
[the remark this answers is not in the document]
Missxiaoguo: Right, I orignally had them, but I deleted them because transitions to "Updated=False" do not affect assertion or reconcilation decision since the test (line476-469) will immediately replace them with Updated=True. I can add them back though

internal/controllers/provisioningrequest_upgrade.go
```diff
@@ -839,6 +866,95 @@ func (t *provisioningRequestReconcilerTask) handleCVUpgradeCompleted(
...
+// PoolsWithControlPlane form a common prerequisite and must be updated before
+// any later wave starts.
+func (t *provisioningRequestReconcilerTask) reconcileWorkerPoolRollout(
+	ctx context.Context, spokeClient client.Client,
+) (ctrl.Result, bool, error) {
+	upgradeStatus := t.object.Status.Extensions.ClusterDetails.ClusterUpgradeStatus
+	if upgradeStatus.WorkerPoolUpgrade == nil {
+		return ctrl.Result{}, false, fmt.Errorf(
+			"workerPoolUpgrade status is missing for completed ClusterVersion upgrade")
+	}
```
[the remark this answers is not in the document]
Missxiaoguo: Compatibility with upgrades started by an older controller version is out of scope because this feature is dev preview. For upgrades initiated by this version, `prepareCVSpec` persists `WorkerPoolUpgrade` before `TriggerCVUpgrade` is called. If persistence fails, the ClusterVersion upgrade is not triggered. So reaching PhaseInProgress or PhaseCompleted with a nil WorkerPoolUpgrade indicates invalid status, and returning an error is intentional rather than guessing the rollout behavior.

internal/controllers/provisioningrequest_upgrade.go
browsell, line 1021: [Comment authored by OpenAI Codex]

After a Serial/Parallel retry persists `PauseStateManaged=true` and pauses worker MCPs, `prepareCVSpec` runs again on each pre-start reconciliation. If updated input is invalid, this branch marks the upgrade failed and returns without calling the existing restoration logic. The terminal graph-precondition branches have the same gap.

For non-EUS and EUS intermediate upgrades, this can leave worker MCPs paused indefinitely after a terminal failure. Please route all terminal pre-start failures through the same restore path used by `handleCVUpgradeTriggerRejection`, while preserving paused MCPs for EUS target-hop failures. Please also add a regression test covering this case.

Missxiaoguo: In theory, it's unlikely since the window for the invalid update is very small, but genuinely reachable. I'll make the updates to make it more robust.

Missxiaoguo: Updated!

internal/controllers/provisioningrequest_upgrade.go
```diff
@@ -942,29 +1054,35 @@ func (t *provisioningRequestReconcilerTask) handleCVUpgradePreStart(
...
 	// verify target in availableUpdates.
 	if result, proceed, err := t.checkCVGraphPreconditions(
-		ctx, spokeClient, cv, cvSpec,
+		ctx, spokeClient, cv, cvSpec, action,
 	); result.RequeueAfter > 0 || !proceed || err != nil {
 		return result, err
 	}
 
+	// Prepare the worker MCPs right before triggering the upgrade.
+	err = t.prepareWorkerMCPsForUpgrade(ctx, spokeClient, action)
```
browsell, line 1075: [Comment authored by OpenAI Codex]

This path persists PauseStateManaged and pauses the worker MCPs before triggering the ClusterVersion upgrade. If a transient pre-start error continues until the timeout expires, isCVUpgradeTimedOut marks the upgrade TimedOut and cleans up spoke access, but does not call unpauseNonMasterMCPs.

For non-EUS and EUS-intermediate upgrades, the control plane has not advanced, so the controller can leave worker MCPs paused after the upgrade becomes terminal. Please apply the same restoration logic used for terminal pre-start failures to these timeout paths, while retaining the intentional paused state for EUS target-hop timeouts, and add a regression test.

Missxiaoguo: Thanks for raising this. The transisent-error timeout scenario described is different from a definitive pre-start rejection. The transisent error would happen during or after `prepareWorkerMCPsForUpgrade`. At that point, `desiredUpdate` may already have been accepted and could resume later after timeout, so automatically unpausing workers would therefore be unsafe.

Additionally, if transient API errors persist long enough to reach the timeout, the cluster likely has a broader connectivity or API-server issue. The same failure may prevent reliable MCP restoration. Administrator investigation is required to investigate the spoke cluster.

I would prefer to leave the spoke cluster in its current state for any timeout scenario and require the administrator to investigate and resolve the issue, as documented in our guidance.

browsell: [Comment authored by OpenAI Codex]

Thanks for the clarification. I agree that a timeout leaves the trigger state uncertain, so automatically unpausing MCPs could be unsafe if the control-plane upgrade was already accepted. I withdraw this finding. The administrator investigation path is appropriate.

